PR #72, opened 2019-05-01
Fix HTML spellings

qaisjp:


leaanthony:
Thanks! It appears I have real problems typing HMTL 😂

PR #74, opened 2019-05-02
add archlinux support

lanzafame:


leaanthony:
Thanks so much for this! Would it be possible to update the Linux section in the README with the commands for Arch users? This would be great.

leaanthony:
Merged with branch rel-0.12 and updated the README. Thank you for this submission!

PR #82, opened 2019-05-10
Fix windows paths

mattn:
filename contains slash separators.

### DIFF
--- a/cmd/templates.go
+++ b/cmd/templates.go
@@ -63,7 +63,7 @@ func NewTemplateList(filenames *mewnlib.FileGroup) *TemplateList {
 		if err != nil {
 			log.Fatalf("corrupt metadata for template: %s", filename)
 		}
-		path := strings.Split(filename, string(filepath.Separator))[0]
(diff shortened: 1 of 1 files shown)


mattn:
Just type `wails init`. `NewTemplateHelper` call `files := mewn.Group("./templates")`. So the files contains slash path separator.
But `NewTemplateList` use `os.PathSeparator` (i.e. backslash on Windows).

leaanthony:
Hi. Thanks for highlighting this. I needed to fix it a slightly different way and have pushed those changes to v0.13.0-pre. Once I release v0.13.0 today, you can pull in the latest fixes using `wails update`. Thanks again!

mattn:
Which commit fixes this issue? I could not find it.

mattn:
I guess that this issue is not fixed in latest.

mattn:
Ah, it still not in master branch. Sorry.

leaanthony:
Hi @mattn! The fix has been pushed to master now. Please update using: go get -u github.com/wailsapp/wails/cmd/wails
Would love your feedback. Thanks.

PR #108, opened 2019-05-26
CMD: LINUX: Arch Linux detection without lsb-release

ocelotsloth:
The existing distribution detection does not work on Arch Linux
without the `lsb-release` package installed. This patch adds
detection using `/etc/os-release` in the same way that f9a1881
uses for Fedora and CentOS detection.

I changed the if statement that Bryn Sinclair used to a case
statement to avoid extra if-else-if statements.

I also needed to add a trim statement to remove the `"` characters
that are present in Arch Linux's `/etc/os-release` file.


ocelotsloth:
Hey @leaanthony I realize now you have a develop branch. Let me know if you want this re-based onto it and I'll resubmit this PR.

ocelotsloth:
Closing. Will reopen with a clean PR into develop

PR #110, opened 2019-05-27
CMD: LINUX: Arch Linux detection without lsb-release

ocelotsloth:
The existing distribution detection does not work on Arch Linux
without the `lsb-release` package installed. This patch adds
detection using `/etc/os-release` in the same way that f9a1881
uses for Fedora and CentOS detection.

I changed the if statement that Bryn Sinclair used to a case
statement to avoid extra if-else-if statements.

I also needed to add a trim statement to remove the `"` characters
that are present in Arch Linux's `/etc/os-release` file.

### DIFF
--- a/cmd/linux.go
+++ b/cmd/linux.go
@@ -77,10 +77,16 @@ func GetLinuxDistroInfo() *DistroInfo {
 		re := regexp.MustCompile(`^NAME=(.*)\n`)
 		// extract the distro name
 		osName := string(re.FindSubmatch(osRelease)[1])
-		// Check distro name against list of RedHat distros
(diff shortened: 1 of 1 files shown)


leaanthony:
Hi there! Thanks for the PR. I'm wondering if this method can be used for Manjaro also?

ocelotsloth:
There's actually an `ID-LIKE` tag for arch-like systems as well as an `ID` tag for Arch itself that would be more appropriate since it would add support for all arch-like systems.

That should also work for Debian and RHEL-like systems as well.

leaanthony:
That sounds promising. Standardising would be great :-)

Issue #118, opened 2019-06-16
Angular support


bh90210:
is there any underway support for Angular?

If there is no one working on it, I can do it. I am still hot from React, I believe this implementation will be much faster.

leaanthony:
Nobody has said they are doing this and who are we to put out your fire 🔥so please feel free to do this! Awesome work!

PR #131, opened 2019-06-22
Angular 8 template

bh90210:
#118


bh90210:
a couple of things.

**a** I was getting a constant error 
```bash
ERROR in src/app/app.component.ts(21,12): error TS2339: Property 'backend' does not exist on type 'Window'.
```

which I ninja by adding this line `// @ts-ignore` abode the complaining line. Essentialy using TypeScript's error exception. Not sure if it is worth investigating since everything seems to work, but I think it is good to know.

**b** Angular uses it's own cli. The way I "fix" this is by downloading a copy and reaching to it (in `frontend/node_modules`) is by `npx`. Now `npx` is supposidly part of `npm` so there should not be something to add here if the point is Wails' dependency to only be `npm` BUT on a fresh Ubuntu 18.04 for example `npx` was missing and had to manually install it.

leaanthony:
Awesome work again. Comments:

A: Technically, you'd need to extend the Window definition to include backend but like you've seen, it doesn't really matter. I think the ts-ignore directive is fine for this scenario.

B: One solution to this is to add ng as a dependency and save it: `npm install ng --save`. Then in your script, reference it locally with `./node_modules/.bin/ng serve`. Try that :-)

bh90210:
https://github.com/npm/npm/releases/tag/v5.2.0

I think the preferred way is via `npx` mainly for consistency with documentation and proposed solutions on previous tickets on the subject.

What I suggest is include preemptively `npx` in the dependencies just to be sure, although probably there was something funny from my part. I mostly wrote it for future reference in case a similar tickets open to have a go-to solution.

But ultimately it is your call :)
what do you think?

leaanthony:
I realised we could do this without npx: just add a "[postinstall](https://docs.npmjs.com/misc/scripts)" hook in package.json that contains "npm install -g ng"

leaanthony:
If this works, I'll test on mac 👍

leaanthony:
Depends on #138

leaanthony:
I'm still experimenting with Codacy so ignore for now :-)

leaanthony:
The mod.go.template will now need to be updated to be like the others in the develop branch. This makes sure people get the correct runtimes at build time 👍

leaanthony:
#141 gives us what we need! We just need to add something like this to the template.json:
```
 "dependencies": [
   {
     "bin": "npx",
     "help": "This template requires 'npx'. Please install with 'npm install -g npx'"
   }
 ]
```

bh90210:
I am anyway closing this issue cause I accidentally removed my wails fork off github.
fixing,testing and making a new pull asap!

PR #165, opened 2019-07-22
Fix Masterminds/semver reference import path (case-sensitivity)

bt:
There's a double-reference in `go.mod` for `Masterminds/semver` which potentially causes the following error when compiling:

```
/go/pkg/mod/github.com/!masterminds/sprig@v2.16.0+incompatible/semver.go:4:2: case-insensitive import collision: "github.com/Masterminds/semver" and "github.com/
masterminds/semver"
```

Given that the package is (currently) stylised as `Masterminds/semver`, it's easiest to fix this by renaming our reference to the capitalised `Masterminds`.

### DIFF
--- a/cmd/semver.go
+++ b/cmd/semver.go
@@ -3,7 +3,7 @@ package cmd
 import (
 	"fmt"
 
-	"github.com/masterminds/semver"
+	"github.com/Masterminds/semver"
 )
 
 // SemanticVersion is a struct containing a semantic version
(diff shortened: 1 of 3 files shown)


leaanthony:
Awesome job! 👍

bt:
Thanks for merging!

Any idea when this will make it into master?

leaanthony:
Never picked up on this comment @bt. As you probably know by now, it's there :-)

Issue #169, opened 2019-07-24
Improve `wails issue`


leaanthony:
Also include:

- GCC version
- npm version
- node version

leaanthony:
TODO: Windows GCC version

pedzed:
What about Windows 10 version info?

![image](https://user-images.githubusercontent.com/8146593/62166894-33278a00-b322-11e9-9f92-e63a6af50706.png)

leaanthony:
Is there an easy way to get that from Go?

bh90210:
#178

leaanthony:
Closing this. Windows support already added.

PR #176, opened 2019-07-30
169 improve wails issue

bh90210:
#169


leaanthony:
Nice Job! 🥇 
Added some string trimming for gcc output (Mac had extra linefeed).

PR #182, opened 2019-08-07
Linux db

leaanthony:


leaanthony:
@bh90210 Let me know what you think of this :-)

bh90210:
checking it asap

bh90210:
please give me a sec though, cause I get `Error: Program 'npm' not found.` from debian and I am not sure if it is something wrong with my Dockerfile..

bh90210:
OK it turns out debian9's default node version is `nodejs is already the newest version (4.8.2~dfsg-1).`
so the proper way to install the latest is via 
`curl -sL https://deb.nodesource.com/setup_12.x | sudo bash - && sudo apt-get install -y nodejs`

maybe debian and ubuntu should be different, since many sub-distros go for one or the other.
in any way the above `linuxdb.yaml` will work for every supported distribution :+1:

leaanthony:
So currently Debian inherits from Ubuntu so already has curl -sL `https://deb.nodesource.com/setup_12.x | sudo bash - && sudo apt-get install -y nodejs`. Did I need to update something?

Issue #184, opened 2019-08-08
linux alpine support


bh90210:
currently there is none, I'm am working on it :+1:

PR #186, opened 2019-08-08
184 linux alpine support

bh90210:
#184


bh90210:
I'm closing this issue.
As I am working on the docker wails dev idea we should we I have something more mature reconsider how to approach it :+1:

PR #198, opened 2019-08-21
Bug fix yaml update

bh90210:
This PR addresses a bug in the `0_setup.go` that prevented proper prerequisites checking.
Additionally some modifications were made in `linuxdb.yaml` addressing a couple of bugs that again prevented prerequisites checking.

At the moment and going to v1.0.0 we will be supporting:
#### Debian 9, Ubuntu 18.04
_Zorin 15, Parrot 4.7, Linuxmint 19_
#### Gentoo

I have updated the README to reflect the final lists of distros/libraries.

What is worth checking is if with the same libraries we can support one LTS version back on major distros. (debian 8, ubuntu 16.08, centos 6, fedora 29) I will be doing checks _slowly_ the next days and will update the README accordingly. 
EDIT: I also see on the docs that we support debian 10 & ubuntu 19.04, will also check if this is correct :+1: 

*One bug remaining is Arch does not perform the prerequisites check still*
If wails dependencies are present `wails init` && `wails build` function as expected.

Thoughts?


cmd/wails/0_setup.go
```diff
@@ -91,19 +91,52 @@ func checkLibraries() (errors bool, err error) {
 		// Check library prerequisites
 		requiredLibraries, err := cmd.GetRequiredLibraries()
 		if err != nil {
-			return false, err
+			return true, err
```
bh90210: so after fixing this bug too, arch still doesn't run the prerequisites tests.

leaanthony: That's so weird. I should get time this weekend to debug this. Will have to debug in the container 📦

leaanthony:
I'm going to merge this as is. Let's treat debugging Arch as a seperate concern.

PR #203, opened 2019-08-23
chore: deduplicate library verificaation code

Vaelatern:


bh90210:
hey @Vaelatern, what exactly happened here?

Vaelatern:
Someone pointed out low hanging fruit for deduplication, so I grabbed it.

cmd/linux.go
```diff
@@ -166,9 +169,9 @@ func RpmInstalled(packageName string) (bool, error) {
 
 // RequestSupportForDistribution promts the user to submit a request to support their
 // currently unsupported distribution
-func RequestSupportForDistribution(distroInfo *DistroInfo, libraryName string) error {
+func RequestSupportForDistribution(distroInfo *DistroInfo) error {
 	var logger = NewLogger()
-	defaultError := fmt.Errorf("unable to check libraries on distribution '%s'. Please ensure that the '%s' equivalent is installed", distroInfo.Name, libraryName)
+	defaultError := fmt.Errorf("unable to check libraries on distribution '%s'.", distroInfo.Name)
```
bh90210: this change makes sense. Seems redundant here to give back the message "Please ensure that the '%s' equivalent is installed"

cmd/linux.go
```diff
@@ -120,6 +120,9 @@ func parseOsRelease(osRelease string) *DistroInfo {
 	return result
 }
 
+// GetPkgInstalled is all functions that use local programs to see if a package is installed
+type GetPkgInstalled func(string) (bool, error)
```
bh90210: may I suggest changing this to `CheckPkgInstalled`? It is a bit confusing as it seems to me that it'll get a pkg installed.

cmd/system.go
```diff
@@ -269,55 +269,33 @@ func CheckDependencies(logger *Logger) (bool, error) {
 		if err != nil {
 			return false, err
 		}
+
+		var libraryChecker GetPkgInstalled
```
bh90210: code became clearer indeed. :+1:

bh90210:
```bash
./0_setup.go:111:52: too many arguments in call to cmd.RequestSupportForDistribution
        have (*cmd.DistroInfo, string)
        want (*cmd.DistroInfo)
```
please go to [wails/cmd/wails/0_setup.go](https://github.com/Vaelatern/wails/blob/deduplicate-code/cmd/wails/0_setup.go) line 111 and replace the line with `return false, cmd.RequestSupportForDistribution(distroInfo)`

leaanthony:
This is fantastic @Vaelatern! Thank you. The Linux support has probably evolved quicker than any other part so it is ripe for tidying up. 👍

leaanthony:
Do you want to include Void into this PR?

cmd/linux.go
leaanthony: Any chance you could remove the full stop at the end of the message? Codacy is moaning :-)

Vaelatern: Done!

Vaelatern:
Fixed the error, @bh90210 

@leaanthony I'd rather rebase the Void support on top of this after it's in, if that's OK.

leaanthony:
@Vaelatern awesome first PR! 👍

PR #205, opened 2019-08-25
fix: undefined: webview during build

Vaelatern:
Fixes a build error for me


leaanthony:
That's totally on me. Cheers for the fix.

PR #228, opened 2019-09-09
feat: Vuetify2 support

MichaelHipp:
The 'vuetify2-support' branch adds a Wails template for support of Vue2 + Vuetify2 + Webpack. The existing "Vuetify Basic" template now has a description that is specific to Vuetify1.5; no other changes were made to the v1.5 template.

I took the liberty of removing dependence on babel-polyfill as its presence/absence seemed of no effect.

The template selection offered by `wails init` now reads as follows:
```
Please select a template:
  1: Angular - Angular 8 template (Requires node 10.8+)
  2: React JS - Create React App v3 template
  3: Vue2/Webpack Basic - A basic Vue2/WebPack4 template
  4: Vuetify1.5/Webpack Basic - A basic Vuetify1.5/Webpack4 template
  5: Vuetify2/Webpack Basic - A basic Vuetify2/Webpack4 template
Please choose an option [1]:```

If we would prefer to omit the Vuetify1.5 template entirely, I will be glad to remove it.

Any other changes/corrections welcome. Thanks for the opportunity to contribute to Wails.

### CODE REVIEW

**Comment by leaanthony** [comment]
Hi @MichaelHipp - thanks for this! The reason `babel-polyfill` was there was so that it worked correctly in IE9. However, according to [this](https://babeljs.io/docs/en/babel-polyfill) it seems to have been deprecated for using core-js instead. We use this in the other templates so if you could add those 2 dependencies as per the page, then that means it'll work on Windows.

I'm not so bothered about the linting issues as it's generated code.

Thanks.

**Comment by MichaelHipp** [comment]
@leaanthony Be glad to. Just to make sure I understand, we need these 2 dependencies:
```
import "core-js/stable";
import "regenerator-runtime/runtime";
```
Do I make a new PR afterwards?

leaanthony:
If the version of Babel is < 7.4.0, then simply reintroduce the polyfill line. If it is >= 7.4.0 then add these 2 dependencies and add the lines as you wrote above:

- [regenerator-runtime](https://github.com/facebook/regenerator/tree/master/packages/regenerator-runtime)
- [core-js](https://github.com/zloirock/core-js)

Just pushing to this branch is sufficient.

MichaelHipp:
@leaanthony  Apologies for going dark. I believe I have it finished. Just need to test a couple more times. Should have something for you this evening.

leaanthony:
No worries @MichaelHipp and awesome news! 👍

leaanthony:
All ok @MichaelHipp ?

PR #266, opened 2019-10-23
fix: detect encoding on windows

leaanthony:


go.mod
```diff
@@ -26,6 +27,9 @@ require (
 	golang.org/x/crypto v0.0.0-20190510104115-cbcb75029529 // indirect
 	golang.org/x/net v0.0.0-20190509222800-a4d6f7feada5 // indirect
 	golang.org/x/sys v0.0.0-20190509141414-a5b02f93d862
+	golang.org/x/text v0.3.0
 	gopkg.in/AlecAivazis/survey.v1 v1.8.4
 	gopkg.in/yaml.v3 v3.0.0-20190709130402-674ba3eaed22
 )
+
+go 1.13
```
bh90210: heh, I think this should leave? :-p

leaanthony: Yeah. It annoys me this is the default.

Issue #313, opened 2020-01-01
(bridge) implement  bridge/dev server remote connection support


tmclane:
Update the bridge.js and supporting Javascript development servers to allow for remote bridge connections.
This will allow for connections made to the Node development server for each of the front0end implementations to successfully be made from remote devices.

The files needing updated are:

* runtime/assets/bridge.js
* cmd/templates/angular-template/frontend/package.json.template
* cmd/templates/vuebasic/frontend/vue.config.js
* cmd/templates/vuetify-basic/frontend/vue.config.js

leaanthony:
Fixed by #314

PR #314, opened 2020-01-01
[313-remote-conn] allow remote connections to the websocket bridge

tmclane:
Update the bridge.js and supporting Javascript development servers to allow for remote bridge connections.
This will allow for connections made to the Node development server for each of the front0end implementations to successfully be made from remote devices.

The files needing updated are:

runtime/assets/bridge.js
cmd/templates/angular-template/frontend/package.json.template
cmd/templates/vuebasic/frontend/vue.config.js
cmd/templates/vuetify-basic/frontend/vue.config.js

*Note*: React was not forgotten it already listens on all interfaces by default.

Addresses issue #313


tmclane:
Codacy is complaining about spaces which should be 2 tab characters in the vue.config.js files.  Is this something you want or is Codacy using an invalid validation setting?

It appeared that the original files didn't include tabs in these files either.

tmclane:
I added a commit to add the `tabs` back in to the two files and that makes Codacy not show any issues at all anymore.  I am intending to squash that back in to the change that caused it to show up in the first place (a5fa9f2e197b2ff1f53191d2c6475ae08d61fcae) if that is desired. If you would rather it be 4 spaces (which is what was in the original file I can just drop that commit from the branch).

leaanthony:
Hey @tmclane! Does this work correctly? The reason I ask is that the Websocket server only maintains a [single connection](https://github.com/wailsapp/wails/blob/master/lib/renderer/bridge.go#L47) so I'm thinking that any messages sent by Go will only be sent to one of the frontends? 
I'm totally down with adding this PR though. Really cool stuff :+1:

tmclane:
> Hey @tmclane! Does this work correctly? The reason I ask is that the Websocket server only maintains a [single connection](https://github.com/wailsapp/wails/blob/master/lib/renderer/bridge.go#L47) so I'm thinking that any messages sent by Go will only be sent to one of the frontends?
> I'm totally down with adding this PR though. Really cool stuff

I hadn't noticed that 'feature' :).  It definitely works with a single connection which is what I was using for testing. I thought it helpful to be able to have at least a single remote connection since I was using my iPad for UI display and updating the `vue` files with Emacs.

FYI: I may attempt to tackle allowing more than one connection then in some future PR.

leaanthony:
Looks like I may have misunderstood your original requirement. This PR is just fine for remote access. Merging.

leaanthony:
I may have misunderstood your original requirement. I thought you needed multiple connections. This is good as is.

Issue #316, opened 2020-01-04
multiple bridge connection support


tmclane:
This issue is a proposal for multiple websocket management in the `renderer/bridge` code:

The proposal includes the following:
* Updating the `interfaces.Renderer` `Dispatch` function signature to accept a `CallbackFunc` argument to be used in place of the `Callback` function defined in the interface.
* updating the `lib.renderer.WebView` and `lib.renderer.Bridge` to use the updated `Dispatch` passing in the relevant private callback function such that the plumbing works properly
* deprecating and removing the `interfaces.Renderer.Callback` function requirement (replaced by previous bullet point)
* updating the `lib.renderer.Bridge` code to use a `map[string]*conn.WebSocket` instead to keep track of connected sessions for notification
* (optional) add a `NewBridge` function to replace the direct use of the `Bridge{}` struct initialization in app.go
* (optional) move the bridge code to a package of renderer -> `lib.renderer.bridge` and create a new `renderer` package level bridge.go containing the `NewBridge` function

The last two marked optional can be removed and the new websocket `session` code be placed inside the original `lib/renderer/bridge.go` file if so desired.  I thought that perhaps since the webview had a similar pattern that this might be a better layout.

Note: The code doing all of the above is found in [tmclane/316-multi-bridge-conn](https://github.com/tmclane/wails/tree/316-multi-bridge-conn)

leaanthony:
Ping me when you're back and we'll go over this.

PR #317, opened 2020-01-04
316-multi-bridge-conn

tmclane:
This PR implements the code backing Issue #316 

Copying the description from the issue:

* Updating the `interfaces.Renderer` `Dispatch` function signature to accept a `CallbackFunc` argument to be used in place of the `Callback` function defined in the interface.
* updating the `lib.renderer.WebView` and `lib.renderer.Bridge` to use the updated `Dispatch` passing in the relevant private callback function such that the plumbing works properly
* deprecating and removing the `interfaces.Renderer.Callback` function requirement (replaced by previous bullet point)
* updating the `lib.renderer.Bridge` code to use a `map[string]*conn.WebSocket` instead to keep track of connected sessions for notification
* removes the `lib.renderer.Bridge.injectCSS` function since it is unused and will not function properly with the session split off from the Bridge itself. If needed this can be reimplemented in such a way that all CSS provided to it is cached in the Bridge for injection into each session as it starts up.
* (optional) add a `NewBridge` function to replace the direct use of the `Bridge{}` struct initialization in app.go
* (optional) move the bridge code to a package of renderer -> `lib.renderer.bridge` and create a new `renderer` package level bridge.go containing the `NewBridge` function

The last two marked optional can be removed and the new websocket `session` code be placed inside the original `lib/renderer/bridge.go` file if so desired.  I thought that perhaps since the webview had a similar pattern that this might be a better layout.


lib/renderer/session.go
```diff
@@ -0,0 +1,85 @@
...
+
+	// Inject bindings
+	for _, binding := range s.bindingCache {
+		s.evalJS(binding, bindingMessage)
+	}
+
+	// FIXME: What should happen here? With multiple sessions occurring this event
+	//        seems less useful.
+	// Emit that everything is loaded and ready
+	//s.eventManager.Emit("wails:ready")
```
leaanthony: People rely on this event so we still need to keep it. It's really just an indicator that the frontend is loaded so the backend can use the frontend runtime. Suggest we emit this on first load. I guess it gets tricky as there's no concept of multiple sessions in the backend.

tmclane: No problem. I think I know where to put it instead.  Doesn't make sense to me to emit this everytime a frontend connection is made? Does it need to happen when a connection is made? Or can it only happen once when the Bridge is initialized and ready?

tmclane: Perhaps a new event like `wails:bridge-session` for everytime a session is initiated / rejoined?

tmclane: I can easily output it only after the first session connects.  Would that be acceptable?

tmclane: Ok this has been corrected. It will now be emitted but only by the first session spawned by the bridge.

leaanthony: > Perhaps a new event like `wails:bridge-session` for everytime a session is initiated / rejoined?

This is a great idea, passing the identifier with it.

tmclane:
Looks like I still have some work to do:

   ```

DEBU[0000] [Events] Got Event                            data="[127.0.0.1:36378]" name="wails:bridge:session:started"
panic: concurrent write to websocket connection

goroutine 10 [running]:
github.com/gorilla/websocket.(*messageWriter).flushFrame(0xc0001d7ac0, 0xc000222001, 0xc00022213c, 0x0, 0x4, 0x50, 0xc0002220f0)

```

leaanthony:
Is this good to.go?

tmclane:
> Is this good to.go?

Let me retest it again and make sure there aren't any other edge cases that popup.  I'll let you know either way.

tmclane:
@leaanthony Just tested it again after rebasing.  
Everything looks good to me and works from multiple machines at the same time.
You are going to squash the commits in correct?  If so I won't bother splitting the last commit up and reallocating it to the previous commits with the fixup rebase command.

Issue #326, opened 2020-01-09
Support Distribution 'Raspbian GNU/Linux'


tmclane:
**Description**
Distribution 'raspbian' is currently unsupported.

**Further Information**
Opening this to submit a PR with support added.

| Name   | Value |
| ----- | ----- |
| Wails Version | v1.0.2-pre3 |
| Go Version    | go1.13.5 |
| Platform      | linux |
| Arch          | arm |
| GO111MODULE   | (Not Set) |
| Distribution ID   | raspbian |
| Distribution Name   | Raspbian GNU/Linux |
| Distribution Version   | 10 |

*Please add any extra information here, EG: libraries that are needed to make the distribution work, or commands to install them*

The basic Debian packages are sufficient.
Had to ensure `npm` and `mewn` were installed along with adding various directories to the `PATH` but that is environmental setup.

bh90210:
@tmclane are u on it or should I do it?

tmclane:
I opened this issue since I have already got it running!  PR shortly

leaanthony:
Ready for testing :-)

leaanthony:
I'm going to assume this is working because @tmclane said his PR worked for him locally.

PR #327, opened 2020-01-09
[326-platform-raspbian] feat: implement raspbian support

tmclane:
Debian dependencies are sufficient to allow this to work.

Tested:
* Angular
* React
* Vuetify2
* Vuetify1.5
* Vue2

### DIFF
--- a/cmd/linux.go
+++ b/cmd/linux.go
@@ -51,6 +51,8 @@ const (
 	ManjaroARM
 	// Deepin distribution
 	Deepin
+	// Raspbian distribution
+	Raspbian
 )
 
 // DistroInfo contains all the information relating to a linux distribution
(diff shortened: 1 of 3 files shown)


tmclane:
#326

cmd/linux.go
```diff
@@ -51,6 +51,8 @@ const (
 	ManjaroARM
 	// Deepin distribution
 	Deepin
+	// Raspbian
```
bh90210: the pattern here is `// Raspbian distribution` otherwise codacy complains..

tmclane: just fixed it! and Codacy did complain!

tmclane: Is there a way to run Codacy checks locally?  I know it does more than just what `golint ./...` does although in this case `golint` would have complained had I run it.

PR #361, opened 2020-03-17
Macos notarize fix

konez2k:
This patch fixes a problem with macos signing / notarize, which prevents the built app running after signing, returning the following error:
`the code is valid but does not seem to be an app`


leaanthony:
This is awesome! Thank you. Could you please do a PR against develop?

PR #370, opened 2020-03-24
fix RunCommandArray bug 

Rested:
where if `exec.LookPath` returns an error program is an empty string in the log.

before if e.g. mewn can't be found in path, log will say 
```
ERROR: Looks like '' isn't installed. Please install and try again.
```
after:
```
ERROR: Looks like 'mewn' isn't installed. Please install and try again.
```


leaanthony:
Nice catch. Could you please make the PR against the develop branch. Thank you!

Rested:
closed in favour of https://github.com/wailsapp/wails/pull/371 :)

PR #371, opened 2020-03-24
fix RunCommandArray bug where if exec.LookPath returns an error progr…

Rested:
…am is an empty string in the log.


leaanthony:
Awesome! Thanks for this!

PR #373, opened 2020-03-25
Bugfix/372 bridge chan

tmclane:
Updates the code to use a `chan string` to guard the websocket from multiple goroutine writes.

### DIFF
--- a/lib/renderer/bridge/session.go
+++ b/lib/renderer/bridge/session.go
@@ -1,7 +1,8 @@
 package renderer
 
 import (
-	"sync"
+	"time"
+	"unsafe"
 
 	"github.com/gorilla/websocket"
 	"github.com/leaanthony/mewn"
@@ -20,7 +21,22 @@ type session struct {
 	ipc          interfaces.IPCManager
 
 	// Mutex for writing to the socket
-	lock sync.Mutex
+	shutdown  chan bool
+	writeChan chan []byte
+
+	done bool
+}
+
+func newSession(conn *websocket.Conn, bindingCache []string, ipc interfaces.IPCManager, logger *logger.CustomLogger, eventMgr interfaces.EventManager) *session {
+	return &session{
+		conn:         conn,
+		bindingCache: bindingCache,
+		ipc:          ipc,
+		log:          logger,
+		eventManager: eventMgr,
+		shutdown:     make(chan bool),
+		writeChan:    make(chan []byte),
+	}
 }
 
 // Identifier returns a string identifier for the remote connection.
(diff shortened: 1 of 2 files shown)


tmclane:
@leaanthony I can't see how to add you as a reviewer.

lib/renderer/bridge/session.go
```diff
@@ -83,8 +98,38 @@ func (s *session) Callback(data string) error {
...
+// Shutdown
+func (s *session) Shutdown() {
+	s.done = true
+	s.shutdown <- true
+	s.log.Debugf("session %v exit", s.Identifier())
+}
 
+// writePump pulls messages from the writeChan and sends them to the client
+// since it uses a channel to read the messages the socket is protected without locks
+func (s *session) writePump() {
```
leaanthony: Great solution!

lib/renderer/bridge/session.go
```diff
@@ -20,7 +20,22 @@ type session struct {
 	ipc          interfaces.IPCManager
 
 	// Mutex for writing to the socket
-	lock sync.Mutex
+	shutdown  chan bool
+	writeChan chan string
```
leaanthony: Suggestion, let's use []byte. More below.

lib/renderer/bridge/session.go
```diff
@@ -20,7 +20,22 @@ type session struct {
...
+
+func newSession(conn *websocket.Conn, bindingCache []string, ipc interfaces.IPCManager, logger *logger.CustomLogger, eventMgr interfaces.EventManager) *session {
+	return &session{
+		conn:         conn,
+		bindingCache: bindingCache,
+		ipc:          ipc,
+		log:          logger,
+		eventManager: eventMgr,
+		shutdown:     make(chan bool),
+		writeChan:    make(chan string),
```
leaanthony: []byte

lib/renderer/bridge/session.go
```diff
@@ -33,19 +48,15 @@ func (s *session) Identifier() string {
...
 
 func (s *session) sendMessage(msg string) error {
-	s.lock.Lock()
-	defer s.lock.Unlock()
-
-	if err := s.conn.WriteMessage(websocket.TextMessage, []byte(msg)); err != nil {
-		s.log.Debug(err.Error())
-		return err
+	if !s.done {
+		s.writeChan <- msg
```
leaanthony: Let's convert it here as this only happens once. A zero copy way to do this (I think) is to do:
`*(*[]byte)(unsafe.Pointer(&msg))`. We can only do this because this string is read only.

lib/renderer/bridge/session.go
```diff
@@ -83,8 +98,38 @@ func (s *session) Callback(data string) error {
...
+		select {
+		case msg, ok := <-s.writeChan:
+			s.conn.SetWriteDeadline(time.Now().Add(1 * time.Second))
+			if !ok {
+				s.log.Debug("writeChan was closed!")
+				s.conn.WriteMessage(websocket.CloseMessage, []byte{})
+				return
+			}
+
+			if err := s.conn.WriteMessage(websocket.TextMessage, []byte(msg)); err != nil {
```
leaanthony: we can now use `msg` directly, avoiding memory allocations on each connection.

leaanthony: Or alternatively, just use `*(*[]byte)(unsafe.Pointer(&msg))` in the `WriteMessage` call :-)

lib/renderer/bridge/session.go
tmclane: Will do.

tmclane:
@leaanthony Made those adjustments!

PR #347, opened 2020-02-19
Add cross-compile support via xgo

konez2k:
Added `-x <platform>` flag to build command, supported platforms are the same as xgo.
You need to install `techknowlogick/xgo` via `go get src.techknowlogick.com/xgo` (this repo supports go modules).
The platform packaging is also supported using the same xgo docker image, for windows windres tool.

By default the executables are stored inside the `build` folder, if you dont specify the architecture, the arch is automatically set to amd64.

Cross-compilation has been tested on Linux.
Packed builds tested on MacOS 10.14 and Windows 10 via:

`wails build -p -x windows/386`.
`wails build -p -x windows`.
`wails build -p -x darwin`.

PR #377, opened 2020-03-28
Feature/cross compilation

tmclane:
This PR supersedes and obsoletes the original PR https://github.com/wailsapp/wails/pull/347.
Relevant code from that PR is included in this branch and was built on top of and extended to complete it.

This branch does the following:

- removes the usage of the local `xgo` binary in favor of calling `docker run` directly
- stops `scripts/build.sh` on error
- updated help command on `-x` to output supported targets
- uses `wailsapp/xgo:latest` as the build image
- runs build process inside the docker image with the same UID as your local user 
  this enables the output files to be owned by your local user instead of root

        | |     / /___ _(_) /____
        | | /| / / __ `/ / / ___/
        | |/ |/ / /_/ / / (__  )  v1.0.3-pre2
        |__/|__/\__,_/_/_/____/   https://wails.app
        The lightweight framework for web-like apps
        
        wails build - Builds your Wails project
        
        This command will check to ensure all pre-requistes are installed prior to building. If not, it will attempt to install them. Building comprises of a number of steps: install frontend dependencies, build frontend, pack frontend, compile main application.
        
        Flags:
        
          -d	Build in Debug mode
          -f	Force rebuild of application components
          -help
            	Get help on the 'wails build' command.
          -p	Package application on successful build
          -t string
            	Generate Typescript definitions to given file (at runtime)
          -verbose
            	Verbose output
          -x string
            	Cross-compile application to specified platform via xgo
            	 - darwin/amd64
            	 - linux/i386
            	 - linux/amd64
            	 - linux/arm-7
            	 - windows/amd64
            	
*Note:* This has only been tested on a Linux host at this time.  Compilation of all supported platforms was successful and they executed correctly on the target platform.


tmclane:
Currently testing a compile on my Mac.

tmclane:
Note I removed `linux/i386` because it seem there is an issue I need to investigate.

leaanthony:
I think 32bit linux is as good as obsolete now. You're probably missing some 32bit compat libs.

cmd/helpers.go
```diff
@@ -89,17 +132,86 @@ func BuildApplication(binaryName string, forceRebuild bool, buildMode string, pa
...
+			"-e", fmt.Sprintf("LOCAL_USER_ID=%v", user.Uid),
+			"-e", fmt.Sprintf("FLAG_LDFLAGS=%s", ldflags),
+			"-e", "FLAG_V=false",
+			"-e", "FLAG_X=false",
+			"-e", "FLAG_RACE=false",
+			"-e", "FLAG_BUILDMODE=default",
+			"-e", "FLAG_TRIMPATH=false",
+			"-e", fmt.Sprintf("TARGETS=%s", projectOptions.Platform+"/"+projectOptions.Architecture),
+			"-e", "GOPROXY=",
+			"-e", "GO111MODULE=on",
```
leaanthony: Am wondering if we need GOPRIVATE... Haven't really thought about it before.

cmd/package.go
```diff
@@ -192,16 +221,35 @@ func (b *PackageHelper) PackageWindows(po *ProjectOptions, cleanUp bool) error {
...
-	windresCommand := []string{windresBatFile, sysofile, tgtRCFile}
-	err = NewProgramHelper().RunCommandArray(windresCommand)
-	if err != nil {
-		return err
+		args := []string{
+			"docker", "run", "--rm",
+			"-v", folder + ":/build",
+			"--entrypoint", "/bin/sh",
+			"wailsapp/xgo:latest",
+			"-c", "/usr/bin/x86_64-w64-mingw32-windres -o /build/" + basename + "-res.syso /build/" + basename + ".rc",
```
leaanthony: I'll need to check this works. The reason `windres.bat` exists is because there was a bug around shelling in Windows/Go.

cmd/wails/4_build.go
```diff
@@ -52,6 +74,25 @@ func init() {
 			return fmt.Errorf("Unable to find 'project.json'. Please check you are in a Wails project directory")
 		}
 
+		// Set cross-compile
+		projectOptions.Platform = runtime.GOOS
+		if len(platform) > 0 {
```
leaanthony: I'm going to pimp `https://github.com/leaanthony/slicer` here as you can simply do `if !supportedPlatforms.Contains(plat)` :-)

lib/renderer/renderer-mewn.go
```diff
@@ -5,7 +5,7 @@ package renderer
...
...5fd7f91a415fcb8c93758471579d712b8b966e6ded97a40d4af44d77aae3e49b0a73076b5d7ac846a311694f3a2fa67a782ad6dc1c946b9b042bceaa3be319e4e021422f78b3c37b2bea924f9b0116e34855fb2908945916f8adf57a75889f952294ec5ed17c25ee8dd8325e275e35d468db30f3701cfda8da5dcc5c86d0e9217430802ea3e7b0a1635109c62372a6724f44d7582322ac1091ac343bd8ae9851cdc663dfea87e35316aa7ae443e01fb2ecbe4aa92737ac2f95d9b44a8fe669799d03fea2a4ce42044678f8c34bc33bcba5cd0e28b060b2ab8dce3364078cc7d7556b343318efabfbfb1754efd5b3fc2107c212c044fce0bdd3cb9d074a50fa092778feb7695912c64154b6f71b00ff7ef843a465f9b9f2338633df9e9c8936b73683ea71de4f6e4065a753f7730dbe71aefce9f05955ed20256844b0ff616727d4b7d63e9e61a867bf5ddd20ad0db8b308e8baf8637fc86fe5f1c6ae63c79bde2db6bc02b78c8f77e6094a77cda3f4ce7cdce55e1739c477fc7aab7d7ccd7f05e5fc12d439b2fd7b0adb98e41e7930d0ad2a8966c6837b5479f55045f572d87f77a30373ec5042e5c187f4c1d87d0ed91ae2270fb1111e3268d4b2ae4153fda26867f0db449bd6ed812a184ebe4185c38d28c6e8f7f8eba08999e14186b27075f6e8e3b75d79b6a64af0f3ff8ef2bc7a880c97881bbb164d257af3e6cd9be84aedd61b1f7d319b4a47b25a6e229c23e7c57fb8b4974b6c9e5eb3faf2e9062de0758b1b6d20bab568e55fb7bab2f96e6b1a3bf7ba5d1502b4b12cf06b7c19eabc8cf1f90c196161c192ff040000ffff8bf286145a140000")
```
leaanthony: Is there a reason why the wails.js hash data has changed?

scripts/build.sh
```diff
@@ -15,9 +15,30 @@ cd lib/renderer
 mewn
 cd ../..
 
-echo "**** Installing Wails locally ****"
+echo "**** Checking if Wails passes unit tests ****"
```
leaanthony: Nice

cmd/helpers.go
tmclane: Probably do.  I have a good way of dealing with that.  Probably should make an `issue` to cover it.

cmd/package.go
tmclane: bear in mind this is executing inside the docker container so that shouldn't matter in that case.

cmd/wails/4_build.go
tmclane: nice!  I was just keeping what was in the original branch.

lib/renderer/renderer-mewn.go
tmclane: Must be from commits before my changes in this branch since I used the original branch as a base.

scripts/build.sh
tmclane: yeah - if only there were more tests :)

tmclane:
Mac OS X Update

No issues compiling `linux/amd64`, `windows/amd64`, and `linux/arm-7` on a OS X host.

One oddity I noticed is that the binaries are acting like the `-w -s` arguments are not being passed in. 
Something to investigate further but I don't think is a blocker for merging this branch.

lib/renderer/renderer-mewn.go
tmclane: I'll drop it out of this branch.

tmclane:
> I think 32bit linux is as good as obsolete now. You're probably missing some 32bit compat libs.

yes I think I missed some :)

cmd/wails/4_build.go
leaanthony: Yeah, I should have revisited the whole project for these scenarios and tidied it up a bit

cmd/package.go
leaanthony: yes, of course. The issue was definitely based on running Go on Windows

lib/renderer/renderer-mewn.go
leaanthony: No need. Looks like there wee dependency updates in npm.

scripts/build.sh
leaanthony: It's a pretty hard setup to test to be fair.

leaanthony:
Have some time to test this today. One thing, could we throw up a message before issuing the docker command so that the user knows what's happening? On my network, it took a while to download xgo and so all i saw was something that looked like a hang:
image

leaanthony:
Windows app compiled on Mac:

- When running the app, you get this message about signing. I'm wondering if there's a way to register with Microsoft so we can sign apps....
image
- The app does run, however it looks like it's been compiled without the `-H` flag, so a terminal appears:
image

However, the app does work!

tmclane:
Yes I didn’t like how long it took to download the docker image. Definitely need to improve that if possible.  I don’t believe the ldflags are making it at least when compiling on my Mac.

leaanthony:
Tried mac->linux/amd64: worked perfectly 👍

leaanthony:
Regarding the windows signing, I guess we can leave that to the user to sign, unless we can pass in their certificate to docker and sign it for them. But the onus has to be them supplying that cert.

tmclane:
Yes we can provide a way to make it happen.

leaanthony:
Linux->Mac build: 

- Looks like the bundled app is being written to the project root. Are we expecting it to be in the build directory?
- Work fine though!

Linux->Windows build:

- Works fine. Just the `-H flag` issue

leaanthony:
Windows: It doesn't work. Looks like the fix for the root user doesn't translate to windows:
image

tmclane:
> Windows: It doesn't work. Looks like the fix for the root user doesn't translate to windows:
> image

Well that is a bummer.  I wouldn't have thought the user mapping would matter on Windows at all!   I'll do more investigation.

cmd/helpers.go
tmclane: > Linux->Mac build:
> 
>     * Looks like the bundled app is being written to the project root. Are we expecting it to be in the build directory?
> 

I meant to change where that was being created and update the path to move it in to the `build` directory.

tmclane:
@leaanthony I believe Windows should now build correctly since I am using userid=1000 if the value from `user.Current()` is not an integer. 

Also I believe the flags should now be processed properly inside the image.  This should remove the console window and get Go to slim the executable down again.

tmclane:
I'm working on adding the following:
* messaging related to downloading / or building the docker image
* option to build the docker image locally
* general cleanup of the build functions (splitting Native and Docker up)

tmclane:
Example Windows image for signing purposes: https://github.com/StefanScherer/dockerfiles-windows/tree/master/signtool

tmclane:
> I'm working on adding the following:
> 
>     * messaging related to downloading / or building the docker image
> 
>     * option to build the docker image locally
> 
>     * general cleanup of the build functions (splitting Native and Docker up)

- Messaging related to the image download + target platform being built
- cleanup of build functions

1. OS X `.app` is now built inside the `build` directory
2. OS X Packager no longer renames the cross-compiled binary
3. Windows resources are created, utilized, and remove in the `build` directory

leaanthony:
Awesome work!

I can confirm the user issue has gone, as the -H flag issue. 

It looks like the icon isn't getting picked up for windows when using `-p`. At least, the icon isn't showing and the build sizes for the binaries are identical regardless of `-p`. Can confirm the manifest, etc are getting built correctly and the `.ico` file does seem to have the icon image. 

image

Also, I noticed the linux/mac binaries are picking up the `.exe` suffix:
image
It would be great if we only did that for the windows builds.

tmclane:
@leaanthony Believe the icon is in there now.  Please try again!
It should now cleanup all those windows files when the compilation succeeds.

cmd/helpers.go
```diff
@@ -90,7 +215,7 @@ func BuildApplication(binaryName string, forceRebuild bool, buildMode string, pa
 	}
 
 	buildCommand := slicer.String()
-	buildCommand.Add("mewn")
+	buildCommand.Add("go")
```
oliverpool, line 218: The `wails serve` command now fails on my machine.

To make it work, I had to move the lines 220-223 (add the `-i` flag) after the line 225 `buildCommand.Add("build")`

cmd/helpers.go
```diff
@@ -111,39 +236,21 @@ func BuildApplication(binaryName string, forceRebuild bool, buildMode string, pa
 				binaryName = strings.TrimSuffix(binaryName, ".exe")
 			}
 		}
-		buildCommand.Add("-o", binaryName)
+		buildCommand.Add("-o", filepath.Join("build", binaryName))
```
oliverpool, line 239: This breaks my `wails serve` command as well.

To fix it, please change line 536 of this file with:

```go
location, err := filepath.Abs(filepath.Join("build", projectOptions.BinaryName))
```

cmd/helpers.go
tmclane: Thanks for the pointers!  I hadn't validated that `wails serve` was working yet.

tmclane:
> Awesome work!
> 
> I can confirm the user issue has gone, as the -H flag issue.
> 
> It looks like the icon isn't getting picked up for windows when using `-p`. At least, the icon isn't showing and the build sizes for the binaries are identical regardless of `-p`. Can confirm the manifest, etc are getting built correctly and the `.ico` file does seem to have the icon image.
> 
> image
> 
> Also, I noticed the linux/mac binaries are picking up the `.exe` suffix:
> image
> It would be great if we only did that for the windows builds.

Will be done.

cmd/helpers.go
```diff
@@ -223,7 +223,7 @@ func BuildNative(binaryName string, forceRebuild bool, buildMode string, project
 
 	if binaryName != "" {
 		// Alter binary name based on OS
-		switch runtime.GOOS {
+		switch projectOptions.Platform {
```
leaanthony, line 226: This doesn't fully solve the issue. When creating a project on Windows, the `binaryname` option contains the `.exe`. We normally strip this out from non-windows platforms, as per the lines 231-235 below. As the docker binaries appear to have `-os-platform` as their suffix, I'm wondering if this check is now failing? Tried to print out the values but get nothing when the code path goes through `BuildDocker`

tmclane: I actually removed the word `.exe` from the code entirely and it STILL appears!  
I think the docker container might somehow be injecting it.   Unfortunately Windows is SO slow for me I might have to reboot into Windows itself on my laptop and see if I can get to the bottom of it.

tmclane: I was looking at the `build.sh` script and it only adds `extensions` for each platform (and `.exe` is definitely an extension).  

It's possible I wasn't running with the code I updated ?  I ran the `scripts/build.sh` to make sure it was updated but I don't believe it was working.    

When cross-compiling with Docker it should already be adding the appropriate extension so we shouldn't have to worry about it.

tmclane: Yes I know it does it for all platforms when executed on Windows!  I duplicated the problem on my machine (very slow machine).

tmclane: Yes I noticed it won't print the Docker commandline in verbose mode on Windows as well.  I added an additional `fmt.Printf` in front and that shows up but I didn't see any reason for the funny name since it wasn't passing that in to the docker run command at all.

I pushed commits so I could fetch them on my windows machine.  I'll let you know when i think its fixed.

leaanthony: Is there anything I can do to help with this? 😃

tmclane: Dang it was a nightmare tracking down where the `xcc.exe` prefix was coming from!  It turns out projects created on Windows have that as their module name inside the `go.mod` file!

leaanthony: Nice catch! Will review this today

leaanthony:
So, i think there may be some crosswires here. Reverting those last 2 commits made everything work. So why did it fail before? I too had a .exe in my module name. So how did it get there? The templates use the binary name for the module name and it looks like, for my side, i named the binary explicitly to `xcc.exe` in the init steps. You could argue it should be stripped, but should it? It *is* the name you are giving it. Do I *think* this is all working. Will test more later.

tmclane:
The ‘exe’ in the module name is fine for non-crosscompile but I believe for crosscompile it is the source of that funny name executable. Surely go adds an exe anyhow when not cross compiling?

leaanthony:
We use -o to explicitly set the output name so we must add it. I'll test more when I can but I'm sure that there are no ".exe" strings added  other than for the -o option

tmclane:
> We use -o to explicitly set the output name so we must add it. I'll test more when I can but I'm sure that there are no ".exe" strings added other than for the -o option

I removed every instance of `exe` in the codebase and it still appears as a prefix to the `xgo` platform-architecture suffix if it is in the `go.mod` file as the module name!
That's why I don't believe the `-o` option matters for the cross compile version.  It should still be there for the `native` case unless I messed something up.

leaanthony:
The 2 commits above should fix the outstanding issues for this PR. Tests run:

# Windows

  - Generate new project: xcc (defaults for all options)
  - go.mod contains expected module name: module xcc
  - Native build produces `build/xcc.exe`
  - `wails build -x windows/amd64` produces `build/xcc-windows-4.0-amd64.exe`
  - `wails build -x darwin/amd64` produces `build/xcc-darwin-10.6-amd64`
  - `wails build -x linux/amd64` produces `build/xcc-linux-amd64`
  - `wails build -x windows/amd64 -p` produces `build/xcc-windows-4.0-amd64.exe`
  - `wails build -x darwin/amd64 -p` produces `build/xcc.app/*`
  - `wails build -x linux/amd64 -p` produces `build/xcc-linux-amd64`

# Linux

  - Generate new project: xcc (defaults for all options)
  - go.mod contains expected module name: module xcc
  - `wails build` produces `build/xcc`
  - `wails build -x windows/amd64` produces `build/xcc-windows-4.0-amd64.exe`
  - `wails build -x darwin/amd64` produces `build/xcc-darwin-10.6-amd64`
  - `wails build -x linux/amd64` produces `build/xcc-linux-amd64`
  - `wails build -p` produces `build/xcc`
  - `wails build -x windows/amd64 -p` produces `build/xcc-windows-4.0-amd64.exe`
  - `wails build -x darwin/amd64 -p` produces `build/xcc.app/*`
  - `wails build -x linux/amd64 -p` produces `build/xcc-linux-amd64`

# Mac

  - Generate new project: xcc (defaults for all options)
  - go.mod contains expected module name: module xcc
  - `wails build` produces `build/xcc`
  - `wails build -x windows/amd64` produces `build/xcc-windows-4.0-amd64.exe`
  - `wails build -x darwin/amd64` produces `build/xcc-darwin-10.6-amd64`
  - `wails build -x linux/amd64` produces `build/xcc-linux-amd64`
  - `wails build -p` produces `build/xcc`
  - `wails build -x windows/amd64 -p` produces `build/xcc-windows-4.0-amd64.exe`
  - `wails build -x darwin/amd64 -p` produces `build/xcc.app/*`
  - `wails build -x linux/amd64 -p` produces `build/xcc-linux-amd64`

tmclane:
Thank you for helping get this across the finish line!

PR #407, opened 2020-05-08
Disable F5 key on windows

leaanthony:


leaanthony:
Looks like Codacy isn't picking up the CWE-126 fix.

Issue #397, opened 2020-04-25
Generate .ico files for Windows builds


dedo1911:
**Description**

After running `wails build -p` it successfully generates icon resources to edit (appname.ico) but this file gets deleted running `wails build` (without -p) or `wails serve`.

This appears to be happening for Windows resources only. macOS appicon.png is not being deleted.

Involved resources:
`appname.rc`
`appname.ico`
`appname.exe.manifest`
`appname-res.syso`

**To Reproduce**
Steps to reproduce the behaviour:
1. Run `wails build -p`
2. Check that resource files have been generated
3. Run `wails build` or `wails serve`
4. Check that resource files have been deleted

**Expected behaviour**
Resource files shouldn't be deleted (at least .ico) upon non packaging builds

**System Details**

| Name   | Value |
| ----- | ----- |
| Wails Version | v1.0.2 |
| Go Version    | go1.13.5 |
| Platform      | windows |
| Arch          | amd64 |
| GO111MODULE   | (Not Set) |
| GCC           | 8.1.0 |
| Npm           | 6.13.7 |
| Node          | v13.8.0 |

leaanthony:
This is working as intended. Originally the thought was just to keep the project clean. What's the scenario where this is causing a problem for you?

dedo1911:
The issue is that .ico file is always the stock wails icon and is not being generated from appicon.png

This means that every time i need to manually add the .ico resource to the root of the project as it’s not being kept or correctly generated.

leaanthony:
The ICO file absolutely should ve generated from the PNG. If that isn't the case, it's a bug.

dedo1911:
Ok I've tried once more and I've modified **only** `appicon.png`
The result is the same:
macOS → builds are working correctly. appicon.png is being used correctly and .app folders display the correct icon.

Windows → builds generate the ico file from scratch using the default one if none is present, instead of generating it from appicon.png.

To get a correct Windows package I would need to manually convert png to ico then run a build again. This is a bug.

leaanthony:
nice catch. I really thought we were generating the ICO file. This absolutely should be happening. Thanks for raising it.

dedo1911:
1.6.0-pre2: icons are being used correctly from "appname.ico" and after build the file is being deleted.
I think only missing is ico generation then.

leaanthony:
[...]

PR #420, opened 2020-05-21
Improved Windows .ico generation

dedo1911:
ico generate multiple sizes (256, 128, 64, 48, 32, 16) as described in https://github.com/wailsapp/wails/issues/397#issuecomment-631412932

### DIFF
--- a/cmd/package.go
+++ b/cmd/package.go
@@ -95,21 +95,6 @@ func generateWindowsIcon(pngFilename string, iconfile string) error {
 		return err
 	}
 
-	outfile, err := os.Create(iconfile)
-	if err != nil {
-		return err
-	}
-	defer outfile.Close()
-
-	ico := windowsIcoHeader{
-		imageType:  1,
-		imageCount: uint16(len(sizes)),
-	}
-	err = binary.Write(outfile, binary.LittleEndian, ico)
-	if err != nil {
-		return err
-	}
-
 	icons := []windowsIcoContainer{}
 
 	for _, size := range sizes {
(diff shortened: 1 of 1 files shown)


cmd/package.go
```diff
@@ -60,83 +59,107 @@ func newPlistData(title, exe, packageID, version, author string) *plistData {
...
 		return err
 	}
-	err = writer.Flush()
+	defer outfile.Close()
+
+	ico := windowsIcoHeader{
+		imageType:  1,
+		imageCount: uint16(len(sizes)),
+	}
+	err = binary.Write(outfile, binary.LittleEndian, ico)
```
leaanthony, line 108: If we write out here and error later, then we'll have corrupt .ico files written. Suggest moving this to the end once we know everything has been generated.

dedo1911: I got the point. we should just delete the corrupted file (if any) once this method exits with an error.

cmd/package.go
```diff
@@ -100,7 +100,7 @@ func generateWindowsIcon(pngFilename string, iconfile string) error {
 	for _, size := range sizes {
 		rect := image.Rect(0, 0, int(size), int(size))
 		rawdata := image.NewRGBA(rect)
-		scale := draw.ApproxBiLinear
+		scale := draw.CatmullRom
```
dedo1911, line 103: I noticed that scaling with NearestNeighbor was bad for small sizes (ApproxBiLinear uses   NearestNeighbor too).
CatmullRom is even better than BiLinear.

Sample comparison CatmullRom vs ApproxBiLinear:
sample_icons

leaanthony:
Meant to say before merging: Please add yourself to the CONTRIBUTORS.MD file :+1:

Issue #352, opened 2020-03-03
How I can set the application icon


drharryhe:
as the title, i want to replace the default applicatioin shown on the left-top corner of the app with my own icon.
thx

leaanthony:
Not possible currently.  What OS are you targetting?

drharryhe:
I am working on win10, but target Windows and MacOS

drharryhe:
well, thanks, I will take a try

leaanthony:
@drharryhe Any updates?

amelkikh:
> well, thanks, I will take a try

Hi! Could you show your solution?

AddictArts:
I've done this on Mac and Windows.

Mac

`replace appicon.png with your desired image`

Windows

`replace <appname>.ico with your desired icon`

In both cases rebuild the app `wails build -p` after your first build.

olinord:
I also have this issue, by following the instructions given with the `wails build -p` I was able to change the icon of the .exe file.

However it did not change the icon in the window frame (of the application when it os run), that is the top left corner of the window frame. Is there any way of making that possible? 
I’m running windows with Wails 1.0.2.

leaanthony:
This was fixed by #424 - Can anyone confirm before closing?

Aceix:
has this `-p` flag been replaced in v2.3.1?

jhwhu:
## Windows

The `windows` directory contains the manifest and rc files used when building with `wails build`.
These may be customised for your application. To return these files to the default state, simply delete them and
build with `wails build`.

- `icon.ico` - The icon used for the application. This is used when building using `wails build`. If you wish to
  use a different icon, simply replace this file with your own. If it is missing, a new `icon.ico` file
  will be created using the `appicon.png` file in the build directory.

TheLambdaFunction:
What about Linux? This issue is to be reopened because setting the Option for the linux side of things does NOT set the icon

leaanthony:
That's because Linux doesn't have a way of doing it (unless things changed). Linux requires hosting the icon outside the binary and reference it in a .desktop file which is why setting it in the application options doesn't work.

TheLambdaFunction:
Then the `wails dev` and `wails build` commands should generate one and use the assets akin to what is beign done for the other platforms. The documentation should also state this, right now this issue is the only place on the web where the correct procedures are described

leaanthony:
[...]

PR #424, opened 2020-05-25
Fix window icon on Windows

dedo1911:
Should fix the request at issue #352


leaanthony:
Does this add the icon to the window corner?

dedo1911:
Yes Sir! this is my window corner at the moment:
immagine

leaanthony:
Fantastic! Please stick around! We need someone with wizard Windows skills :-)

Issue #436, opened 2020-06-15
Allow dialog file filtering by extension name


fdidron:
**Description**

Currently, when opening a file selection dialog, there is no way to specify a file extension to filter the files that can be selected.

I have an [app running on Windows, OS X and Linux](https://ergodox-ez.com/pages/wally) that uses a fork of webview to do that.

That's the only thing that's preventing me to use Wails, so if that seems like a good idea, I'm happy to work on a PR.

leaanthony:
Hi @fdidron. That sounds awesome! Please go ahead. We work off the develop branch and then merge to ~~master~~ main for stable releases. Thanks!

leaanthony:
NOTE TO SELF: Update the API docs.

fdidron:
I see that the docs were updated, closing this issue now

PR #438, opened 2020-06-16
feat: add file extension filtering to select/save dialogs

fdidron:
References #436 

TODO
- [x] Implement with GTK
- [ ] Implement with WinAPI
- [ ] Implement with Cocoa
- [ ] Update docs

To add a file extension filter, you need to pass a comma delimited string to `SelectFile` / `SelectSaveFile` :

```go
s.runtime.Dialog.SelectSaveFile("*.jpg,*.png")
```

I added a variadic parameter to these functions to avoid a breaking change. So if you don't need any filter you can just dismiss the params.
```go
s.runtime.Dialog.SelectSaveFile()
```

@leaanthony This is an implementation for GTK only, I will work on WinAPI / Cocoa if you think the current implementation is right.

### DIFF
--- a/lib/interfaces/renderer.go
+++ b/lib/interfaces/renderer.go
@@ -17,9 +17,9 @@ type Renderer interface {
 	NotifyEvent(eventData *messages.EventData) error
 
 	// Dialog Runtime
-	SelectFile() string
+	SelectFile(filter string) string
 	SelectDirectory() string
-	SelectSaveFile() string
+	SelectSaveFile(filter string) string
 
 	// Window Runtime
 	SetColour(string) error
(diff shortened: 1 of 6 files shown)


fdidron:
@leaanthony Actually using the params, maybe now is the good time to also allow setting the dialog tiltle setting:

```go
SelectFile(title string, filter string) string

//Default title "Select File", no filters
SelectFile()
//Custom title, no filters
SelectFile("It would be nice to select a file now")
//Default title "Select File", filters pdf
SelectFile("", "*.pdf")
```

leaanthony:
This is great! Thanks for making it variadic. Totally don't want to break the API. Merging this in.

fdidron:
@leaanthony  Waaaaaaiiiiitttt :) I need to make it work on OSX and Windows, right now it only works on Linux.

leaanthony:
I should have said: yes, title would be great too...

NOTE TO SELF: Update https://github.com/wailsapp/docsv1/blob/master/content/reference/runtime.md before closing this ticket.

leaanthony:
I'm ok with that. It's not a breaking change. Besides, no pressure ;-)

PR #439, opened 2020-06-17
436 file dialog filter for win32 and cocoa

fdidron:
Reference #436 @leaanthony This is the remaining stuff related to #438 , support for OSX / Windows

### DIFF
--- a/lib/renderer/webview/webview.h
+++ b/lib/renderer/webview/webview.h
@@ -1792,7 +1792,7 @@ struct webview_priv
   WEBVIEW_API void webview_dialog(struct webview *w,
                                   enum webview_dialog_type dlgtype, int flags,
                                   const char *title, const char *arg,
-                                  char *result, size_t resultsz)
+                                  char *result, size_t resultsz, char *filter)
   {
     if (dlgtype == WEBVIEW_DIALOG_TYPE_OPEN ||
         dlgtype == WEBVIEW_DIALOG_TYPE_SAVE)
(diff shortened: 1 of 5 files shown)


lib/renderer/webview/webview.h
```diff
@@ -2213,12 +2239,16 @@ struct webview_priv
...
+                                  char *result, size_t resultsz, char *filter)
   {
     if (dlgtype == WEBVIEW_DIALOG_TYPE_OPEN ||
         dlgtype == WEBVIEW_DIALOG_TYPE_SAVE)
     {
       NSSavePanel *panel;
+      NSString *filter_str = [NSString stringWithUTF8String:filter];
+      filter_str = [filter_str stringByReplacingOccurrencesOfString:@"*."
+                                     withString:@""];
+      NSArray *fileTypes = [filter_str componentsSeparatedByString:@","];
```
leaanthony, line 2251: Do we need to trim spaces? Will it accept `*.mpg, *.mpeg `?

lib/renderer/webview/webview.h
```diff
@@ -1832,6 +1832,32 @@ struct webview_priv
                     FOS_NOTESTFILECREATE | FOS_NODEREFERENCELINKS |
                     FOS_FORCESHOWHIDDEN | FOS_DEFAULTNOMINIMODE;
       }
+      if (filter[0] != '\0')
+      {
+        int count;
+        int i=0;
+        char* token;
+        char* filter_dup = strdup(filter);
+        for (count=1; filter[count]; filter[count]==',' ? count++ : *filter++);
```
leaanthony, line 1841: Out of interest, why does count start at 1?

lib/renderer/webview/webview.h
fdidron: Good point will add trimming in Goland since I think it should be applied to all platforms

lib/renderer/webview/webview.h
fdidron: This counts the number of tokens when splitting the filter string. It is used to size the array used to contain the filter structs required by win32's dialog. So if there's only one filter (no comma) we need an array size of 1.

leaanthony:
Awesome job!

PR #457, opened 2020-07-08
refactor home and config dir to stdlib, update min go ver to 1.13

konez2k:
- Remove mitchellh/go-homedir dependency in favor of the stdlib os.UserHomeDir (supported since 1.12)

if we can move the min go version to 1.13, we can introduce the following improvements too:

- Use os.UserConfigDir as configuration path (respects XDG user dirs with fallback on Linux, `%AppData%` on windows and `$HOME/Library/Application Support` on mac)
- Rename the config directory to wails (since its now inside a subdirectory and not on $home)

### DIFF
--- a/.github/workflows/latest-pre.yml
+++ b/.github/workflows/latest-pre.yml
@@ -13,10 +13,10 @@ jobs:
         os: [ubuntu-latest, windows-latest, macOS-latest]
     steps:
 
-    - name: Set up Go 1.12
+    - name: Set up Go 1.13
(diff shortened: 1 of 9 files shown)


leaanthony:
Hi @konez2k :-) 
This is a great idea. I'm ok with moving to Go 1.13 and we probably should have by now.

konez2k:
great! will update this PR asap

konez2k:
@leaanthony ready for review

leaanthony:
Awesome job!

Issue #469, opened 2020-08-01
Runtime emit can't handle long messages


lyimmi:
**Description**

If the emitted message is too long it causes a javascript error. I'm converting the messages to bas64 so I don't have to hassle with escaping the strings for the js side.

**To Reproduce**
Steps to reproduce the behaviour:
 1. Emitt a long string from go to js. 

**Expected behaviour**
Get the emitted message from go.

**Screenshots**
![Screenshot from 2020-08-01 23-52-37](https://user-images.githubusercontent.com/8627125/89111259-e6562200-d453-11ea-8723-08fa1543a7a5.png)

**System Details**

| Name   | Value |
| ----- | ----- |
| Wails Version | v1.7.1 |
| Go Version    | go1.14.5 |
| Platform      | linux |
| Linux      | Ubuntu 20.04|
| Arch          | amd64 |
| GO111MODULE   | on |
| GCC           | 9.3.0 |
| Npm           | 6.14.7 |
| Node          | v14.7.0 |

**Additional context**
What is in the string does not matter, if you send spaces the error is the same.

`t.runtime.Events.Emit("stats", uEnc)`

This is one of my messages that ends up jumbled:
docker stats --no-stream --format "{{.Name}}\\t{{.CPUPerc}}\\t{{.MemUsage}}\\t{{.MemPerc}}"

[...]

PR #471, opened 2020-08-03
fix(bridge): fixed long message emit issue #469

Lyimmi:
A simple fix for emitting long messages with runtime. As discussed at #469.


leaanthony:
Awesome! Thanks!

Issue #419, opened 2020-05-21
Ivalid JSON data


andrewrynhard:
#####################################################
     If you have a technical issue, please do not open a bug this way! 
                      Please use the `wails issue` command!                          
#####################################################

**Description**
I am trying to send a kubernetes struct to the frontend using events. I get an error from here: https://github.com/wailsapp/wails/blob/94c56115a582082ec9205f4c7e1b4f2e50e077e5/runtime/js/core/events.js#L111

**To Reproduce**

**Expected behaviour**
I'd like to see the error. Looks like `e` is caught but never sent in the error description. It would be very useful to know the error.

**Screenshots**

**System Details**
Please provide your platform, GO version and variables, etc

**Additional context**

leaanthony:
Great call! I think it would be better to just prepend the error message of the original error and send that back. Thoughts?

leaanthony:
Something like:
```
e.message = 'Invalid JSON data sent to notify [Event name = ' + eventName +'] ' + e.message); 
```

leaanthony:
What makes this interesting to me is that this error is from an internal call from backend to frontend so I'd be really interested in how you triggered it.

leaanthony:
I've created branch https://github.com/wailsapp/wails/tree/419-better-json-error-reporting if you want to play with it

andrewrynhard:
Hmm. So I built our app with this and and the error seems empty. I am also having a tough time reproducing dependably. Give me a bit more time to dig.

PR #475, opened 2020-08-12
Fix double quotes encoding in event Data passed from go to JS in Emit

Unix4ever:
Way to reproduce:

```go
test := map[string]string{
        "key": `"value"`,
}

runtime.Events.Emit("test", &test)
```

Then in eval call JS will try to parse it and will fail:
```js
JSON.parse('{"key":"\"value\"}')
VM287:1 Uncaught SyntaxError: Unexpected token h in JSON at position 10
    at JSON.parse (<anonymous>)
    at <anonymous>:1:6
```

Proposed fix is to double encode it to json before evaluation. Not sure how it will affect performance.
In any case `\"` should be replaced with `\\"` somehow to make event sender more robust.

If you know any way to configure json marshaller to have it output proper data right away, feel free to close this PR.


Unix4ever:
https://github.com/wailsapp/wails/issues/419 that may be what triggered that issue

leaanthony:
Hey, thanks for this. I need to look into why this might be the case and the consequence of this change.

leaanthony:
String escaping has been one of the most tricky parts about this project! I'm now thinking that double encoding is wrong and potentially we should be looking at `json.RawMessage` as discussed in this article: https://goinbigdata.com/how-to-correctly-serialize-json-string-in-golang/

Unix4ever:
Thanks for the feedback.
I'm not sure I understand your suggestion... I don't get how RawMessage encoding can help to resolve this issue.
In your code you run JS eval function, where you inline your JSON data. Problem is that `JSON.parse` can't process JSON output generated by golang if any string contains double quotes.

https://play.golang.org/p/tZwvn7eND9X
```js
JSON.parse("{\"key\":\"\\\"value\\\"\"}")
{key: ""value""}
JSON.parse('{"key":"\"value\""}')
VM162:1 Uncaught SyntaxError: Unexpected token v in JSON at position 9
    at JSON.parse (<anonymous>)
    at <anonymous>:1:6
```

I've missed one change actually... No single quotes are required with my change.

Unix4ever:
And I missed conversion of `[]byte` to `string`... Fixed both issues.

tmclane:
@Unix4ever Is this solved then?

Unix4ever:
If you're fine with my way to solve it, then it's fixed :)

tmclane:
Well it would seem the `solution` you mentioned is not in this PR. 
I don't know how to tell if it was the `right` way to solve this or not.  But if you are literally embedding JSON inside JSON then yes the inner JSON `MUST` be double encoded.

Unix4ever:
Yep... Sorry for the confusion. This is actually a `solution` to the problem.

Unix4ever:
> But if you are literally embedding JSON inside JSON then yes the inner JSON MUST be double encoded.

Problem is not limited to having JSON string inside JSON object.
Any occurrence of " or ' in your JSON encoded output breaks event evaluation right now.
My fix handles both " and ' occurrence.

tmclane:
Your example is actually the correct output for a double quoted string.

 If you have `'"value"'` it will be seen as a string containing `"` characters.
When this is JSON encoded you will end up with "\"value\"" because that is the only way to embed a '"' character in JSON.
You are forcing a double decode of the value `key` element itself.

Why do you want to do that? 

If it didn't encode the '"' it would be producing invalid JSON.

Unix4ever:
I've hit the problem because I have JSON encoded strings in json object, but I actually wanted to point out that it may be triggered by just having any quotes in any strings in the object. So don't really treat this example as real use case scenario.

> Your example is actually the correct output for a double quoted string.

It's correct, but js eval doesn't parse it. That's the issue.

tmclane:
I don't believe we want to add error handling and correction for buggy applications.
The application developer should understand what is allowed to be sent and what isn't. 

Honestly I think the better solution here is to update the javascript side to better handle malformed json.

I'm not a Javascript expert but shouldn't JSON decoding be done with `JSON.parse`?
https://www.w3schools.com/js/js_json_parse.asp

That should handle embedded quotes correctly

Unix4ever:
Problem lies here:
```go
message := fmt.Sprintf("window.wails._.Notify('%s','%s')", event.Name, data)
...
err := session.evalJS(message, notifyMessage)
```

Let's imagine data is:
 ```json
{"message": "it's working"}
```

JSON evaluation will be:
```js
window.wails._.Notify('event', '{"message": "it's working"}')
```

It's just invalid code.

tmclane:
My point is that the values should be `json encoded` on the sending side.  If the Wails `.Notify` function isn't encoding it properly that is where this should be fixed.  Neither side should send invalid json to the other.

smira:
@tmclane with the existing code:

```go
fmt.Sprintf("window.wails._.Notify('%s','%s')", event.Name, data)
```

It's a pure RCE, as if I can control `data`, I can put there e.g. `foo'); alert("Ooops!); ('a` to cause alert to pop up

I think fix proposed by @Unix4ever is the right one: whenever code to be executed in JavaScript is built in Go, every argument to the function should go through `JSON` instead of relying on manual quoting.

tmclane:
`data` is supposed to be already JSON encoded. 
If you look on line 161 you will see the `json.Marshal` call.

```
	// Default data is a blank array
	data := []byte("[]")

	// Process event data
	if event.Data != nil {
		// Marshall the data
		data, err = json.Marshal(event.Data)
		if err != nil {
			h.log.Errorf("Cannot unmarshall JSON data in event: %s ", err.Error())
			return err
		}
	}

	message := "window.wails._.Notify('" + event.Name + "','" + string(data) + "')"
```

tmclane:
You are worried about an RCE in an application you are coding?

smira:
@tmclane you're putting this value (whatever it is, but it's a string which is result of JSON encoding) into Javascript code. basically you're generating Javascript code on the fly. And this should be done correctly, so value should be escaped properly.

`fmt.Sprintf('%s', value)` is not a proper way to escape value for Javascript, `json.Marshal` is a proper way though

it doesn't really matter what `value` is in this case - simple string or JSON-encoded something

I'm worried about RCE in this project and any app which builds on top of it

tmclane:
`data` can contain only two values:

 - []
 - JSON.Marshalled(Event.data) 

There is no way any other value could get in there unless the code is changed!  

The REAL ISSUE here is that you are expecting to be able to `eval` that value.  You shouldn't be doing that unless you know it is safe.

The code that receives the JSON formatted data is : https://github.com/wailsapp/wails/blob/f86996705b99faa172a70ac4da690a59d522b391/runtime/js/core/events.js#L109

You will note that it is doing a `JSON.parse(data)` on the received data.  This will properly handle the data being sent from the Golang side.

https://github.com/wailsapp/wails/blob/f86996705b99faa172a70ac4da690a59d522b391/lib/renderer/bridge/bridge.go#L160

I am not convinced you found a real issue at this point.

smira:
I'm not sure how to convince you:

```
>>> print(json.dumps("lala'; alert('oops'); '"))
"lala'; alert('oops'); '"
```

try putting this value into your `fmt.Sprintf('%s')` code and see what happens when this gets eval'd

smira:
> The REAL ISSUE here is that you are expecting to be able to eval that value. You shouldn't be doing that unless you know it is safe.

To make it clear: I do not expect my data to be eval'd.

It's the code 

```
 message := fmt.Sprintf("window.wails._.Notify('%s','%s')", event.Name, data) 
```

which builds Javascript code to be eval'd.

tmclane:
Would you please post a complete example showing this issue?

How do you pass in something that does something undesirable in the `Event.data` value?

smira:
It's in the description of this PR from @Unix4ever.

I just expanded on that to show that RCE is possible.

Unix4ever:
And working RCE snipped is:
```go
w.wailsRuntime.Events.Emit("test", "]'); alert('hi there'); //")
```

Imagine application processing some variable input and sending it to the frontend. 
If you don't have control, you may get anything there.

Yes, you're right. Application developer can escape the code on the higher level and never send this data down to Wails.
But I'm not sure why not to close this obvious potential vulnerability if it's possible.
Otherwise it just becomes another way for people to shoot in their leg while using this project.

tmclane:
Ok so I made an application using your example snippet that does execute the javascript sent to it by the backend in the `Emit` call.

Thank you for bearing with me!

I'll see if I can come up with a solution that seems a bit more elegant. 

Note: The backend code can always execute javascript on the frontend view anyhow.  So I don't know why you would go to the trouble of doing this when there are easier more direct methods.

lib/renderer/bridge/bridge.go
```diff
@@ -152,12 +152,20 @@ func (h *Bridge) NotifyEvent(event *messages.EventData) error {
...
 		data, err = json.Marshal(event.Data)
 		if err != nil {
-			h.log.Errorf("Cannot unmarshall JSON data in event: %s ", err.Error())
+			h.log.Errorf("Cannot marshal JSON data in event: %s", err.Error())
+			return err
+		}
+
+		// double marshalling is required to ensure that all double quotes are escaped properly for JS
+		// to run evalJS, otherwise any " in any string field in the object will cause evalJS to fail.
+		data, err = json.Marshal(string(data))
```
tmclane, line 161: This appears to actually base64 encode the data instead of the expected double marshalling.

See the following playground for an similar setup:
https://play.golang.org/p/v-4YrTH6DJr

tmclane, line 161: I meant to say if you leave off the `string()` conversion you can automatically get base64 encoding.

https://play.golang.org/p/v-4YrTH6DJr

Unix4ever: Right, but then it will require base64 decoding on Javascript side. Given that generated JSON is already being properly escaped, I think base64 is not needed here.

leaanthony:
Hi all. Thanks for raising this. Please bear with us - I'll review this on Friday and get a solution in place.

leaanthony:
@Unix4ever - I noticed this PR is for master. Could you please make this PR for `develop`? Thanks.

leaanthony:
@Unix4ever - I've created a new PR for this here: https://github.com/wailsapp/wails/pull/478

The solution needed to be applied to the non-serve version of the code too. I've added you to the contributors list as I consider this your fix. Please let me know if this works for you. Thanks :+1:

PR #484, opened 2020-08-30
Sync store

leaanthony:
This PR brings with it a new synchronised state store. It allows you to set up a data store in both Go and JS and writing to it will keep both in sync, as well as triggering callbacks to subscribers. It's heavily modelled on Svelte's concept of a store. To connect the frontend and backend stores, they must share the same ID. Example usage:

```js
const runtime = require('@wailsapp/runtime');

function start() {
	var mystore = runtime.Store.New('Counter');
...
	mystore.set(0); // Set the value in the store
	mystore.update( currentValue => currentValue + 1); // Increments the store value
	mystore.subscribe( (currentValue) => {
              console.log('store was updated with new value: ' + currentValue);
        })
```

```go
// Counter is what we use for counting
type Counter struct {
	r     *wails.Runtime
	store *wails.Store
}

// WailsInit is called when the component is being initialised
func (c *Counter) WailsInit(runtime *wails.Runtime) error {
	c.r = runtime
	c.store = runtime.Store.New("Counter")
	c.store.Set(0)
	c.store.Update( func(currentValue interface{}) interface{} {
		return process(currentValue)
	})
	c.store.Subscribe( func(currentValue interface{}) {
		fmt.Printf("Store was updated with new value: %+v\n", currentValue)
	})
	return nil
}
```

It's considered Beta at this stage (everything appears to work - needs more testing).

The Vanilla template has been updated to provide an example of this as it wasn't much beyond the simple hello world example.

`wails.Runtime` has been moved into its own package with it being aliased back into the `wails` package for compatibility reasons. `Store` has also been aliased into the `wails` package to make it appear as a first class type. 

## Steps to test

1. Checkout this PR branch
2. `cd scripts` + `go run build.go`
3. `cd runtime/js/runtime` + `sudo npm link` to get the updated runtime into global cache
4. Generate a new vanilla app. The build may fail!
5. `cd frontend` + `npm link '@wailsapp/runtime'`
6. `wails build -d` again

Once this is approved, I'll push the new module to npmjs. I believe it's backwards compatible.


leaanthony:
I'll fix the linting in a bit :facepalm:

runtime/store.go
```diff
@@ -0,0 +1,117 @@
...
+
+// New creates a new store
+func (p *StoreProvider) New(name string) *Store {
+
+	result := Store{
+		name:    name,
+		runtime: p.runtime,
+	}
+
+	result.setupListner()
```
tmclane, line 47: Typo?

leaanthony: Oooph good catch. Will tidy this PR.

leaanthony:
It's just an interface{} so you can use what you like. It's up to you to manage it mind. Eg all numbers from JS are treated as floats.

leaanthony:
I'm going to put this PR on hold. I've had some more thoughts around it and I'm starting to think that it should be initialised like bind and do automatic data transformation for you. Should be relatively easy.

leaanthony:
This updated PR contains a big change: Instead of `Update(func(interface{})interface{})` we now have `Update(interface{})`. This allows the developer to not worry about type conversions.

Previous:
```
// Decrement will decrement the counter
func (c *Counter) Decrement() {

        decrement := func(data interface{}) interface{}  {      
                switch value := data.(type) {
                case float64:
                        return int(value) -1
                default:
                        return value.(int) -1
        }

        // Update the store using the decrement function
        c.store.Update(decrement)
}
```

Now:
```
// Decrement will decrement the counter
func (c *Counter) Decrement() {

        decrement := func(data int) int {
                return data - 1
        }
        // Update the store using the decrement function
        c.store.Update(decrement)
}
```
This also supports pointers to structs!

Some caveats:

* The type of the input and return values in the function must be the same as the type of the default value of the store. If you don't you'll get a runtime error. It currently `log.Fatal`s but happy to change this to some other kind of error. But honestly, your signature is wrong. You should stop. Now.
* This supports structs but any updates from the frontend MUST be parsable by Go's json package. If it isn't, an error is passed to your error handler (set using `OnError`). If you don't set this error handler, it's going to silently fail. Any better suggestions?

leaanthony:
@tmclane his is now finalised and ready to Go! Managed to get a go2go version working for the Update methods which massively simplifies the way the store works. We wouldn't need reflection at all...

Issue #497, opened 2020-09-09
App is a black screen after building but works fine in development


artooro:
**Description**

I just built a new Wails app using the Angular template. When running in dev mode using wails serve and npx ng serve it works great. But if I build it using wails build or wails build -p it is just a black window. There is nothing in standard output and I can't find any useful troubleshooting info in the console.

**Screenshots**
![Screen Shot 2020-09-09 at 09 07 10](https://user-images.githubusercontent.com/393395/92602327-01e6f080-f27c-11ea-9f8f-0176d96fb3f2.jpg)

**System Details**

| Name   | Value |
| ----- | ----- |
| Wails Version | v1.7.1 |
| Go Version    | go1.15 |
| Platform      | darwin |
| Arch          | amd64 |
| GO111MODULE   | (Not Set) |
| GCC           | 4.2.1 |
| Npm           | 6.14.7 |
| Node          | v14.9.0 |

**Additional context**
In development the app works fine in both Safari and Chrome. If there is a way for me to see any logs from the built app, that would be great to know and be documented.

leaanthony:
What version of MacOS? 10.15.5 works fine:

<img width="1214" alt="image" src="https://user-images.githubusercontent.com/1943904/92644502-cfde9880-f326-11ea-98cd-11ec11f87d3c.png">

leaanthony:
v10.15.6
<img width="1394" alt="image" src="https://user-images.githubusercontent.com/1943904/92648053-3a460780-f32c-11ea-8f7f-089bf0745bd4.png">

leaanthony:
Can confirm `wails build -p` works too

artooro:
@leaanthony yeah the raw template works fine, but the app I built from the template does not. I was hoping there would be some type of log from the built binary that would show me what is going on.

leaanthony:
[667809ih[][k[]';.;llll](url)

leaanthony:
Ah, building with -d will dump logs when run from terminal

artooro:
Ah thanks, I missed that flag. So with the debug logs everything looks nominal so unfortunately doesn't help me resolve the actual issue. I'll have to rip out pieces of my code until it's narrowed down.

artooro:
Spoke too soon, looks like in debug mode it's also possible to open the javascript console in the app window, where I noticed this error. "Unhandled Navigation Error".
[...]

PR #500, opened 2020-09-10
resolve angular routing broken when app is built

artooro:
As discovered in issue #497 there is an angular routing issue. This pull request fixes the template to use hash based routing to workaround the issue.


leaanthony:
This is great! Thanks. Will review.

leaanthony:
Please PR to `develop`. Thanks.

Issue #460, opened 2020-07-16
Cross compilation using xgo does not copy local (unpublished) packages


DanielKrolopp:
**Description**

I've been trying to cross compile for windows from my Linux machine and ran into an issue. I have a go module in a different directory than the one in which my wails project is located---I import this module in a go.mod file in my wails project directory using the replace keyword. When I try to compile, I get the following message: 
```
Compiling for windows-4.0/amd64...
go: github.com/username/projectname/client@v0.0.0-00010101000000-000000000000: parsing /home/dk/go/src/github.com/username/projectname/client/go.mod: open /home/dk/go/src/github.com/username/projectname/client/go.mod: no such file or directory
```

I suspect that the output is from the Docker container, and that the necessary files are not being copied in.

**To Reproduce**
Steps to reproduce the behaviour:
1. Import a local go module outside the immediate project directory using go modules' `replace` keyword from within the go.mod of your wails project directory.
2. Run `wails build -x windows/amd64 --verbose`. The error described above should appear.
3. Wails will claim that the project built successfully, but the build folder will be empty.

**Expected behaviour**
The cross-compilation builds a functioning windows executable.

**System Details**

| Name   | Value |
| ----- | ----- |
| Wails Version | v1.7.1 |
| Go Version    | go1.14.2 |
| Platform      | linux |
| Arch          | amd64 |
| GO111MODULE   | (Not Set) |
| GCC           | 7.5.0 |
| Npm           | 6.14.4 |
| Node          | v12.17.0 |

leaanthony:
The project directory is mounted for the build so if it's not in there then yes, it won't get picked up. What do you think would be a good solution to this?

tmclane:
This is a tricky one.  If you make the go module accessible over the internet go itself will copy it into the build container.
If it is private it is a bit harder.

One thought i had would be to:
- optionally mount the GOPATH directory into the build container at `/go`.

leaanthony:
That's not a bad idea @tmclane! We could then add it to $GOPATH. Trying to think how this works with modules.

tmclane:
Well as long as you place the module in the GOPATH it will work fine.

Perhaps we could allow additional mount mappings to help place them appropriately inside the container.

leaanthony:
I guess the problem is that `go.mod` can reference any arbitrary location on disk. What about doing `go mod vendor` first to pull it into the project directory? I know wails itself has an issue with that so potentially remove that directory before compiling

[...]

PR #508, opened 2020-09-13
implement gopath handling during cross compilation

tmclane:
Allows the developer to specify a `-gopath` path argument during build which is mounted to the docker container at `/go` which is configured as the GOPATH inside the runtime.

Allows the developer to specify a `-gopath` path argument during build which causes the location on your host to be mounted at `/go` inside the cross-compiler container.  
Enabling the `replace github.com/original/library => github.com/tmclane/library` style of substitution where you edit files on the host and they are made available in the cross-compile environment.

Addresses: #460

### DIFF
--- a/cmd/helpers.go
+++ b/cmd/helpers.go
@@ -18,6 +18,8 @@ import (
 	"github.com/leaanthony/spinner"
 )
 
+const XGO_VERSION = "latest"
+
 var fs = NewFSHelper()
 
 // ValidateFrontendConfig checks if the frontend config is valid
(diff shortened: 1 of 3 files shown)


tmclane:
Just realized I likely need to make this a sub-command of the `-x` cross-compile sub-command of the `build` command.

leaanthony (commented):
Beast mode!

A couple of Qs:

- Did you pull out xgoVersion for $FUTURE_REASONS?
- Is `/go` known to the container? As in, does it hold some significance currently or do we need to update the container too?

tmclane:
/go is the default go path inside the container if not overridden.

tmclane:
I am planning on using xGoVersion to hold the version of the wailsapp/xgo image tag.  Currently it only has latest so of limited value. I want to switch to versioned images so we know what capabilities are available in each one .

leaanthony:
This is sweet.

PR #522, opened 2020-10-09
feat: Vue3, Vue Router, Vuex, and Typescript Template

kmuchmore:
Adds a Vue3 Typescript template with Vuex and Vue Router.


leaanthony:
I think I need to put a bullet in the hound.

leaanthony:
Sorry for the slow response on this. What platforms have you tested this on?

cmd/templates/vue3-full/frontend/src/views/Home.vue
```diff
@@ -0,0 +1,40 @@
...
+export default defineComponent({
+  name: "Home",
+  components: {
+    HelloWorld,
+  },
+  setup() {
+    
+    const message = ref("Click the Icon");
+
+    const getMessage = () => {
```
leaanthony, line 31: Without a polyfill, arrow functions will cause the windows builds to fail.

leaanthony:
Can confirm this doesn't work on windows:



Digging into the docs and issue tracker, it seems that IE11 support is not available which puts me in a predicament. I don't really want to be fragmenting the templates but I don't want to hold V1 back whilst V2 is being developed. Before merging this I may need to only present templates that are available for a given platform... This would need to be handled by both `init` and `build`.

leaanthony:
I've just merged #528 so that we can get this through. Could you please rebase your PR on develop and update your template to use the `Platforms` key to indicate the supported platforms. The best way of testing is set it to be `["windows"]` then see if `init` and `build` throw warnings. Thanks 👍

leaanthony:
@kmuchmore - any updates? Cheers 👍

kmuchmore:
I won’t have time until probably this weekend to look at this. Sorry

leaanthony:
No rush and thanks for this awesome contribution!

leaanthony:
I'm going to merge this as-is then update it to set the verified platforms.

PR #533, opened 2020-10-20
feat: tags build stage

gelleson:
Added field tags for forwarding at the build stageю

### DIFF
--- a/cmd/helpers.go
+++ b/cmd/helpers.go
@@ -248,6 +248,10 @@ func BuildNative(binaryName string, forceRebuild bool, buildMode string, project
 
 	buildCommand.AddSlice([]string{"-ldflags", ldFlags(projectOptions, buildMode)})
 
+	if projectOptions.Tags != "" {
+		buildCommand.AddSlice([]string{"--tags", projectOptions.Tags})
+	}
+
 	if projectOptions.Verbose {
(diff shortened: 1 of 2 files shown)


gelleson:
it's the same pull request but more clean.

cmd/helpers.go
```diff
@@ -143,6 +143,7 @@ func BuildDocker(binaryName string, buildMode string, projectOptions *ProjectOpt
 		"-v", fmt.Sprintf("%s:/build", filepath.Join(fs.Cwd(), "build")),
 		"-v", fmt.Sprintf("%s:/source", fs.Cwd()),
 		"-e", fmt.Sprintf("LOCAL_USER_ID=%v", userid),
+		"-e", fmt.Sprintf("FLAG_TAGS=\"%s\"", projectOptions.Tags),
```
leaanthony, line 146: I don't think the escaped quotes are needed here from what I can see https://github.com/karalabe/xgo/blob/c5ccff8648a77686d274c1984e1cd6319736f5cb/xgo.go#L297

leaanthony (changes requested):

cmd/helpers.go
gelleson: I have fixed it :)

leaanthony:
Thanks!

PR #537, opened 2020-10-25
Update Vuetify in package.json

aayush420:


leaanthony:
Please provide more information. Have you tested this on different platforms etc?

aayush420:
@leaanthony  No, didn't test a thing. Just saw that the latest version was greater than the one in use. Hence created a PR. Do you want me to run any specific tests?

leaanthony:
Please check the default template works 😃

aayush420:
Yeah It does. In fact, I've created a PR only after modifying my local app. No errors are reported in Dev console. No errors on `wails serve` or `build`. And the prod version is also working as usual.

PR #586, opened 2021-01-08
✨ Allow selecting of multi files/folders

### LINKED ISSUE #584: Question: Is not possible to open file dialog to select multiple files ?
I only see option to select directory or select single file.  `Renderer` interface also doesn't have method like selectFiles

```
type Renderer interface {
	Initialise(AppConfig, IPCManager, EventManager) error
	Run() error

	// Binding
	NewBinding(bindingName string) error

	// Events
	NotifyEvent(eventData *messages.EventData) error

	// Dialog Runtime
	SelectFile(title string, filter string) string
	SelectDirectory() string
	SelectSaveFile(title string, filter string) string

	// Window Runtime
	SetColour(string) error
	Fullscreen()
	UnFullscreen()
	SetTitle(title string)
	Close()
}
```

Is it not implemented in wails or is not implemented in webview itself ?

dev-drprasad:
closes #584

### DIFF
--- a/lib/renderer/webview/webview.h
+++ b/lib/renderer/webview/webview.h
@@ -1879,7 +1879,7 @@ struct webview_priv
           token = strtok(NULL, ",");
           i++;
         }
-        printf("ls\n");
+
         for (int i=0; i < count; i++) {
           wchar_t *wFilter = (wchar_t *)malloc(4096);
           MultiByteToWideChar(CP_ACP, 0, filters[i], -1, wFilter, 4096);
@@ -1904,12 +1904,11 @@ struct webview_priv
       {
         goto error_dlg;
       }
-      printf("ls\n");
       if (flags & WEBVIEW_DIALOG_FLAG_MULTIPLE) {
         if (dlg->lpVtbl->GetResults(dlg, &multires) != S_OK) {
           goto error_dlg;
         }
-        DWORD count;
+        /* DWORD count;
         multires->lpVtbl->GetCount(multires,&count);
         for (DWORD i = 0; i < count; i++) {
           IShellItem *shellitem;
@@ -1918,7 +1917,7 @@ struct webview_priv
           shellitem->lpVtbl->GetDisplayName(shellitem, SIGDN_FILESYSPATH, &filepath);
           wcscat(ws, filepath);
           CoTaskMemFree(filepath);
-        }
+        } */
       } else {
         if (dlg->lpVtbl->GetResult(dlg, &res) != S_OK) {
           goto error_dlg;
@@ -1931,6 +1930,7 @@ struct webview_priv
       strncpy(result, s, resultsz);
       result[resultsz - 1] = '\0';
       CoTaskMemFree(ws);
+      return;
     error_result:
       res->lpVtbl->Release(res);
       multires->lpVtbl->Release(multires);
--- a/lib/renderer/webview/webview.h
+++ b/lib/renderer/webview/webview.h
@@ -121,6 +121,7 @@ struct webview_priv
 
 #define WEBVIEW_DIALOG_FLAG_FILE (0 << 0)
 #define WEBVIEW_DIALOG_FLAG_DIRECTORY (1 << 0)
+#define WEBVIEW_DIALOG_FLAG_MULTIPLE (2 << 0)
 
 #define WEBVIEW_DIALOG_FLAG_INFO (1 << 1)
 #define WEBVIEW_DIALOG_FLAG_WARNING (2 << 1)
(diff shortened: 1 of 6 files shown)


leaanthony (commented):
Thanks for this! There's a lot more to consider like the `runtime.Dialog.Open` method. Perhaps create a new method `runtime.Dialog.OpenMultiple` or an optional Options field to the current Open? Thoughts? Also, please rebase the PR on `develop`. Thanks!

dev-drprasad:
`runtime.Dialog` doesn't have method like `Open`. Do you mean `SelectFile` ?

Or do you mean `renderer.Dialog` ?

dev-drprasad:
Currently we have `webview.Dialog` which returns `string`

I would like to change return type to interface: `func (w *webview) Dialog(...) interface{}`

Create two more methods to use inside `renderer`'s `SelectFile` and `SelectFiles`:

```
func (w *webwview) DialogOpen(...) string
func (w *webwview) DialogOpenMultiple(...) []string
```

What do you think ?

dev-drprasad:
I made changes to reflect what my idea is. I changed interface: `WebView`, now `Dialog` returns `interface{}`. I am not sure whether this is good and might be break anything. Let me know your opinion on this and what do you recommend ?

leaanthony:
This looks ideal! I'll see if I can find some time to check it out today. 

Regarding the return type, that's the main issue we have always faced! The API should always have been `[]string` but changing that now is problematic for a lot of people. I think leaving it as it currently is is ok. If people want multiple files they can call the new `DialogOpenMultiple`. Perhaps we could add something that detects whether a call to `Dialog` is trying to use the multiple files flag and just return the first value and log a warning? That way the API doesn't change and the user knows when they've made a mistake?

dev-drprasad:
Makes sense. I will revert interface change.

I realized real problem is getting array of strings from C to Go. I think I found a solution. I implemented that solution in Cocoa's  webview API (OS X). I recommend you to test/review that part. I will do implement in others tomorrow.

lib/renderer/webview/webview.h
```diff
@@ -2298,8 +2308,22 @@ struct webview_priv
...
+              [[url path] cStringUsingEncoding:NSUTF8StringEncoding];
+          int len = strlen(cstr);
+          char *cstr_copy = (char *)malloc(sizeof(char) * (len + 1));
+          strcpy(cstr_copy, cstr);
+          paths[i] = cstr_copy;
+          i++;
+        }
+        paths[i] = NULL;
+
+        memcpy(result, paths, resultsz);
```
dev-drprasad, line 2325: I will rename variables to match  their purpose later

lib/renderer/webview/webview.h
```diff
@@ -2298,8 +2308,22 @@ struct webview_priv
...
-        const char *filename = [[[panel URL] path] UTF8String];
-        strlcpy(result, filename, resultsz);
+        /* https://stackoverflow.com/a/15318065/6748719 */
+        int filesCount = [[panel URLs] count];
+        char **filePaths = (char **)malloc(sizeof(char *) * (filesCount + 1));
+        int i;
+        for (id url in [panel URLs]) {
+          const char *filePath = [[url path] UTF8String];
+          int len = strlen(filePath);
+          char *filePathCopy = (char *)malloc(sizeof(char) * (len + 1));
```
leaanthony, line 2318: I think this is redundant. `[[url path] UTF8String]` allocates the extra memory so you could just store those directly into filepaths. We also need to free this memory too 😄

lib/renderer/webview/webview.h
```diff
@@ -2298,8 +2308,22 @@ struct webview_priv
...
+          strcpy(filePathCopy, filePath);
+          filePaths[i] = filePathCopy;
+          i++;
+        }
+        filePaths[i] = NULL;
+
+        memcpy(result, filePaths, resultsz);
+        memcpy(resultlen, &filesCount, sizeof(filesCount));
```
leaanthony, line 2326: An alternative approach to this is just have a string that uses a `\n` delimiter to split the filenames up. You just keep appending to it.

dev-drprasad: I agree `\n` separated strings is simple solution. But AFAIK, `\n` valid character inside filename. Anyway I will go with your suggestion cuz i don't think people use `\n` inside filenames

leaanthony:
I don't think it is but we should check.

lib/renderer/webview/webview.h
```diff
@@ -1881,20 +1904,36 @@ struct webview_priv
...
-      {
-        goto error_dlg;
-      }
-      if (res->lpVtbl->GetDisplayName(res, SIGDN_FILESYSPATH, &ws) != S_OK)
-      {
-        goto error_result;
+      if (flags & WEBVIEW_DIALOG_FLAG_MULTIPLE) {
+        if (dlg->lpVtbl->GetResults(dlg, &multires) != S_OK) {
+          goto error_dlg;
+        }
```
dev-drprasad, line 1910: @leaanthony I tried for like hours and I couldn't get it work in windows. I think its panic-ing at `GetResults`. See if you can help me with windows implementation. I tried my best

leaanthony: I think I found the issue and boy, it's subtle! Look what the docs say about GetResult:
Screen Shot 2021-01-17 at 6 21 42 am
Looks like we need to use `GetResults()`!

dev-drprasad: I think you missed little `s` :). Indeed I am using `GetResults`. 

I think `GetResults` is method of interface `IFileOpenDialog` not `IFileDialog`. But I actually tried with `IFileOpenDialog` also in my local (by replacing all instances of `IFileDialog` with `IFileOpenDialog`), still same issue happened (I am not sure the way i did is right).


If you think its too complicated and we're better off without this feature, I am totally OK with it. We can drop this PR. I found https://github.com/ncruces/zenity which can be used to select multiple files. I am using it in my project without an issue.

Issue #592, opened 2021-01-23
Wails init using Svelte fails


SophieAu:
**Description**

When trying to initialize a new project using the Svelte template, the frontend build fails with what looks like a bundler error.

**To Reproduce**
Steps to reproduce the behaviour:
1. run `wails init` selecting Svelte (3) as the frontend template
4. See below for the error that appears

**System Details**

| Name   | Value |
| ----- | ----- |
| Wails Version | v1.11.0 |
| Go Version    | go1.15.5 |
| Platform      | darwin |
| Arch          | amd64 |
| GO111MODULE   | on |
| GCC           | 12.0.0 |
| Npm           | 6.14.9 |
| Node          | v14.4.0 |

**Additional context**
This is the error message:

```
wails init                                                                                                                                                                             
Wails v1.11.0 - Initialising project

The name of the project (My Project): test-project
Project Name: test-project
The output binary name (test-project): 
Output binary Name: test-project
Project directory name (test-project): 
Project Directory: test-project
Please select a template (* means unsupported on current platform):
  1: Angular - Angular 8 template (Requires node 10.8+)
  2: React JS - Create React App v4 template
  3: Svelte - A basic Svelte template
  4: Vanilla - A Vanilla HTML/JS template
  5: Vue3 Full - Vue 3, Vuex, Vue-router, and Webpack4
  6: Vue2/Webpack Basic - A basic Vue2/WebPack4 template
  7: Vuetify1.5/Webpack Basic - A basic Vuetify1.5/Webpack4 template
  8: Vuetify2/Webpack Basic - A basic Vuetify2/Webpack4 template
Please choose an option [1]: 3
Template: Svelte
✓ Generating project...
⣷ Building project (this may take a while)...
Wails v1.11.0 - Building Application

✓ Ensuring frontend dependencies are up to date (This may take a while)
⢿ Building frontend...
src/main.js → public/build/bundle.js...
(!) Error when using sourcemap for reporting an error: Can't resolve original location of error.
src/main.js: (5:12)
(!) Error when using sourcemap for reporting an error: Can't resolve original location of error.
src/main.js: (16:7)
(!) Error when using sourcemap for reporting an error: Can't resolve original location of error.
node_modules/svelte/internal/index.mjs: (41:12)
(!) Error when using sourcemap for reporting an error: Can't resolve original location of error.
node_modules/svelte/internal/index.mjs: (1830:9)
  4017 | 
[...]

PR #593, opened 2021-01-24
Fix wails init failing using the Svelte template

SophieAu:
As described in #592 , the `spec` option leads to a failing frontend. Removing it fixes this and seems to not break anything.


tmclane (approved):

leaanthony (approved):

leaanthony:
Perfect @SophieAu ! Thank you for doing this.

PR #600, opened 2021-02-04
Prevent running expired listeners

GeekStocks:
Adding an expired check around the running of the listener would prevent expiration problems or race conditions from affecting the behaviour.


GeekStocks:
To be clear, more work needs to follow on why the expiration is not functioning as expected here. However, to not use listener.expired to prevent problems is a missed opportunity.

GeekStocks:
I am closing this PR because I made a mistake by asking for it to go to `wailsapp:master` - I should have been more careful. I will resubmit to `v1.12.0-pre2` later on.

PR #604, opened 2021-02-05
Handle event expirations immediately

GeekStocks:
_Basically, the concept of `expired` has gone away in favor of just dealing with the expiration task immediately. This results in simplified code that doesn't allow expired listeners to pile up in the slices._

Change notes:

1)	`eventListener` struct no longer needs `expired` field because we are taking care of the expiration immediately. Also brought to near top of file instead of being buried in the struct methods.

2)	Code reviewed the entire file to ensure no `expired` logic anywhere else.

3)	Used “clone and truncate” method of slice element removal to handle the expiration immediately. There are references and a playground snippet in the comments to demonstrate this for the uninitiated and to demystify this complexity.

4)	tightened the `mutex` closer to the action for concurrent performance reasons.

5)	alphabetized `struct` fields for quicker locating; same for methods.

6)	added `manager_test.go` to prove this logic out. **NEED HELP GETTING STARTED**. Please see `TestLifecycle` for details. I’d like to write some deep tests but setting up the test instance is giving me fits.

### DIFF
--- a/lib/event/manager.go
+++ b/lib/event/manager.go
@@ -14,11 +14,11 @@ type Manager struct {
 	incomingEvents chan *messages.EventData
 	listeners      map[string][]*eventListener
 	log            *logger.CustomLogger
+	mu             sync.Mutex
 	quitChannel    chan struct{}
 	renderer       interfaces.Renderer // Messages will be dispatched to the frontend
 	running        bool
 	wg             sync.WaitGroup
-	mu             sync.Mutex
 }
 
 // NewManager creates a new event manager with a 100 event buffer
@@ -32,19 +32,13 @@ func NewManager() interfaces.EventManager {
 	}
 }
 
-// PushEvent places the given event on to the event queue
-func (e *Manager) PushEvent(eventData *messages.EventData) {
-	e.incomingEvents <- eventData
-}
-
 // eventListener holds a callback function which is invoked when
-// the event listened for is emitted. It has a counter which indicates
-// how the total number of events it is interested in. A value of zero
+// the event being listened for is emitted. It has a counter which
+// indicates the total number of events to allow. A value of zero
 // means it does not expire (default).
 type eventListener struct {
 	callback func(...interface{}) // Function to call with emitted event data
 	counter  uint                 // Expire after counter callbacks. 0 = infinite
-	expired  bool                 // Indicates if the listener has expired
 }
 
 // Creates a new event listener from the given callback function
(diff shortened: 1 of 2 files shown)


lib/event/manager.go
```diff
@@ -137,44 +141,44 @@ func (e *Manager) Start(renderer interfaces.Renderer) {
...
+							// this listener has just expired; remove it NOW
+							// see fast method: https://yourbasic.org/golang/delete-element-slice/
+							// https://play.golang.org/p/j0YjKUN0NL1
+							lastIndex := len(e.listeners[event.Name]) - 1
+
+							// overwrite expired currentIndex listener with lastIndex listener
+							e.listeners[event.Name][currentIndex] = e.listeners[event.Name][lastIndex]
+
+							// remove the (now cloned) lastIndex listener from slice
+							e.listeners[event.Name] = e.listeners[event.Name][:lastIndex]
```
leaanthony, line 172: so the problem I found with this approach is that you are iterating on `e.listeners[event.Name]` at line 145 which will cause issues if you alter the struct. This may work if iterating on a copy of the struct ?

leaanthony: This is pretty much the code I had in today but found race conditions if you smashed the `On` methods

GeekStocks: >...which will cause issues if you alter the struct. This may work if iterating on a copy of the struct ?

So @leaanthony you said `struct`, I think you meant `slice` (`e.listeners[event.Name]`)?  I'm not trying to be pedantic, but concurrency is a devil and I just want to be clear about what changes you are concerned about. I am **not** concerned about changes to other fields in the `struct`, but yes I am **very** concerned about changes in the `slice`. _Assuming you meant `slice`..._

I mentally see that race condition now; we've made an error here. When you are smashing `.On()` you would be increasing `len(slice)` in a Goroutine possibly at the same time when `.Start()` has already computed `lastIndex` for the clone and truncate operation in another Goroutine. When that happens then clone and truncate will not work properly. With a `lastIndex` value that is _no longer_ the last index, it will truncate the intended target **and** the new one just added to the end of the `slice`!

So the answer is to ensure that **all** code that changes the `slice` is wrapped in a `mutex`. I have code reviewed the file again, and other than `.Start()` where this PR originally made changes, the **only** other location where the `slice` is altered is `.addEventListener()`. All the other methods delegate the task of adding a new listener to the `slice` there. 

_I have just now added a new commit to this PR that does this, and with all locations now wrapped in a `mutex`, your race conditions go away_ 💯

GeekStocks: I've created a demo of the race condition [in the playground](https://play.golang.org/p/URLCzsi7gcp)

leaanthony: Yeah i meant slice. I also wrapped it all in a mutex but that didn't stop the race condition. I'm not sure what the range semantics are and whether it's safe to remove items from a slice you are iterating over. Judging by online comments, it isn't. My first approach to this was to add all valid listeners to a new slice and reassing that but that's heavy. I think the best bet is to make a copy of the listener slice, iterate over that but update the original slice.

leaanthony: In this scenraio, the copy approach would be best. https://freshman.tech/snippets/go/copy-slices/

leaanthony: Where did we get to with this?

leaanthony:
No

leaanthony:
Closing due to lack of response.

Issue #608, opened 2021-02-15
App window left unfocused selecting file


rh12503:
**Description**
On my Mac, after selecting a file through `runtime.Dialog.SelectFile()`, the main app window is left out of focus. 

**System Details**
macOS 10.15.5

Could this issue be resolved if there was a function that called `SetFocus(w->priv.hwnd);`, similarly to [https://github.com/wailsapp/wails/blob/master/lib/renderer/webview/webview.h#L1493](url)?

leaanthony:
Hey! Yes that's quite annoying. Looks like you might have found the answer. Happy to accept a PR for this and I'm sure would be appreciated by many others too.

RH12503:
Thanks for your quick reply! I'm not very familiar with developing frameworks of this type, but I'll try and see if I can get it working. I'll open a PR if I manage to.

RH12503:
Got it working on my Mac by adding 
```
defer w.window.Dispatch(func() {
	w.window.Focus()
})
```
to `SelectFile()`, where `Focus()` calls a C function containing `[w->priv.window makeKeyWindow];`. Since you made the library, could there be any issues with this code?

Also, although adding support for other operating systems would be pretty simple, I'm unable to test Windows and Linux. Is there anyone who could test it out, or should I make a PR of what should work?

leaanthony:
I thought I'd replied to this already but apparently not! If you make a PR for all the systems, I can test them 👍

RH12503:
Seems like there could be a bit of misunderstanding on my part, but the PR I opened (#609) already covers Linux/Windows, and just needs to be tested.

leaanthony:
Ah, my bad. Will check it out over the weekend 👍

leaanthony:
Confirmed working on Windows

leaanthony:
Confirmed working in Linux! Nice job!

leaanthony:
Added to latest pre-release. Please let me know how it goes: `wails update -pre`. 👍

RH12503:
Works perfectly, thanks for your help! Closing this issue now. :)

PR #609, opened 2021-02-16
Put main window into focus after a dialog closes

RH12503:
This PR should be a fix to #608, putting the main window into focus after a dialog closes. 

Tested and works on macOS 1.15.5, but it still needs to be tested on Windows and Linux. (Although it should work for both)


leaanthony:
Wow! That was quick! Can test on Linux and Windows.

PR #612, opened 2021-02-25
Support for minimum and maximum window sizes

RH12503:
This PR should add support for minimum and maximum window sizes on macOS, Windows, and Linux. 

Tested and works on my Mac, but it still needs to be tested on Windows and Linux. (Although I'm a bit uncertain about Windows)


lib/renderer/webview/webview.h
```diff
@@ -75,6 +75,11 @@ struct webview_priv
   DWORD saved_style;
   DWORD saved_ex_style;
   RECT saved_rect;
+
+  int min_width
```
leaanthony, line 79: These lines need semicolons

leaanthony:
Hi. Can confirm this doesn't work on Windows. I needed to add semicolons to the var declarations. In debug mode it would run but the size wasn't applying. In prod mode the app would crash.

As a suggestion, perhaps add these to the config too so that they get applied on startup. 

Will try linux next 👍

RH12503:
Sorry about the careless mistakes! I've added some changes - does Windows work now? If Windows or Linux don't work, could you send any error output? Thanks!

RH12503:
I've installed Parallels on my Mac so I don't have to keep bothering you for testing. I'll let you when I get it working so you can see if it works on the real OSs!

leaanthony:
Perfect! Can test to confirm when you're ready.

RH12503:
I've got it working on Mac + Windows and Linux on Parallels, so hopefully it works on all the OSs now! 

I've also added them to the config, but it requires both dimensions to be set to work (eg. MinWidth and MinHeight both need to be set). I'm not sure if you want to add a struct for the size, or if the current design is okay?

config.go
```diff
@@ -115,6 +147,23 @@ func (a *AppConfig) merge(in *AppConfig) error {
 	if in.Height != 0 {
 		a.Height = in.Height
 	}
+	
+	if in.MinWidth != 0 {
```
leaanthony, line 151: If you don't specify a Min/Max will it use -1?

config.go
```diff
@@ -127,6 +176,10 @@ func newConfig(userConfig *AppConfig) (*AppConfig, error) {
 		Width:     800,
 		Height:    600,
 		Resizable: true,
+		MinWidth: -1,
```
leaanthony, line 179: Wondering if this should be 0

lib/renderer/webview/webview.h
```diff
@@ -169,6 +179,7 @@ struct webview_priv
   WEBVIEW_API int webview_inject_css(struct webview *w, const char *css);
   WEBVIEW_API void webview_set_title(struct webview *w, const char *title);
   WEBVIEW_API void webview_focus(struct webview *w);
+  WEBVIEW_API void webview_minsize(struct webview *w, int width, int height);  WEBVIEW_API void webview_maxsize(struct webview *w, int width, int height);
```
leaanthony, line 182: Looks like there's some odd wrapping here. `go fmt` should fix this.

leaanthony:
This looks really good @RH12503! I've added a couple of comments, more clarifications than anything.

leaanthony:
Testing feedback for Windows: Looks like it's mainly working. There's 2 small things:
  * If the initial width and height are greater than maxwidth and maxheight then we should clamp the width and height values
  * Looks like the DPI code makes for funky resizes. If I open a 1024x768 window on hidpi it looks bigger than after a resize clamps it down to max size (also 1024x768).

RH12503:
Thanks for the testing! The reason I used -1 as a default value instead of 0 is because MinWidth and MinHeight could be set to 0. The value -1 indicates that it hasn't been set by the user and to use system defaults. However, I could change it to 0 if it fits the framework better.

RH12503:
I've added some changes based on your suggestions. Sorry for the changes in some of the files - running `go fmt` seemed to modify them. Hopefully it works on Windows now!

leaanthony:
Can confirm this works well on Windows 👍

leaanthony:
Sorry for the delay in testing on Linux. Will get that done and hopefully merged at the weekend 👍

leaanthony:
Finally got round to testing on Linux! For the most part it works well! Good Job! The only thing I was able to manage to do was go fullscreen which I'm not sure I tested on the other platforms. So whilst resizing by dragging the window border certainly works, double clicking the window bar breaks the model. 

Done a lot of reading and can't seem to find a way to prevent it. It doesn't look like electron can handle this either. Thoughts?

RH12503:
I can confirm that maximizing works on Mac as expected, but I'll test it out on Windows too. 

To fix this on Linux, I think two things could work. After detecting when the title bar is double-clicked: 
A) Clamp the window dimensions.
B) Disable the event. (This could cause confusion eg. if they have some other double-click action)

I'll do some research and see if there's a solution though. Thanks for testing! :)

leaanthony:
If a maximum size is set then we should prevent maximizing on Mac right?

RH12503:
Sorry for the lack of clarity, I meant that on Mac the window only grew to its maximum width and height when the title bar was double clicked. 

Unfortunately, I can’t think of any way for the window to grow only to the maximum dimensions when maximized on Linux. 

A simple fix for both Windows and Linux would be to disable maximizing so double clicking on the title bar wouldn’t do anything. (As both APIs have properties for that)

Would that solution be okay?

leaanthony:
It's certainly worth a try!

RH12503:
I've tested it out on Ubuntu 20.04.2.0 and it seems double-clicking the title bar doesn't grow the window past the maximum width/height set. Is something else occurring on your end?

leaanthony:
I tested on PopOS 20.10 and it did grow past the maximum size. I'll double check and grab a video.

RH12503:
I've tested this out on PopOS 20.10 - and it definitely grows beyond the set size. However, even when I disabled all actions on the title bar using GTK, it still grew (while it didn't on Ubuntu). 

This probably means that GTK 3 is lacking support for PopOS, but I'll look into alternative solutions.

leaanthony:
Thanks for testing. There must be other people who have dealt with this.
I'll see if i can find anything out too.

RH12503:
I've added a commit that should fix the issue, but I've never seen anything like it done before so it definitely needs to be tested extensively.

leaanthony:
Wow, that's smart. I'm happy to target window managers vs distros as I think that's where the issue is

leaanthony:
Hi, I've tested this on PopOS and whilst it technically works, it glitches quite badly. If it's any consolation, I took some time to install electron and set up the same scenario with it. Can confirm it does not handle this at all! I think we can forget about this scenario for now. Could you possibly revert the most recent change to try and overcome this issue? Also, why does `webview_setmaxsize` also set minimum size and vice versa?

RH12503:
That's pretty weird how even Electron doesn't have a fix for this! I've reverted the commit, but I'll see if there's another way. 

If I recall correctly, the reason `webview_setmaxsize` sets both the minimum and maximum sizes is because while testing, I found that `gtk_window_set_geometry_hints` overwrites one hint if you don't specify the other for some reason (even with the masks).

leaanthony:
That's weird about the masks 🤨 I'm sure I had them separate for v2 linux but I'll need to check when I thaw the code! Let me know when you're satisfied with this then we'll look to merge it. We'll also need to update the docs!

leaanthony:
It's been a pleasure working on this with you @RH12503! Awesome job!

PR #621, opened 2021-03-19
fix: V2 code gen issue

### LINKED ISSUE #620: V2 code gen issue
Hi im testing v2 with vue3, and the assets.h  generate an invalid code, in my case:
const unsigned char *assets[] = { html0, javascript1, 2, css2, 0x00 }; , i add 2 js files in html head using:
<script type="module" crossorigin src="/assets/index.js"></script>
  <link rel=“modulepreload” href=“/assets/second.js”>

amaury-tobias:
In some bundler enviroments like vite the output on index.html uses link
tag with rel="modulepreload" to load javascript, add support to handle
this files on assetbundle, more info
https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/modulepreload
this is experimental, so maybe its necessary review what types are supposed to be supported

fix #620

### DIFF
--- a/v2/internal/html/assetbundle.go
+++ b/v2/internal/html/assetbundle.go
@@ -112,6 +112,9 @@ func (a *AssetBundle) processHTML(htmldata string) error {
 					if attr.Key == "as" && attr.Val == "script" {
 						asset.Type = AssetTypes.JS
 					}
+					if attr.Key == "rel" && attr.Val == "modulepreload" {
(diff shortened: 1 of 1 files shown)


leaanthony:
This is absolutely incredible! Thank you! Today, I've switched *all* V2 work to the `v2-alpha` branch. Please consider this `main` for v2 work.

PR #630, opened 2021-03-24
Proof of concept: vite js template and frontend.bridgeFile option

homburg:
- Added a working project template for `vite.js`
- Converted `bridge.js` / `init.js` to es modules for `vite.js` (can probably be avoided with vite config)
- Added `bridgeFile` option for where to generate bridge js file


leaanthony:
Hi. Thanks for this! There's a few inter-related things going on so I wanted to work with you on what that means for this PR. We're working through https://github.com/wailsapp/wails/pull/568 to get some portability for the CLI. This affects calculating the path for the bridge file, which will affect this PR. 
I also need to review whether moving the bridge to a different location will affect all the current templates. 
Could you explain again why you need to move this file? Is it because you are running bridge mode, then you install some new dependencies, but want to still keep the bridge running? Whilst it will install the latest version of the runtime, it should be replacing the file when you start the bridge. If it isn't then we should fix that.

homburg:
> Hi. Thanks for this! There's a few inter-related things going on so I wanted to work with you on what that means for this PR. We're working through #568 to get some portability for the CLI. This affects calculating the path for the bridge file, which will affect this PR.
> I also need to review whether moving the bridge to a different location will affect all the current templates.
> Could you explain again why you need to move this file? Is it because you are running bridge mode, then you install some new dependencies, but want to still keep the bridge running? Whilst it will install the latest version of the runtime, it should be replacing the file when you start the bridge. If it isn't then we should fix that.

Hi, you're welcome :-)

I think there a few different arguments to move the bridge file out of `node_modules`.

1. In this particular setup, the issue is that `vite` will create and cache a "pre-bundled" bundle from dependencies. So I always get a bundle from the initial `node_modules/@wailsapp` source. Unless I explicitly delete the `node_modules/.vite` cache folder.

2. In some package manager setups (pnpm, yarn pnp) `node_modules` packages are symlinked from a central location or not placed in `node_modules` at all.

3. When trying most of the current templates, I would get confusing js errors in the browser, eg. `window.wails._` is not defined. Because the bridge file was not updated or it was cached from earlier. If i was including a specific generated file it would be more obvious that the bridge file needed to be generated and not just replaced.

leaanthony:
Thanks for the detailed reply @homburg! The deal with bridge is that it basically "replaces" the API of the runtime. Ideally I'd like to have them as 2 distinct modules that are included depending on an environment variable but I couldn't find an easy way to do that. For v2, the bridge is currently part of the runtime code so there's no need to switch, it just determines what mode it's running and does the right thing. For v1 that's a lot harder.

Let me get the other PR out of the way then we can move forward with this 👍

leaanthony:
@homburg Would it be possible to make the bridge file location a separate PR? I think this will be easier, then we can manage the site template as a separate PR after that?

leaanthony:
@homburg - you know, for v2 I've taken the approach of including the bridge code in the module and it is determined at runtime to select the appropriate initialisation path. I think I'll back port that and see if that satisfies the issue.

leaanthony:
Moving Vite related discussion to here: https://github.com/wailsapp/wails/discussions/717

PR #633, opened 2021-03-25
feat(cli): use gitconfig name as name prompt default value

### LINKED ISSUE #628: CLI: make name prompt default value based on git name
No

In the `wails setup` name prompt, if no existing wails config is found, we can make the default value the `name` field from `~/.gitconfig`. 

**Describe alternatives you've considered**
Alternatively we can use the currently logged in user's name. No additional libraries would be required as this functionality is [part of os/user](https://golang.org/pkg/os/user/#Current). I have an implementation of this [here](https://github.com/hi019/wails/commit/0a434e140aeecc9dfa09ccace2a6226f3e180e44) but can also make a PR for the other option above.

**Additional context**

hi019:
Fixes #628

### DIFF
--- a/cmd/helpers.go
+++ b/cmd/helpers.go
@@ -595,3 +595,9 @@ func ldFlags(po *ProjectOptions, buildMode string) string {
 	}
 	return ldflags
 }
+
+func getGitConfigValue(key string) (string, error) {
+	output, err := exec.Command("git", "config", "--get", "--null", key).Output()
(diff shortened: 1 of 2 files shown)


leaanthony:
I like this approach! Perhaps we should do the same for email?

hi019:
Good idea, will update this PR soon

leaanthony:
This is great!

leaanthony (approved):

PR #635, opened 2021-03-26
fix(cli): use npm run serve as serve command in all templates

### LINKED ISSUE #587: Align npm serve commands
The templates have different script names for starting the frontend in serve mode.

hi019:
Fixes #587 

Still need to test but don't see why this wouldn't work


leaanthony:
Looks good! Thanks!

PR #647, opened 2021-04-02
Fixes TypeScript generated typings

Igogrek:
Currently generated typings produce invalid TypeScript code and look  something like this:
```
Interface ServerApi {
	AddNewServer: (...args : any[]) => Promise
	DeleteServer: (...args : any[]) => Promise
	GetExistingServers: (...args : any[]) => Promise
}

Interface Backend {
	ServerApi: ServerApi
}

declare global {
	interface Window {
		backend: Backend;
	}
}
```
First of all, some `export` is required, since augmentations of window aren't allowed otherwise (TS error). 
`interface` keyword should start with a lowercase and `Promise` should have a typed response, even if it's an `any`.

After this small change typings look like this:
```
interface ServerApi {
	AddNewServer(...args : any[]):Promise<any>
	DeleteServer(...args : any[]):Promise<any>
	GetExistingServers(...args : any[]):Promise<any>
}

interface Backend {
	ServerApi: ServerApi
}

declare global {
	interface Window {
		backend: Backend;
	}
}
export {};
```

TypeScript validity can be checked here https://www.typescriptlang.org/play for example

### DIFF
--- a/lib/binding/manager.go
+++ b/lib/binding/manager.go
@@ -115,15 +115,15 @@ func (b *Manager) generateTypescriptDefinitions() error {
 
 	for structname, methodList := range b.structList {
 		structname = strings.SplitN(structname, ".", 2)[1]
-		output.WriteString(fmt.Sprintf("Interface %s {\n", structname))
+		output.WriteString(fmt.Sprintf("interface %s {\n", structname))
 		for _, method := range methodList {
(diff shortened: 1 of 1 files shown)


leaanthony:
Thank you for this! TS support has always been experimental so it's great to firm this up a bit.

PR #655, opened 2021-04-06
Handle use cases where the function is not from main

sithembiso:
Without this fix, Wails doesn't work if you try to bind a function from a package other than `main`.


leaanthony:
Hi @sithembiso. It's taken me a while to get to this!

Issue #658, opened 2021-04-08
Wails apps not displaying gui


chrisfair:
**Description**

As of yesterday morning I am no longer able to see any wails guis this is the case with my own app and also the Wally app for my keyboard

**To Reproduce**
Attempt to run the application you will be presented with a black screen 

If run from command line I get this message
Overriding existing handler for signal 10. Set JSC_SIGNAL_FOR_GC if you want WebKit to use a different signal
Control

**Expected behaviour**
I expect to see the gui

**Screenshots**
This is a screen shot of wally running which ran just fine the other day
![wally](https://user-images.githubusercontent.com/4054917/114038683-4f8e0080-983f-11eb-95eb-6f53b77d9b94.png)

If running from cli you will see this message

Overriding existing handler for signal 10. Set JSC_SIGNAL_FOR_GC if you want WebKit to use a different signal
Control

**System Details**

| Name   | Value |
| ----- | ----- |
| Wails Version | v1.16.0 |
| Go Version    | go1.16.3 |
| Platform      | linux |
| Arch          | amd64 |
| GO111MODULE   | (Set) |
| GCC           | 10.2.0 |
| Npm           | 7.8.0 |
| Node          | v15.14.0 |
| Distro         |Manjaro 21.0.1 |
|Kernel         |5.10.28|
|Shell            |Fish 3.2.0-60|
|Windows Manager | I3 |
|Processor    |Ryzen 7 x1700|
|Memory    |DDR 4 32GB 2400MHz|

**Additional context**
Pretty straight forward "It don't work" it is possible that it is a problem with webkit.

leaanthony:
Wow, that's super weird. Are you saying that new builds don't work or any apps don't work?

fdidron:
Hi @chrisfair,

Can you give me the result of the following command: pacman -Qs webkit2gtk

Also, while I'm looking into this, there's a cli version of wally available here if that helps: https://github.com/zsa/wally-cli/releases

fdidron:
@leaanthony FYI I got similar reports from users on Arch / Ubuntu with the webkit2gtk version 2.32.0-1 .

leaanthony:
It sounds like WebKit has been removed/upgraded and now the app can't find it? I wonder if there's something we can do in terms of warning about that. This is the first time I've heard of this.

leaanthony:
@fdidron thanks for your help in diagnosing this 👍

fdidron:
[...]

PR #659, opened 2021-04-09
fix: #658 polyfill window.external for latest webkit2gtk versions

fdidron:
With webkit2gtk >= 2.32, the window.external object is not passed to the webview causing the wails js runtime to crash.

To avoid a possible breaking change by modyfing the webview rendering code, I updated the js runtime to polyfill `window.external` if it's undefined.


leaanthony:
This is amazing! Awesome job finding that and it's a perfect fix. I'll tag a new release shortly 👍

leiless:
Hi, @fdidron, by saying:

> With webkit2gtk >= 2.32, the window.external object is not passed to the webview

I wonder how do you confirm this, I'm not using the wails app, but I came across a similar symptom for Eclipse SWT.

https://bugs.eclipse.org/bugs/show_bug.cgi?id=573616

PR #667, opened 2021-04-11
Support for default colored title bar

RH12503:
Right now, if a color isn't specified in the config, a white title bar is used. This PR makes it so if no color is specified the color of the title bar isn't changed. 

Tested and works on macOS, and setting the color of the title bar doesn't seem to work on Linux or Windows anyway. 

I'm not sure if there's a reason for having white as the default color, but if this change doesn't work feel free to close this.


leaanthony:
I like it!

PR #669, opened 2021-04-11
Change window.external handling to fix broken Windows builds

### LINKED ISSUE #668: Windows builds fail to run after recent window.external change
**Description**

With the recent change to fix webkit2gtk, the `window.external` variable gets assigned to every time the app starts up. When this happens in IE, it throws an error that "assignment to read-only properties is not allowed in strict mode".

**To Reproduce**
1. Run `wails build` on Windows and open the app.

**Expected behaviour**
The UI will load.

**Screenshots**
image

**System Details**

| Name   | Value |
| ----- | ----- |
| Wails Version | v1.16.2 |
| Go Version    | go1.16.2 |
| Platform      | windows |
| Arch          | amd64 |
| GO111MODULE   | (Not Set) |
| GCC           | 9.2.0 |
| Npm           | 6.14.11 |
| Node          | v14.16.0
 |

- [x] This issue is for Windows and I have read the [Windows Developer Guide](https://wails.app/guides/windows/)

aschey:
Resolves #668 

With the recent change to fix webkit2gtk, the `window.external` variable gets assigned to every time the app starts up. When this happens in IE, it throws an error that "assignment to read-only properties is not allowed in strict mode".

This change makes it so the code to assign to window.external is never reached in IE so the error is prevented. I had to make an explicit `if` statement to prevent webpack from optimizing the code and still causing the error to occur.

I tested this on both Linux and Windows.


leaanthony:
Oh, really nice pickup! Thanks for doing this.

leaanthony:
I'll get a new release out shortly.

aschey:
Great, thank you!

PR #689, opened 2021-04-21
Add User Application Cache Directory

etiennemarais:
## Summary

I added a small enhancement to the runtime filesystem to enable getting the user's application cache directory.

## Motivation

`os.UserCacheDir` returns the default root directory to use for user-specific cached data. Users should create their own application-specific subdirectory within this one and use that. The Go library call wraps the correct platforms' cache directory to allow an application to have it's data sandboxed and in a place that makes sense. 

This will allow a Wails application to store a local cache sqlite database file or write files/media and other files in directory that is namespaced by the Application.

### Use Cases

- sqlite database file to power application local cache

```
# MacOS
/Users/<your-user-name>/Library/Application Support/<application-name>/data
```

Screenshot 2021-04-21 at 10 25 20

### Changes

- Allow for the user application cache directory to be available via the runtime filesystem.

### DIFF
--- a/runtime/filesystem.go
+++ b/runtime/filesystem.go
@@ -14,3 +14,8 @@ func NewFileSystem() *FileSystem {
 func (r *FileSystem) HomeDir() (string, error) {
 	return os.UserHomeDir()
 }
+
+// UserCacheDir returns the user's application cache directory
+func (r *FileSystem) UserCacheDir() (string, error) {
+	return os.UserCacheDir()
+}


leaanthony:
Hey, sorry for taking so long to respond. Do you think people would rather do this vs just using the Go method directly? I'm happy for it to go in, but I'm getting more and more feedback that people would like to use vanilla Go where possible.

leaanthony:
Closing due to no response. Please reopen if you still want to merge this.

PR #700, opened 2021-05-01
[v2] Implement server to client notify support.

akhudek:
This PR implements the client to server notify functionality in V2. It consists of an update to the server to send the message, and and update to the javascript runtime. I've done a minor version bump on the runtime library.


akhudek:
Hmm, as far as I can tell this entire file uses spaces. Is the decision to use tabs from after this file was created? Or is it a Go thing?

leaanthony:
Thanks for this. I'm going to disable that checker. Would rather encode via prettier, just need to determine how

PR #703, opened 2021-05-02
feat(cli):Added vue3-js template support

misitebao:
1. Added vue3-js template support
2. Added Chinese readme
3. Optimize the original readme


leaanthony:
Thanks for this and apologies for the delay. How is this different to the current vue3 template?

misitebao:
@leaanthony 
The current Vue3 is a template based on Typescript. This template is based on Javascript, nothing more. Because there are still many people who don't like to use Typescript, they only like pure Javascript. |

leaanthony:
Ah perfect. Sorry I didn't understand. I've been pretty ill lately so things have been moving slowly. Should pick up this week 🤞

misitebao:
@leaanthony 
Take care of your body and health is the most important thing. I hope you get well soon.🍰

PR #708, opened 2021-05-07
hack: enable builds on non-Windows

tmclane:
Howdy!

So I was on a flight and was attempting to get wails to build on my mac.
Ran in to the Windows specific code inside the `base` and just added this small hack to get it to work.

```
 go build
../../internal/ffenestri/windows/x64/x64.go:5:12: pattern webview.dll: no matching files found
```

I do not really want this merged since I think a better solution should be created first.

### DIFF
--- a/v2/internal/ffenestri/windows/x64/x64.go
+++ b/v2/internal/ffenestri/windows/x64/x64.go
@@ -1,3 +1,5 @@
+// +build windows
+
 package x64
 
 import _ "embed"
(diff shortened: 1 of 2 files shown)


leaanthony:
Ah yrs. I may have temporarily borked Mac builds. I thought x64.go was o.my imported by a windows file therefore guarded. Will check. Thanks!

leaanthony:
I couldn't reproduce but happy for this to go in as it's the right thing to do.

PR #709, opened 2021-05-08
[V2] width height fix

akhudek:
Currently if MaxWidth and MaxHeight are left as default, it causes Width and Height to be set to 0. This PR fixes this and adds tests for the width and height behaviour of MergeDefaults.


leaanthony:
Sorry this took so long. Been ill! Thanks!

PR #718, opened 2021-05-18
Bug/shadowed err

tmclane:
small bug I found when building on a mac.

```
 #] go build .
# github.com/wailsapp/wails/v2/pkg/commands/build
../../pkg/commands/build/packager_darwin.go:155:4: err is shadowed during return
```


leaanthony:
I must have been too eager to push 😃 Thanks!

Issue #614, opened 2021-03-01
Crash after save dialog on Mac


rh12503:
**Description**
After opening a save dialog with a filter, not interacting with the dialog, and overwriting a file, the app crashes. 

**To Reproduce**
Using the _.svg_ extension as an example:
1. Save a file _Untitled.svg_ in your default save location. (the location which the save dialog opens to)
2. Open a save dialog using a filter `c.r.Dialog.SelectSaveFile("", "*.svg")`
2. Don't interact with any part of the save dialog 
3. Press the save button, and select replace when prompted 

**System Details**
macOS 10.15.5

**Additional context**
The error
`Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'ROSP: sandbox extension failure: nil result or wrong class type ((null)) for finalURL`
suggests that the default filename is blank, despite appearing as _Untitled.extension_ to the user.

Pretty sure this could be fixed by adding `[panel setNameFieldStringValue:defaultName];`, although it's weird that the dialog recognizes that there is a duplicate file.

leaanthony:
Thanks for reporting this! Did you try the `setNameFieldStringValue` fix? If so, would be happy to merge it!

leaanthony:
@RH12503 -Did you get anywhere with the `setNameFieldStringValue` fix?

leaanthony:
I'm not able to reproduce this in Big Sur

RH12503:
Sorry - completely forgot about this issue! I'll see if `setNameFieldStringValue` fixes this.

leaanthony:
Cool. What version of MacOS are you using? Can't reproduce

RH12503:
I'm on macOS 1.15, and it seems for sandboxed apps before Big Sur, if the default filename is Untitled the save dialog crashes. 

The only way to fix this might be to change the default name to something besides "Untitled", but I'm not sure if you want to do that. 

Related post: [https://stackoverflow.com/q/21039088/15283541](https://stackoverflow.com/q/21039088/15283541)

leaanthony:
I've got a Mojave machine. Will see if I can reproduce on that.

leaanthony:
@RH12503 did `setNameFieldStringValue` fix this for you locally?

RH12503:
So sorry, I've been busy lately and completely forgot about this! Using `setNameFieldStringValue` works, but the default name can't be set to "Untitled" for some reason. 

[...]

PR #741, opened 2021-06-24
Fix crash on Mac save dialog when replacing file

RH12503:
Fix to #614


leaanthony:
Weird, but sure! 😄 Thanks!

PR #762, opened 2021-07-21
[v2] handle inline javascript in assetbundle.go

s12chung:
Think it would be nice to allow inline javacsript.

I've created an app via `create-react-app` and stumbled upon this. Also, I bumped into errors, when none of these conditions were met: https://github.com/wailsapp/wails/blob/v2-alpha/v2/internal/html/assetbundle.go#L97-L118

Think better error reporting might be a good idea so users can remove those `link` tags?

### DIFF
--- a/v2/internal/html/assetbundle.go
+++ b/v2/internal/html/assetbundle.go
@@ -137,7 +137,7 @@ func (a *AssetBundle) processHTML(htmldata string) error {
 						break
 					}
 				}
-				if !paths.Contains(asset.Path) {
+				if !paths.Contains(asset.Path) && asset.Path != "" {
 					err := asset.Load(a.basedirectory)
 					if err != nil {
 						return err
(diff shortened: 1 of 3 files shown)


s12chung:
Thinking about this further, using a directory + `default to "index.html" for HTML file` feels like a better interface to `wails`.

`wails` can package the entire directory. All static HTML generators generate a directory. A directory is what `s3` or github pages takes for static websites. You get the following benefits:

- No need to parse a HTML file
- No need for templates, just ask users to specify the directory they output to

After digging around, I feel like the motivation of your design is keeping everything within a single binary. So you're doing a `asset -> javascript`, `javascript -> C`, `C -> javascript`, `javascript -> asset` conversion, which is harder to work with.

Maybe something like this is better? `assets folder -> gzip`, `gzip -> cache dir`. And webview can read the directory. I'm willing to help out wit this direction.

leaanthony:
Nice!

Issue #748, opened 2021-06-30
Added Chinese document support


misitebao:
This issue is used for submissions related to Chinese documents

leaanthony:
I will get onto this, I promise! :-)

PR #765, opened 2021-07-26
Synchronize chinese readme

misitebao:
- format file
- synchronize Chinese readme
- use a more standard language identifier

(#748)


leaanthony:
Nice work as always @misitebao !

Issue #786, opened 2021-08-28
Support Distribution 'NixOS'


diogox:
**Description**
Distribution 'nixos' is currently unsupported.

**Further Information**

| Name   | Value |
| ----- | ----- |
| Wails Version | v1.16.6 |
| Go Version    | go1.16.7 |
| Platform      | linux |
| Arch          | amd64 |
| GO111MODULE   | (Not Set) |
| Distribution ID   | nixos |
| Distribution Name   | NixOS |
| Distribution Version   | 21.05.20210820.f5c2a9a |

leaanthony:
Hey there! I don't have nixos, but supporting it should be pretty easy. Please follow [this guide](https://wails.app/guides/distro/) and feel free to raise a PR

leaanthony:
I should also mention that in v2, we use automatic discovery to determine the dependencies. The nix package manager isn't currently on that list so your PR would help me add it. Thanks!

diogox:
I have made an PR, please lmk what you think :)

NixOS is a somewhat different linux distro, so this PR will allow wails to support NixOS, but on NixOS's side I still need to package wails so that it'll work. I'll try to do that next, but that shouldn't impact this repo whatsoever :+1:

leaanthony:
I've merged the PR and this is released as part of v1.16.7. Please close if you think it's complete.

diogox:
I've managed to get this working on NixOS. I'm not 100% with the package I made for it so I'll keep trying to improve it, but this is all set on this end :+1:

PR #788, opened 2021-08-29
Add NixOS support

diogox:
Related to #786.

By itself, this won't work it NixOS. I'll try to package wails for NixOS so that it will. This PR will, however, allow wails to work in NixOS.


leaanthony:
Hey there! Could you posibly rebase this on `develop` please? We use that branch until a stable release is made. Thanks!

PR #833, opened 2021-09-29
Added JS case to mimecache

codydbentley:
I was trying to setup a Vite+Vue3-TS application so I could create a template and ran into an issue with the asset server serving JS files as 'text/plain', which results in the following console error: 

`Failed to load module script: Expected a JavaScript module script but the server responded with a MIME type of "text/plain". Strict MIME type checking is enforced for module scripts per HTML spec.`

### DIFF
--- a/v2/internal/frontend/assetserver/mimecache.go
+++ b/v2/internal/frontend/assetserver/mimecache.go
@@ -5,8 +5,9 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
+
+	"github.com/gabriel-vasile/mimetype"
 )
-import "github.com/gabriel-vasile/mimetype"
 
 var (
 	cache = map[string]string{}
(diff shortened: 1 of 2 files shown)


leaanthony:
Thanks - This is great! I'm starting to wonder if there's a better mimetype library if we're patching css AND js 😂

codydbentley:
@leaanthony even better, a friend pointed me to [this RFC](https://datatracker.ietf.org/doc/html/rfc4329#section-7.1) to callout `text/javascript` as obsolete, so I made the change. Then today I run across this what appears to be a [newer spec](https://datatracker.ietf.org/doc/draft-ietf-dispatch-javascript-mjs/) saying `application/javascript` is now obsolete, go back to using `text/javascript` 🤦‍♂️

On mime type libs, I can see the issue though. I can't think of a any reliable way to accurately detect JS or CSS, and in those cases, I think it would be fine to just settle for file extension based mime.

leaanthony:
I'll revert the change

PR #840, opened 2021-09-30
Fixed missing comma that breaks bindings with more than one package

codydbentley:
Came across this issue when testing multiple packages and it seemed simple enough to fix


leaanthony:
Nice one!

PR #849, opened 2021-10-05
add template.JSEscapeString to ensure ExecJs normal execution

marktohark:
when I pass a string that have \n \t... 
it will fail to Execjs.
I add JSEscapeString to ensure any special char can be converted.


marktohark:
I thought my discovery question too small.

PR #859, opened 2021-10-07
PutAreBrowserAcceleratorKeysEnabled => false

marktohark:
in order to solve below ↓
image

and also you need to merge PR first from [https://github.com/leaanthony/go-webview2/pulls](https://github.com/leaanthony/go-webview2/pulls)

### DIFF
--- a/v2/internal/frontend/desktop/windows/frontend.go
+++ b/v2/internal/frontend/desktop/windows/frontend.go
@@ -279,6 +279,7 @@ func (f *Frontend) setupChromium() {
 	//acceleratorsWebviewShouldProcess := slicer.Int([]int{ctrlV, ctrlC, ctrlX, ctrlZ, ctrlA, arrowLeft, arrowRight, arrowUp, arrowDown, keyDel})
 
 	chromium.AcceleratorKeyCallback = func(vkey uint) bool {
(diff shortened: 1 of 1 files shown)


leaanthony:
This is fantastic! I didn't realise that setting existed! Some questions:

- Does the callback still get fired in Go? 
- Does this fix your onKeyDown issue?
- I think we still need to post the message in w32 because menus need them. Did you try with a menu?
- If we return true instead of false, does that affect the onKeyDown event in JS?

Thanks for this awesome PR!

leaanthony:
Also, just FYI, we keep track of the minimum Runtime version here: https://github.com/wailsapp/wails/blob/master/v2/internal/ffenestri/windows/wv2runtime/wv2runtime.go#L8 
If we use anything that is released in a later version we just need to be sure we update this to ensure the user has the capabilities. It looks like ICoreWebView2Settings3 is in 1.0.864.35 so we should be ok. https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2settings3?view=webview2-1.0.992.28

marktohark:
thank your mention. that is very helpful.
I just try with menu. it's work.
image
menu callback is targeted. at the sametime, js keydown event will be targeted.
if don't wanna target js event, you may do something in AcceleratorKeyCallback.

PR #865, opened 2021-10-11
add \r\n for awaitIPC callback

marktohark:
if js file last line is comment, it will get error.

like this ↓
//something comment
↓
//something comment});

in order to fix above bug.
add \r\n for awaitIPC callback


leaanthony:
Really nice catch. This code will be used by Mac and Linux so perhaps it would be best to use backtick strings with a return at the start? Could you test that please? Thanks!

PR #906, opened 2021-10-31
use environment variables instead of flags in dev to support CLI apps

phoenix147:
use environment variables to supply the development binary with flags to support CLI arguments in the app


leaanthony:
Wow that was fast :-) What testing did you do?

phoenix147:
I've tested the assetdir and devserverurl via commandline and from the wails.json, and both worked as expected.
I also tested the loglevel flag via command line.

The underlying app with urfave/cli implemented started as expected, but at the moment there is no option to supply it with arguments while using wails dev. Should i add this to this PR or raise another as soon as is ready?

However, i don't find it intuitive, that the assetdir and devserverurl change the wails.json, they should just override it. Maybe we can provide a "-save" flag instead. But this is part of another discussion.

PR #970, opened 2021-11-24
[v2] Use invoke to dispatch callbacks on windows

### LINKED ISSUE #969: [v2] Windows: Backend callbacks get lost when moving window
**Description**
Callbacks get not triggered on MainThread, when the windows is getting moved or when one clicks on the title bar and keeps pressing the mouse button.

**To Reproduce**
1. Create a new v2 template and adjust main.js with the following code
```js
let nameElement = document.getElementById("name");
nameElement.focus();

let counter = 0;
window.greet = function () {
  let name = nameElement.value;
  counter++;
  window.go.main.App.Greet(name + " " + counter).then((result) => {
    document.getElementById("result").innerText = result;
    setTimeout(greet, 1000);
  });
};

nameElement.onkeydown = function (e) {
  console.log(e)
  if (e.keyCode == 13) {
    window.greet()
  }
}
```
2. Compile for windows and run the app on windows.
3. Click on the greet button and see that the counter gets increased every second.
4. Click on the title bar of the window and keep pressing, the counter suddenly stops and the callback is not triggered.

**Expected behaviour**
Callbacks should be called regardless of clicking on the title bar or moving the window.

**System Details**
Wails Version: v2.0.0-beta.21
MacOS Version: 12.0.1
Go Version: go1.17.2

**Additional context**
The problem seems to be caused by using PostThreadMessage to change into the main thread for executing the callback here: https://github.com/wailsapp/wails/blob/9735bd1b0192cbbc0ce48808d422b7ac726a9211/v2/internal/frontend/desktop/windows/window.go#L116

According to this [Thread messages are eaten by modal loops](https://devblogs.microsoft.com/oldnewthing/20050426-18/?p=35783) a nested modal loop might just catch the message but it will never reach the root message loop here:

https://github.com/wailsapp/wails/blob/9735bd1b0192cbbc0ce48808d422b7ac726a9211/v2/internal/frontend/desktop/windows/window.go#L88-L100

Another devblog post from MS [Why are my posted messages getting lost when the user drags my window around?
](https://devblogs.microsoft.com/oldnewthing/20130905-00/?p=3313) suggest to use PostMessage instead of PostThreadMessage and to use a WNDPROC message.

I've added a new `Invoke` method to winc in order to address this issue, I'll provide a PR on winc in order to discuss it. As soon as this `Invoke` method exists, we could use it in wails to change to the GUI thread.

stffabi:
Fix #969 depends on leaanthony/winc#3

### DIFF
--- a/v2/go.mod
+++ b/v2/go.mod
@@ -28,7 +28,7 @@ require (
 	github.com/leaanthony/slicer v1.5.0
 	github.com/leaanthony/typescriptify-golang-structs v0.1.7
 	github.com/leaanthony/webview2runtime v1.1.0
-	github.com/leaanthony/winc v0.0.0-20210921073452-54963136bf18
+	github.com/leaanthony/winc v0.0.0-20211124105230-0330cfc6d50c
 	github.com/leaanthony/winicon v1.0.0
 	github.com/matryer/is v1.4.0
 	github.com/olekukonko/tablewriter v0.0.4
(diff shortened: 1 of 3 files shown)


v2/go.mod
```diff
@@ -84,3 +84,5 @@ require (
 	golang.org/x/xerrors v0.0.0-20200804184101-5ec99f83aff1 // indirect
 	gopkg.in/warnings.v0 v0.1.2 // indirect
 )
+
+replace github.com/leaanthony/winc v0.0.0-20210921073452-54963136bf18 => github.com/stffabi/winc v0.0.0-20211124100926-dc9a1c2bfe1c
```
leaanthony, line 88: :-)

leaanthony:
I've merged in the PR on winc. Happy to move on this once the `replace` is removed. Is there anything further you'd like to discuss on this? Hugely appreciative of this PR! 🙏

leaanthony (changes requested):
I think we're good to go once the replace has gone 🚀

stffabi:
Thanks for merging the winc PR. You're very welcome for this PR, it took me quite some time to get this one fixed. But it was an interesting experience in winapi 😄

PR #973, opened 2021-11-24
docs: fix image path errors and optimize assets paths

misitebao:
Fixed the problem of the wrong path of the sponsored list picture. At the same time, moved the picture resources to the `assets` folder to optimize the current situation of the root directory file confusion.


leaanthony:
Nice one @misitebao !!

PR #982, opened 2021-11-29
[v2] windows: Do not apply WindowStartState on reloads

### LINKED ISSUE #981: [v2] Windows: A reload restores the WindowStartState
**Description**
We use reloads `runtime.WindowReload()` during the lifetime of our frontend and a reload triggers a restore of the WindowStartState on Windows.

**To Reproduce**
1. Create a new v2 template with a WindowStartState of Maximised and a button which triggers a reload.
2. Compile it for Windows and let it run.
3. The app is started maximised, resize the window a smaller size.
4. Click the reload button
5. The WindowState has been reset, and is now again maximised.

**Expected behaviour**
WindowStartState should only be applied for the very first load of the frontend.

**System Details**
Wails Version: v2.0.0-beta.20
MacOS Version: 12.0.1
Go Version: go1.17.2

stffabi:
Fix #981

### DIFF
--- a/v2/internal/frontend/desktop/windows/frontend.go
+++ b/v2/internal/frontend/desktop/windows/frontend.go
@@ -43,6 +43,8 @@ type Frontend struct {
 	bindings                                 *binding.Bindings
 	dispatcher                               frontend.Dispatcher
 	servingFromDisk                          bool
+
+	hasStarted bool
 }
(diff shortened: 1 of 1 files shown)


leaanthony:
Nice pick up! Thanks

PR #989, opened 2021-12-01
[v2] Do not block during processing of messages

### LINKED ISSUE #986: [v2] Windows: A long running backend call blocks the main thread
**Description**
Having a backend call in go which takes a long time, e.g. executing an external program, leads to a completely blocked main thread and therefore the WebView is also blocked.

**To Reproduce**
1. Create a new v2 template
2. Add a sleep to the Greet method.
3. Click on the button.
4. Window is completely blocked.

**Expected behaviour**
The window should not block and the WebView should still be responsive, e.g. for scrolling etc.

**System Details**
Wails Version: v2.0.0-beta.21
MacOS Version: 12.0.1
Go Version: go1.17.2

**Additional context**
I also checked how Darwin behaves in this case. It's slightly different, at least the Window is not blocked because `ProcessingMessages` is decoupled and send to another Go-Routine through a channel. So the window is responsive, but other backend calls are also get blocked until the first is done.

@leaanthony what's your opinion on this? We should at least fix the blocking of the main thread on windows, but what do you think about concurrent backend calls?
I advocate for not blocking backend calls if another backend call is still running, at least up to some degree, e.g. allow up to 10 concurrent calls?

stffabi:
Fix #986 depends on leaanthony/winc#4

### DIFF
--- a/v2/go.mod
+++ b/v2/go.mod
@@ -40,7 +40,6 @@ require (
 	github.com/tdewolff/test v1.0.6 // indirect
 	github.com/tidwall/sjson v1.1.7
 	github.com/wzshiming/ctc v1.2.3
-	github.com/xyproto/xpm v1.2.1
 	github.com/ztrue/tracerr v0.3.0
 	golang.org/x/mod v0.4.1
 	golang.org/x/net v0.0.0-20210510120150-4163338589ed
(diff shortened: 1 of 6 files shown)


v2/go.mod
```diff
@@ -82,3 +81,5 @@ require (
 	golang.org/x/xerrors v0.0.0-20200804184101-5ec99f83aff1 // indirect
 	gopkg.in/warnings.v0 v0.1.2 // indirect
 )
+
+replace github.com/leaanthony/winc v0.0.0-20211124105230-0330cfc6d50c => github.com/stffabi/winc v0.0.0-20211201085336-530acbf04fef
```
stffabi, line 85: Will be removed as soon as leaanthony/winc#4 has been merged.

PR #991, opened 2021-12-01
[windows] Support drawing a frame for frameless to support hiding the titlebar

### LINKED ISSUE #990: [v2] Windows: Support a window without titlebar but with a border
It would be awesome to have support for a window just without a titlebar but with a border. Sometimes one want's just to have a hidden titlebar but still have a border around the window. This is especially interesting if the background of the frontend is mostly white.

A flag to enable drawing a border for windows if frameless is active or some other kind of configuration options to achieve this.

stffabi:
Idea to fix #990

### DIFF
--- a/v2/internal/frontend/desktop/windows/window.go
+++ b/v2/internal/frontend/desktop/windows/window.go
@@ -34,6 +34,9 @@ func NewWindow(parent winc.Controller, appoptions *options.App) *Window {
 	var dwStyle = w32.WS_OVERLAPPEDWINDOW
 	if appoptions.Frameless {
 		dwStyle = w32.WS_POPUP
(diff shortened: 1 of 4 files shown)


leaanthony:
This is great. We can migrate to top level config if the other platforms support it. do you think you could include a doc update in this pr? I want to try and keep the site up to date with PRs. If not I can do it later. Many thanks 🙏

stffabi:
Yes I can adjust the docs, I've just seen that they are on the top level of the repo. I was searching in the v2 folder before... I'll update it tomorrow, because it's getting quite late here 😄

stffabi:
I've added the docs.

stffabi:
Thanks for merging.

PR #1002, opened 2021-12-03
Fix error: cannot use uri (type *_Ctype_char) as type string in argument to common.TranslateUriToFile.

ianmjones:
This happens when building for linux. Full error before fix:

../wails/v2/internal/frontend/desktop/linux/frontend.go:331:47: cannot use uri (type *_Ctype_char) as type string in argument to common.TranslateUriToFile

gofmt reformatted a couple of comments in the fixed file too.


stffabi (approved):
Awesome, thanks so much for this fix.

leaanthony:
Really great to see you here Ian 🙂 Thanks for this 🙏

PR #1004, opened 2021-12-03
[v2] Improve processRequest: Handle errors and behave more like a webserver

### LINKED ISSUE #1003: [v2] linux: requests keep in pending state if loading the asset fails
**Description**
On darwin a request keeps in pending state if loading of the asset fails, e.g. if the file is not found or another error occurs during loading of the asset.

**To Reproduce**
1. Create a new v2 template
2. Import a script into index.html which doesn't exist.
3. Start in dev mode and open the dev-tools on darwin.
4. The request keeps in pending state.

**Expected behaviour**
The request should be finished with a 404 status code.

**System Details**
Wails Version: v2.0.0-beta.21
MacOS Version: 12.0.1
Go Version: go1.17.2

**Additional context**
Returning a 404 or a 500 status code for other errors would also suit to #996, where some improvements were added that `processRequest` behaves more like as static webserver.

stffabi:
This also fixes that requests remain in "pending" state on
darwin if e.g. a file is not found or an error occurs during
loading of the file.

Fix #1003

### DIFF
--- a/v2/internal/frontend/desktop/darwin/Application.h
+++ b/v2/internal/frontend/desktop/darwin/Application.h
@@ -41,7 +41,7 @@ void Quit(void*);
 const char* GetSize(void *ctx);
 const char* GetPos(void *ctx);
 
-void ProcessURLResponse(void *inctx, const char *url, const char *contentType, void* data, int datalength);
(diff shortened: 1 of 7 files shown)


leaanthony:
Fantastic! Thanks for picking this up. I'll do this Linux part as it's in progress 👍

PR #1007, opened 2021-12-03
[v2] feature / on-before-close resolve #978 

### LINKED ISSUE #978: v2) support window close event control
### Discussed in https://github.com/wailsapp/wails/discussions/977

<div type='discussions-op-text'>

<sup>Originally posted by **Ironpark** November 26, 2021</sup>
# Proposal
Name: OnCloseBefore or OnBeforeClose

Type: func(ctx context.Context) (prevent bool)

This callback is Called when the window close button is pressed, just before the application close. It provides the application context and if the return value is true , ignores the close event and does not close the application.

## Why
If we can control the close event fires or not
"There is work in progress. Are you sure you want to quit?" or "Are you sure you want to exit?" UX scenarios that prevent user mistakes, such as popping up a dialog like this, become possible.

## Example

```go
 func OnCloseBefore(ctx context.Context) (prevent bool){
    selection, _ := runtime.MessageDialog(b.ctx, runtime.MessageDialogOptions{
        Type:              runtime.WarningDialog,
        Title:               "There is work in progress!",
        Message:       "Are you sure you want to exit?",
        Buttons:       []string{"Yes", "No"},
        DefaultButton: "Yes",
        CancelButton:  "No",
    })
    return selection == "No"
 }
```</div>

Ironpark:


leaanthony:
Thanks for doing this! Have you tested the permutations? (Lin/Win/Mac)(Set/Unset)?

Ironpark:
> Thanks for doing this! Have you tested the permutations? (Lin/Win/Mac)(Set/Unset)?

Yes, but currently only tested on mac os.

Incidentally, also fixed hideOnClose not working properly on Mac.

leaanthony:
Would it be possible to update the website with details for the new hook please? Thanks 🙏

leaanthony:
I've done it already - all good :-)

Issue #950, opened 2021-11-18
Possible race condition in the EventManager?


meatherly:
**Description**

My app has been crashing on start up with fatal error: concurrent map read and map write. I use the Store to add a lot of items to it to monitor in the frontend of the application. When reviewing the stack trace from the go lang race tool it points back to the event manager.

**To Reproduce**
Steps to reproduce the behaviour:
1. Add a 50 or more new Store items (`runtime.Store.New`) during the `WailsInit` callback.
2. After adding them start changing their values which would cause it to broadcast a value with the Event Manager.
3. You should get a `fatal error: concurrent map read and map` in the logs.  

**Expected behaviour**
The event manager mutex should be locked when reading and writing to the event listener map.

**System Details**

| Name   | Value |
| ----- | ----- |
| Wails Version | v1.16.7 |
| Go Version    | go1.17.2 |
| Platform      | darwin |
| Arch          | amd64 |
| GO111MODULE   | on |
| GCC           | 13.0.0 |
| Npm           | 6.14.11 |
| Node          | v14.16.0 |

**Additional context**
It looks like this problem was going to be solved with #600 and the lines I highlighted might be all that is needed to solve it: https://github.com/wailsapp/wails/pull/604/files#diff-e56817b1f9a346ddb17641708f825441e9618d677a11f092a2a89fa9ca0dc9beR52-R68

here's some of the logs I got from using golang's race tool:
```
==================
WARNING: DATA RACE
Write at 0x00c00025a2a0 by main goroutine:
  runtime.mapassign_faststr()
      /Users/michael/.asdf/installs/golang/1.17.2/go/src/runtime/map_faststr.go:202 +0x0
  github.com/wailsapp/wails/lib/event.(*Manager).addEventListener()
      /Users/michael/go/pkg/mod/github.com/wailsapp/wails@v1.16.7/lib/event/manager.go:60 +0xf8
  github.com/wailsapp/wails/lib/event.(*Manager).On()
      /Users/michael/go/pkg/mod/github.com/wailsapp/wails@v1.16.7/lib/event/manager.go:79 +0x56
  github.com/wailsapp/wails/runtime.(*Events).On()
      /Users/michael/go/pkg/mod/github.com/wailsapp/wails@v1.16.7/runtime/events.go:19 +0x11c
  github.com/wailsapp/wails/runtime.(*Store).setupListener()
[...]

PR #1012, opened 2021-12-06
Fix for #950

meatherly:
Locking the event manager before mutating it. The mutex was never locked before accessing the listeners which caused a concurrency error. This should be the only thing needed to fix the error.


leaanthony:
Nice! Thank you!

PR #1025, opened 2021-12-15
Added react template

AlienRecall:


leaanthony:
Nice one. Be aware that your production section in the README has commands for Wails v1 :-)

AlienRecall:
> Nice one. Be aware that your production section in the README has commands for Wails v1 :-)

Oh I didn't noticed that, now it should be ok

Issue #1049, opened 2021-12-29
Logic to (dis)allow to resize window differs on Windows vs Linux/Darwin


pierrejoye:
#####################################################
     **V1 users: If you have a technical issue, please do not open a bug this way!** 
                      Please use the `wails issue` command!
          If you do not do this then the issue may be closed automatically.                      

 NOTE: If your bug is related to Windows, make sure you read 
 the [Windows Developer Guide](https://wails.app/guides/windows/)
#####################################################

**Description**

**To Reproduce**

Simple example code:

```
package main

import (
	"embed"
	"log"

	"github.com/wailsapp/wails/v2/pkg/options/mac"

	"github.com/wailsapp/wails/v2"
	"github.com/wailsapp/wails/v2/pkg/logger"
	"github.com/wailsapp/wails/v2/pkg/options"
	"github.com/wailsapp/wails/v2/pkg/options/windows"
)

//go:embed frontend/src
var assets embed.FS

//go:embed build/appicon.png
var icon []byte

func main() {
	// Create an instance of the app structure
	app := NewApp()

	// Create application with options
	err := wails.Run(&options.App{
		Title:     "Your-POS",
		Width:     720,
		Height:    570,
		MinWidth:  720,
		MinHeight: 570,
		//MaxWidth:          1280,
		//MaxHeight:         740,
		DisableResize:     false,
		Fullscreen:        false,
		Frameless:         true,
		StartHidden:       false,
		HideWindowOnClose: false,
		RGBA:              &options.RGBA{R: 33, G: 37, B: 43, A: 255},
		Assets:            assets,
		LogLevel:          logger.DEBUG,
		OnStartup:         app.startup,
		OnDomReady:        app.domReady,
		OnShutdown:        app.shutdown,
		Bind: []interface{}{
			app,
		},
		// Windows platform specific options
		Windows: &windows.Options{
			WebviewIsTransparent: false,
			WindowIsTranslucent:  false,
			DisableWindowIcon:    false,
		},
		Mac: &mac.Options{
			TitleBar:             mac.TitleBarHiddenInset(),
			WebviewIsTransparent: true,
			WindowIsTranslucent:  true,
			About: &mac.AboutInfo{
				Title:   "Vanilla Template",
				Message: "Part of the Wails projects",
				Icon:    icon,
			},
		},
	})

	if err != nil {
		log.Fatal(err)
	}
}
```

By default the window is resizable, if I am not mistaken.

[...]

PR #1053, opened 2021-12-30
Allow window resizing regardless of frameless or other options (#1049)

pierrejoye:
Fix bug #1049 on Windows where the disableResize option depends on the frameless option.


leaanthony:
Awesome. Thanks for this.

PR #1068, opened 2022-01-10
Add templates for using Svelte with Vite

BillBuilt:
Add templates for using Svelte with Vite and optionally TailwindCSS v3


leaanthony:
Thanks!

PR #1069, opened 2022-01-11
[v2] Fix: win window maximise/minimise

### LINKED ISSUE #916: [V2] Windows - runtime.WindowMaximise() cannot take effect
**Description**
runtime.WindowMaximise() cannot take effect

**To Reproduce**
Setting runtime.WindowMaximise() anywhere does not take effect，
Maximize first, then restore to the default width and height

**Expected behaviour**
Window is maximized normally

**Screenshots**
No

**System Details**
Go 1.17.2
wails v2.0.0-beta.15
windows10 64bit 21H1

**Additional context**
No

- :ballot_box_with_check: This issue is for Windows and I have read the [Windows Developer Guide](https://wails.app/guides/windows/)

redraskal:
-  Fixes #916 
- Tested on Windows 11

### Issue:
Using `WindowStartState:  options.Maximised` in app options is 100% functional BUT manually setting the window state like
```
func (b *App) startup(ctx context.Context) {
	runtime.WindowMaximise(ctx)
}
```
does not work.

### Solution:
Found out the calls to Show in [navigationCompleted](https://github.com/wailsapp/wails/blob/0d402492db8dd36d455d745ed5f7a1a980ab7000/v2/internal/frontend/desktop/windows/frontend.go#L480) were negating the maximize request in startup.

I added a navigationCompleted check in WindowMaximise and WindowMinimise for windows devices. If chromium is not ready when called, the WindowStartState will be modified before startup and ensures we do not mistakenly call a window Show an extra time.

### Alternate solution:
Set the StartHidden option to true in app options & use the following func
```
func (b *App) startup(ctx context.Context) {
	runtime.WindowShow(ctx)
	runtime.WindowMaximise(ctx)
}
```

### DIFF
--- a/v2/internal/frontend/desktop/windows/frontend.go
+++ b/v2/internal/frontend/desktop/windows/frontend.go
@@ -199,7 +199,11 @@ func (f *Frontend) WindowHide() {
 }
 func (f *Frontend) WindowMaximise() {
 	runtime.LockOSThread()
-	f.mainWindow.Maximise()
+	if f.hasStarted {
+		f.mainWindow.Maximise()
+	} else {
+		f.frontendOptions.WindowStartState = options.Maximised
+	}
 }
 func (f *Frontend) WindowUnmaximise() {
 	runtime.LockOSThread()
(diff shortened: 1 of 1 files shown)


leaanthony:
Thanks for this! I won't be able to look at it today but hopefully can tomorrow.

Issue #1031, opened 2021-12-21
[v2] (wails:dev) smart rebuild


ironpark:
Currently, the wails dev command rebuilds both the frontend and the backend even if the backend code changes that do not affect the frontend.

When the frontend's code base is light, it's not a big problem, but if the frontend build time exceeds a few minutes, it feels like a big problem in productivity.

How about conditionally skipping the frontend build by checking for changed files or checking if the generated "Wails JS" has changed?

leaanthony:
Agreed! Happy to accept a PR for it!

Ironpark:
@leaanthony 

In the last few days, after reviewing #1071 PR and looking at the source code related to the dev command, I found some areas for improvement.

- watch ignore
 It is hardcoded to omit the `node_modules` and `build` directories in the process of setting the directory to watch inside the `initialiseWatcher` function. It seems to be a better design to change this part so that it can be set as a flag.

- add ignore `.git` directory
 It seems like a good idea to ignore the `.git` directory by default, like `node_modules` .

- recv `watcher.Errors` channel stream
 In our current implementation, we are only passing Events from watcher and not reading Errors . If messages accumulate in the Errors channel, watcher 's loop may hang and it may cause a potential bug that causes the build to stop.

 - go file ast check
By using the default `go/parser` package, when the go file is changed, you can parse the file and check whether the build is possible by checking for errors. This prevents build from being triggered too often while editing the source code.

leaanthony:
Thanks for the feedback! I'll answer each point:

> * watch ignore
>   It is hardcoded to omit the `node_modules` and `build` directories in the process of setting the directory to watch inside the `initialiseWatcher` function. It seems to be a better design to change this part so that it can be set as a flag.

Yeah, we *could* add this to the config.

> * add ignore `.git` directory
>   It seems like a good idea to ignore the `.git` directory by default, like `node_modules` .

If we did the above config we could add this by default

> * recv `watcher.Errors` channel stream
>   In our current implementation, we are only passing Events from watcher and not reading Errors . If messages accumulate in the Errors channel, watcher 's loop may hang and it may cause a potential bug that causes the build to stop.

Can you please provide more details on this - thanks!

> * go file ast check
>   By using the default `go/parser` package, when the go file is changed, you can parse the file and check whether the build is possible by checking for errors. This prevents build from being triggered too often while editing the source code.

If you want to code this, I'd be happy to merge it. It's not a small task!

Ironpark:
[...]

PR #1071, opened 2022-01-12
Ignore frontend rebuild if go files updated in dev

leaanthony:
Should not rebuild frontend project if `.go` file was updated in dev mode

### DIFF
--- a/v2/cmd/wails/internal/commands/dev/dev.go
+++ b/v2/cmd/wails/internal/commands/dev/dev.go
@@ -530,6 +530,7 @@ func doWatcherLoop(buildOptions *build.Options, debugBinaryProcess *process.Proc
 	interval := time.Duration(flags.debounceMS) * time.Millisecond
 	timer := time.NewTimer(interval)
 	rebuild := false
+	ignoreFrontend := false
 	reload := false
 	assetDir := ""
 	changedPaths := map[string]struct{}{}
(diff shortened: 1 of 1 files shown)


leaanthony:
@Ironpark - Please test this as a fix for #1031

leaanthony:
@Ironpark any feedback?

Ironpark:
```go 
if _, exists := extensionsThatTriggerARebuild[ext]; exists {
    rebuild = true
    ignoreFrontend = strings.HasSuffix(ext, "go")
    timer.Reset(interval)
    continue
}
```
If you modify the 'go' code before rebuilding after modifying the frontend code, the frontend build doesn't seem to work. 

It is suggested to modify it as follows.
```diff
if _, exists := extensionsThatTriggerARebuild[ext]; exists {
    rebuild = true
+    if(!ignoreFrontend){
        ignoreFrontend = strings.HasSuffix(ext, "go")
+    }
    timer.Reset(interval)
    continue
}
//
rebuild = false
buildOptions.IgnoreFrontend = ignoreFrontend
+ ignoreFrontend = true
```

Ironpark:
Also, since the default flag setting only processes files with the go extension, if you apply the patch, the frontend will not be built at all unless you set a separate flag.

Therefore, it would be better to add `css`, `html`, `js`, and `jsx`, which are often used in front-end development, to the default value of the extensions flag.

Maybe it would be nice to add `ts`, `tsx`, `scss` etc as well.
```diff
func defaultDevFlags() devFlags {
	return devFlags{
		devServerURL:    defaultDevServerURL,
		compilerCommand: "go",
		verbosity:       1,
-		extensions:      "go",
+		extensions:      "go,js,jsx,html,css",
		debounceMS:      100,
	}
}
```

leaanthony:
@Ironpark - I think this is basically redundant now that we are using Vite? Thoughts?

PR #1082, opened 2022-01-16
Updated windows window/frontend to fix issue with html select positioning

MikeSchaap:
Fixes #1081

Need to wait for the following PR (https://github.com/leaanthony/go-webview2/pull/5) to merge so `go-webview2` exposes the `NotifyParentWindowPositionChanged()` method.

Was not sure how to correctly expose the chromium object to the window object. Do you agree with the proposed fix? Doesn't feel really solid imo

Note: First time creating a public Pull Request, so if I can improve certain parts, please let me know :)

PR #1085, opened 2022-01-18
[v2] windows: Improve frameless window handling

### LINKED ISSUE #1084: [V2] No Animation When Minimizing, Maximizing and Restoring and Some Bugs on Frameless Window
**Description**
You know when minimizing or maximizing the animation appear on Windows (Not sure if this works on OSX) as the window moves to the task bar. However if the window set to frameless, then the animation does not happen and the window just disappear in a flash.

**To Reproduce**
1. Set ``Frameless: true`` in ``&options.App``.
2. Use the runtime method ``window.runtime.WindowMinimise()`` in Javascript on a button.

**Expected behaviour**
Animation should starts as normal.

**System Details**
```cmd
System
------
OS:             Windows 10 Enterprise LTSC 2019
Version:        1809 (Build: 17763)
ID:
Go Version:     go1.17.6
Platform:       windows
Architecture:   amd64

WebView2        N/A             Installed       97.0.1072.62
npm             N/A             Installed       6.14.13
*upx            N/A             Available

* - Optional Dependency

Diagnosis
---------
```

stffabi:
The following features are now supported for framless windows
- Allow aero snap
- Minimize and maximize animations
- Allow minimizing/restoring by clicking onto the taskbar icon
- If EnableFramelessBorder is used, the default application border with glass effect is used

Change: Menu is not supported anymore if the window is frameless.

Fix #1084

### DIFF
--- a/v2/internal/frontend/desktop/windows/window.go
+++ b/v2/internal/frontend/desktop/windows/window.go
@@ -3,6 +3,11 @@
 package windows
 
 import (
+	"fmt"
+	"log"
+	"syscall"
+	"unsafe"
+
 	"github.com/leaanthony/winc"
 	"github.com/leaanthony/winc/w32"
 	"github.com/wailsapp/wails/v2/pkg/menu"
@@ -32,15 +37,10 @@ func NewWindow(parent winc.Controller, appoptions *options.App) *Window {
 	}
 
 	var dwStyle = w32.WS_OVERLAPPEDWINDOW
-	if appoptions.Frameless {
-		dwStyle = w32.WS_POPUP
-		if winoptions := appoptions.Windows; winoptions != nil && winoptions.EnableFramelessBorder {
-			dwStyle |= w32.WS_BORDER
-		}
-	}
 
 	winc.RegClassOnlyOnce("wailsWindow")
 	result.SetHandle(winc.CreateWindow("wailsWindow", parent, uint(exStyle), uint(dwStyle)))
+	winc.RegMsgHandler(result)
 	result.SetParent(parent)
 
 	loadIcon := true
@@ -55,16 +55,11 @@ func NewWindow(parent winc.Controller, appoptions *options.App) *Window {
 
 	result.SetSize(appoptions.Width, appoptions.Height)
 	result.SetText(appoptions.Title)
-	if appoptions.Frameless == false {
-		if !appoptions.Fullscreen {
-			result.EnableMaxButton(!appoptions.DisableResize)
-			result.SetMinSize(appoptions.MinWidth, appoptions.MinHeight)
-			result.SetMaxSize(appoptions.MaxWidth, appoptions.MaxHeight)
-		}
-		// Only call EnableSizable for normal windows, frameless windows are always not resizable per default and
-		// the resizing for those will be initiated by the frontend see processMessage.
-		// If EnableSizable is enabled for frameless windows, a small white titlebar will be shown.
-		result.EnableSizable(!appoptions.DisableResize)
+	result.EnableSizable(!appoptions.DisableResize)
+	if !appoptions.Fullscreen {
+		result.EnableMaxButton(!appoptions.DisableResize)
+		result.SetMinSize(appoptions.MinWidth, appoptions.MinHeight)
+		result.SetMaxSize(appoptions.MaxWidth, appoptions.MaxHeight)
 	}
 
 	if appoptions.Windows != nil {
@@ -79,7 +74,6 @@ func NewWindow(parent winc.Controller, appoptions *options.App) *Window {
 
 	// Dlg forces display of focus rectangles, as soon as the user starts to type.
 	w32.SendMessage(result.Handle(), w32.WM_CHANGEUISTATE, w32.UIS_INITIALIZE, 0)
-	winc.RegMsgHandler(result)
 
 	result.SetFont(winc.DefaultFont)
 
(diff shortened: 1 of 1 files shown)


stffabi:
@leaanthony this PR should fix several issues and improve the support for frameless window on windows. Currently I've tested the code on Windows 11 because I don't have a Windows 10 machine, but maybe you and @KiddoV could give it a try on Windows 10.

leaanthony:
Thanks heaps for this! What are the specific scenarios this fixes and does it affect the "emulated" resize done in JS?

stffabi:
It fixes the missing features that a normal window has, like "aero snapping" the animations when minimizing,maximizing the window or allowing to minimize/restore the window by clicking into the taskbar.

I tried to get rid of the "emulated" resize but I didn't succeed, the emulated resize still works but it would need some improvements like mentioned in #1084.
- The border where the resizing gets activated is hardcoded to 16 pixels. This border size should be retrieved from the system.
- Current if the window is maximized the resizing is still active.

leaanthony:
> It fixes the missing features that a normal window has, like "aero snapping" the "animations when minimizing,maximizing the window" or "allowing to minimize/restore the window by clicking into the taskbar".
> 
> I tried to get rid of the "emulated" resize but I didn't succeed, the emulated resize still works but it would need some improvements like mentioned in #1084.
> 
> * The border where the resizing gets activated is hardcoded to 16 pixels. This border size should be retrieved from the system.
> * Currently if the window is maximized `enableResize` is still active.

I can certainly fix those 2 issues if need be. Is there a better way of doing frameless window resize *without* the border? Is this what people generally do? Should I just remove the JS based resize?

v2/internal/frontend/desktop/windows/window.go
```diff
@@ -93,3 +86,76 @@ func NewWindow(parent winc.Controller, appoptions *options.App) *Window {
...
+				}
+
+				return 0
+			}
+		}
+	}
+	return w.Form.WndProc(msg, wparam, lparam)
+}
+
+// TODO this should be put into the winc if we are happy with this solution.
```
leaanthony, line 133: I think it should be in winc too. Why don't we start here and then move it out after?

v2/internal/frontend/desktop/windows/window.go
```diff
@@ -32,15 +36,10 @@ func NewWindow(parent winc.Controller, appoptions *options.App) *Window {
...
-	if appoptions.Frameless {
-		dwStyle = w32.WS_POPUP
-		if winoptions := appoptions.Windows; winoptions != nil && winoptions.EnableFramelessBorder {
-			dwStyle |= w32.WS_BORDER
-		}
-	}
 
 	winc.RegClassOnlyOnce("wailsWindow")
 	result.SetHandle(winc.CreateWindow("wailsWindow", parent, uint(exStyle), uint(dwStyle)))
+	winc.RegMsgHandler(result)
```
leaanthony, line 42: What do you think we should do about this PR that also moves this call? https://github.com/wailsapp/wails/pull/1082/files

stffabi:
> I can certainly fix those 2 issues if need be. Is there a better way of doing frameless window resize _without_ the border? Is this what people generally do? Should I just remove the JS based resize?

I thought to fix those issues in another PR and leave this to just the frameless window improvements, but I don't know if I'll find time to fix them. So if you have any spare time for this, that would be awesome. 

I've really tried to remove the JS based resize and to implement the `WM_NCHITTEST`, but unfortunately as soon as the WebView2 uses the whole client area no `WM_NCHITTEST` message will be send anymore because those will be processed by WebView2.
This seems to be a limitation of WebView2 and there's some ongoing talk about this in the context of WPF on the Feedback repo. So I think for the time being the JS based resizing is the only way to go, if we wanna support resizing on frameless. Maybe there's some sophisticated way with painting everything by our own, but at least for me I would run out of spare time, especially because we currently don't need resize in our frameless app. And it would be also hard to maintain over different windows versions.

The border size could be retrieved from the system with the following API calls:
```go
x := w32.GetSystemMetrics(w32.SM_CXSIZEFRAME)
y := w32.GetSystemMetrics(w32.SM_CYSIZEFRAME)
```

Another thing we won't be able to emulate is the "Win10" feature, where the resizing could be started already outside of the window frame. Maybe this would be possible if we do a lot of drawing of the frame by ourselves, but from my opinion this is out of scope.

v2/internal/frontend/desktop/windows/window.go
stffabi: AFAIK it was only moved to prevent a `nil` pointer dereferencing when `notifyParentWindowPositionChanged` wasn't set. Maybe the best would be to check for `nil` before calling the `notifyParentWindowPositionChanged` or adding `EventMangers` on `ControlBase` in winc for `OnMove` like the other Events, e.g. `OnSize`

v2/internal/frontend/desktop/windows/window.go
stffabi: Yes we could do that, but I think we should at least expose `ScaleWithWindowDPI`.  leaanthony/winc#13 fixes a bug in the scaling and also exposes it.

leaanthony:
> I thought to fix those issues in another PR and leave this to just the frameless window improvements, but I don't know if I'll find time to fix them. So if you have any spare time for this, that would be awesome.

Yeah can certainly try.
 
> I've really tried to remove the JS based resize and to implement the `WM_NCHITTEST`, but unfortunately as soon as the WebView2 uses the whole client area no `WM_NCHITTEST` message will be send anymore because those will be processed by WebView2. 

This is the exact reason that stopped resize from working :/

> This seems to be a limitation of WebView2 and there's some ongoing talk about this in the context of WPF on the Feedback repo. So I think for the time being the JS based resizing is the only way to go, if we wanna support resizing on frameless. Maybe there's some sophisticated way with painting everything by our own, but at least for me I would run out of spare time, especially because we currently don't need resize in our frameless app. And it would be also hard to maintain over different windows versions.

Agree.

> The border size could be retrieved from the system with the following API calls:
> 
> ```go
> x := w32.GetSystemMetrics(w32.SM_CXSIZEFRAME)
> y := w32.GetSystemMetrics(w32.SM_CYSIZEFRAME)
> ```

I was thinking of just setting those values on the frontend flags at startup if frameless was being used and not `EnableFramelessBorder`.

> Another thing we won't be able to emulate is the "Win10" feature, where the resizing could be started already outside of the window frame. Maybe this would be possible if we do a lot of drawing of the frame by ourselves, but from my opinion this is out of scope.

Yeah, I don't think we can do much about that. Thanks for your time again!

stffabi:
I've just discovered one drawback of this PR. The `Menu` won't be shown anymore, whereas before this PR it did.

@leaanthony what do you think, does a `Menu` make sense in frameless mode? I opt for if a user wants a menu in frameless, it must be done in the frontend analogous to the Close, Minimize,... buttons.

stffabi:
> I was thinking of just setting those values on the frontend flags at startup if frameless was being used and not EnableFramelessBorder.

It should always be set because even with `EnableFramelessBorder` default windows resizing won't work. The resizing of windows is disabled as soon as `WM_NCCALCSIZE` returns the whole window area and the WebView takes the whole window size.

stffabi:
Frameless window is so hard on windows...

leaanthony:
> I've just discovered one drawback of this PR. The `Menu` won't be shown anymore, whereas before this PR it did.
> 
> @leaanthony what do you think, does a `Menu` make sense in frameless mode? I opt for if a user wants a menu in frameless, it must be done in the frontend analogous to the Close, Minimize,... buttons.

It's fine. If you want a border and a menu maybe don't make a frameless window 😝 It's a fair compromise.

v2/internal/frontend/desktop/windows/window.go
stffabi: I was able to remove the `ScaleWithWindowDPI` call here.

stffabi:
@KiddoV was able to successfully test the latest changes on Windows 7, Windows 10, thanks so much. I've further tested the changes on Windows 11.

So it seems everything is fine and I'm going to mark the PR `ready for review`.

stffabi:
Thanks for merging 🚀

PR #1086, opened 2022-01-19
[linux] execute JS on main thread

tmclane:
Uses `g_idle_add` to execute the Javascript on the main thread when it is idle.
Works fine on both my amd64 and arm64 linux machines.


leaanthony (approved):

leaanthony:
Out of the park! And good to see you again 😀

PR #1092, opened 2022-01-22
Added TypeScript interface export for the window.go object

albert-sun:
The `wailsjs/runtime/runtime.d.ts` already has the `window.runtime` interface exported by default. I changed one line of code to add an export for the `window.go` interface within `wailsjs/go/bindings.d.ts` for linting purposes (the compiled JavaScript previously correctly referenced `window.go` during runtime even with the linting error).

The test located in `generate_test.go` failed as well but seems to be a false positive - it seemed to give the same error both before and after the change so I fixed it. Here's the original failure output when running `go test` in the directory:
```
--- FAIL: Test_goTypeToJSDocType (0.00s)
    --- FAIL: Test_goTypeToJSDocType/[]int (0.00s)
        generate_test.go:83: goTypeToJSDocType() = Array<number>, want Array.<number>
    --- FAIL: Test_goTypeToJSDocType/[]bool (0.00s)
        generate_test.go:83: goTypeToJSDocType() = Array<boolean>, want Array.<boolean>
FAIL
exit status 1
FAIL    github.com/wailsapp/wails/v2/internal/binding   0.006s
```

### DIFF
--- a/v2/internal/binding/generate.go
+++ b/v2/internal/binding/generate.go
@@ -129,7 +129,7 @@ func (b *Bindings) GenerateBackendTS(targetfile string) error {
 	store := b.db.store
 	var output bytes.Buffer
 
-	output.WriteString("interface go {\n")
+	output.WriteString("export interface go {\n")
 
 	var sortedPackageNames slicer.StringSlicer
 	for packageName := range store {
--- a/v2/internal/binding/generate.go
+++ b/v2/internal/binding/generate.go
@@ -201,6 +201,8 @@ declare global {
 
 func goTypeToJSDocType(input string) string {
 	switch true {
+	case input == "interface{}":
+		return "any"
 	case input == "string":
 		return "string"
 	case input == "error":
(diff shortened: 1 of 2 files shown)


leaanthony:
Thanks for identifying and taking time to fix this issue. I think it looks fine but just wondering if it affects the JSDoc, which we do output. Would the `[]` syntax mentioned on this thread work for both you think? https://stackoverflow.com/questions/32295263/how-to-document-an-array-of-objects-in-jsdoc

albert-sun:
It looks like either works for documentation purposes reading https://jsdoc.app/tags-type.html (under the sub-heading titled "An array of MyClass instances."). Also, the stray period from the tests seems to be a result of inconsistent JSDoc guidelines which have changed over the years, see: https://stackoverflow.com/questions/47822791/why-does-jsdoc-allow-both-arraystring-and-array-string-and-is-one-better-th

albert-sun:
Also added another test case for testing `interface{}` and fixed the handling by adding a case before the `number` case statement. It was getting caught because of the `int` prefix and mistakenly being documented as a number instead of `any`.

leaanthony (changes requested):
Ignore this :-)

leaanthony (approved):

PR #1109, opened 2022-01-26
chore(action): add auto sync script

misitebao:
Added site auto-sync script(8:00 and 20:00 every day) to keep mirror sites in sync with the official site at all times.

### DIFF
--- a/.github/workflows/deploy-website-to-wails.top-mirror.yml
+++ b/.github/workflows/deploy-website-to-wails.top-mirror.yml
@@ -1,10 +1,9 @@
-name: Deploy mirror | 部署镜像
+name: Sync upstream branch and deploy mirror | 同步上游分支并部署镜像
 
 on:
-  push:
-    branches: [master]
-  # pull_request:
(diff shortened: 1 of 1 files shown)


leaanthony:
Happy to do auto-sync but please explain what this is doing. Looks like we need to add secrets?

misitebao:
@leaanthony 
This is to automatically synchronize the `wailsapp/wails` warehouse to `misitebao/wails`, and rebuild and deploy the mirror website (wails.top). It only works in the `misitebao/wails` warehouse, and you don't need to do anything.

leaanthony:
I'm trying to understand why we need the upstream branch. Why not sync from master?

misitebao:
@leaanthony 
It is the upstream repository, not the upstream branch, because the mirror website construction and deployment must be done in the `misitebao/wails repository. It is impossible for me to manually maintain synchronization with the wailsapp/wails repository, so I can only use scheduled tasks to achieve it.

PR #1111, opened 2022-01-27
Create CONTRIBUTING.md

misitebao:
Subsequent Supplementary Item Catalog Description.

### DIFF
--- a/v2/CONTRIBUTING.md
+++ b/v2/CONTRIBUTING.md
@@ -1,0 +1,3 @@
+# Contributing
+
+## Directory layout description


leaanthony:
I think we should have more in this file before committing. I'll see if I can get something together over the weekend 👍

bh90210:
have we decided that we are moving forward with this option?
if yes, have we decided the best strategy or is it still up for debate? (latest relevant [doc](https://wails.app/development/))

leaanthony:
Still up for debate.

bh90210:
Firstly I vote for disabling the [wiki](https://github.com/wailsapp/wails/wiki) & `CONTRIBUTING.md` should be reflected [here](https://wails.io/docs/guides/developing-wails), through the magic of robots :stuck_out_tongue:.

IMO we should be pragmatic and learn from previous experience. For various reasons last flow strategy (gitflow) was not adhered in practice. My suggestion is to consider switching to [Release Flow](https://devblogs.microsoft.com/devops/release-flow-how-we-do-branching-on-the-vsts-team/) (or any equivalent trunk/master based flow).
Reasons behind are multifold. From a contributor's perspective the simplest move of fork > create new branch, name it something relevant > open PR against master seems to work fine. Release flow supports that without complications like `feature/*` or `bugfix/*` etc. 
Having this as main starting point for selecting a new flow/strategy, that is ease of contributors following rules, I think the next thing to consider is the release cycle (and automation). The flow/strategy should reflect the desired release cycle. If there is a formal strategy releases could benefit from automation and not be burdensome if maintainers can be sure that the flow is being followed (ie that head is truly stable).
Release flow release cycle feats a project like Wails really well. In their document they relate the releases with sprints, this wouldn't work for us, but we could replace scrum sprints with fixed time periods ie. bi-weekly or monthly. This way new features get available asap without the need of users using unstable/experimental versions and have to keep track of which version they have installed to get the latest features.
This will naturally put more pressure on contributors and reviewers but at least automation can help here.

To summarise, anything trunk based, flow should reflect the fact we encourage (and/or support) only one version - the latest*, some attention to release cycles, as simple as possible for the newcomer contributor, burden to the maintainers to adhere the flow not the other way around, some automation to save time/make contributing & reviewing more lean experience.

*this is not set in stone, but for now (now being v2 to -> v3).

leaanthony:
Yeah, the contribution page was written nearly 2 years ago when trying to maintain stable and unstable release branches. I'm in favour of ditching unstable for "use master". I'm in favour of trunk based dev - just PR to master. With release cycles, I've had experience with releases getting postponed because of 1 PR getting spun out. I'm thinking that for v2->v3 we tag new features/bug fixes once they are merged (maybe end of day). The question is how to indicate that a release has known bugs after the fact. Maybe have a feature of the cli that whenever it runs it pulls down a json file from github that has a list of known issues for that release. We can retrospectively update that and flag issues to existing users. It's just a thought...

bh90210:
I like the end of day release idea and I will raise it to upon merge!

quick thoughts/ideas on automation: 
1. After merging a PR against master automation should do the rest. All focus should be the review (for direct pushes to master maybe not but then maybe even people with write access should avoid directly pushing to master?). This will help tackle the `CHANGELOG.md` update too.
2. On PR level a trigger-comment could produce a new tag (ie branchname-test) with binaries + instructions posted in the PR so users can actually have a clear pattern of testing. Preferably download the appropriate binary for platform  > run it/test > discard it. No need to mess with original wails installation (or Go). If PR solves a bug or adds a wanted feature upon successful testing, it gets merged > user `wails update` > happy times. The way it is being done now it is adhoc, sometimes via a branch, others via a tag, other via the cli, others via go install etc etc.
3. Rebase active branches (including forked branches on active PRs) upon merge/push on master. The forked branches rebase would be tricky but I believe doable (?).

leaanthony:
I'm not sure it's efficient to produce binaries per PR. Tagging is a better approach and then people can use `wails update -pre <tag>` for testing

misitebao:
I am sorting out the editing content, which involves three files `CHANGELOG`, `CONTRIBUTING` and `CODE_OF_CONDUCT`. I will try to make it perfect.

leaanthony:
I'm all for using a `CONTRIBUTING.md` generator as a template

PR #1114, opened 2022-01-28
Feature/v2 linux menus

leaanthony:


v2/internal/frontend/desktop/linux/gtk.go
```diff
@@ -0,0 +1,73 @@
...
+extern void unblockClick(GtkWidget* menuItem, gulong handler_id);
+*/
+import "C"
+import "unsafe"
+import "github.com/wailsapp/wails/v2/pkg/menu"
+
+func GtkMenuItemWithLabel(label string) *C.GtkWidget {
+	cLabel := C.CString(label)
+	result := C.gtk_menu_item_new_with_label(cLabel)
+	C.free(unsafe.Pointer(cLabel))
```
tmclane, line 20: Do we need to worry about a similar race condition like we saw before in e713c439f0656b57bd77a06b27aa839058f5e99e?

tmclane (approved):
Looks good other than my comment / wondering about the C.free calls.

v2/internal/frontend/desktop/linux/gtk.go
leaanthony: Nah, this takes a copy immediately and stays in scope

PR #1131, opened 2022-02-07
[v2, windows] Apply suggested window size during DPI changed event

stffabi:
This adjusts the window size if the window is moved from one
monitor to another with a different DPI scaling

### DIFF
--- a/v2/go.mod
+++ b/v2/go.mod
@@ -27,7 +27,7 @@ require (
 	github.com/leaanthony/idgen v1.0.0
 	github.com/leaanthony/slicer v1.5.0
 	github.com/leaanthony/typescriptify-golang-structs v0.1.7
-	github.com/leaanthony/winc v0.0.0-20220202120322-32d63aaf77de
+	github.com/leaanthony/winc v0.0.0-20220207200219-3173d94f373e
 	github.com/leaanthony/winicon v1.0.0
 	github.com/matryer/is v1.4.0
 	github.com/olekukonko/tablewriter v0.0.4
--- a/v2/go.sum
+++ b/v2/go.sum
@@ -128,8 +128,8 @@ github.com/leaanthony/slicer v1.5.0 h1:aHYTN8xbCCLxJmkNKiLB6tgcMARl4eWmH9/F+S/0H
 github.com/leaanthony/slicer v1.5.0/go.mod h1:FwrApmf8gOrpzEWM2J/9Lh79tyq8KTX5AzRtwV7m4AY=
 github.com/leaanthony/typescriptify-golang-structs v0.1.7 h1:yoznzWzyxkO/iWdlpq+aPcuJ5Y/hpjq/lmgMFmpjwl0=
 github.com/leaanthony/typescriptify-golang-structs v0.1.7/go.mod h1:cWtOkiVhMF77e6phAXUcfNwYmMwCJ67Sij24lfvi9Js=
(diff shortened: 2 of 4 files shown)


stffabi:
This PR adjust the window size as soon as the cursor with the window reaches the monitor with the changed DPI scaling. Unfortunately the webview itself only adjust to the new scaling as soon as moving the window has finished and landed on the new monitor. As of now I was unable to trigger WebView2 to redraw itself and adjust to the new scaling factor during the move.

leaanthony:
That's a tricky problem! Perhaps there's something on the webview2feedback repo on this: https://github.com/MicrosoftEdge/WebView2Feedback

stffabi:
Couldn't find anything during a short search on the Feedback repo, but maybe using the new [ICoreWebView2Controller3](https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2controller3?view=webview2-1.0.1072.54) and setting the [RasterizationScale](https://docs.microsoft.com/en-us/microsoft-edge/webview2/reference/win32/icorewebview2controller3?view=webview2-1.0.1072.54#put_rasterizationscale) would make it work.

stffabi:
Or we could create an issue over there, maybe we get some information how to handle it from Microsoft.

sebastian0x62:
Just tested with latest .30 beta. Not working here with Windows 11. But actually, it seems the whole `window.go` file is ignored. :-( I just made a `go build ./cmd/wails` (inside v2 directory) and used the resulting `wails.exe` binary. Obviously wrong approach.

stffabi:
@sebastian0x62  Did you update your go.mod file with a replace? See also [use bleeding edge](https://wails.io/docs/guides/bleeding-edge).

stffabi:
> This PR adjust the window size as soon as the cursor with the window reaches the monitor with the changed DPI scaling. Unfortunately the webview itself only adjust to the new scaling as soon as moving the window has finished and landed on the new monitor. As of now I was unable to trigger WebView2 to redraw itself and adjust to the new scaling factor during the move.

I could narrow it down further. The problem only exists for `frameless` mode and happens because `SendMessage` in `startDrag` is a blocking call and executes a modal MessageLoop.
https://github.com/wailsapp/wails/blob/3decd70319042eae1cc71eb1194432ab124cf78c/v2/internal/frontend/desktop/windows/frontend.go#L472-L478

As a result the `processMessage` is being blocked until the moving of the window has been finished and during this time, WebView2 is unable to process further winapi messages like the `WM_DPICHANGED` and also new messages from the frontend to the backend are blocked.

stffabi:
Another issue that currently exists is the MinWidth/MinHeight handling during the DPI change might use the wrong DPI value, because `WM_DPICHANGED` and determining the DPI from the nearest monitor might be inconsistent. Adjusting the window size with `SetWindowPos` calls into `WM_GETMINMAXINFO` to get the new constraints, but if the wrong DPI is used there the constraints might be wrong.

Moving from a 175% to 100% monitor might end up in using the min constraints calculated with 175% and then the window size will be increased to match the min on the 175% monitor.

sebastian0x62:
> @sebastian0x62 Did you update your go.mod file with a replace? See also [use bleeding edge](https://wails.io/docs/guides/bleeding-edge).

@stffabi Of course not. :-( Sorry for that. Have not yet 100% internalized the process. The DPI change detection works.

stffabi:
@sebastian0x62 no worries, glad you have brought up this issue and took your time to test it. Thank you so much.

stffabi:
@leaanthony I think we are good to go. I've also included another fix for frameless window maximizing.

leaanthony:
LGTM!

PR #1152, opened 2022-02-15
[v2, windows] Do not show caption buttons if frameless and translucent+transparency

### LINKED ISSUE #1150: [v2]When EnableFramelessBorder is true, the minimize, maximize and close buttons are invalid
image

stffabi:
Fix #1150


stffabi:
@leaanthony we are good to go, just another Windows fix 😄

PR #1157, opened 2022-02-16
[v2, windows] Invert the frameless border flag and rename it

### LINKED ISSUE #1147: V2 - Improved Borderless Windows (shadows and W11 rounded edges)
On Windows 11, when using wails in frameless mode, the window is not correctly rounded and the (aero) shadow is missing. The shadow is also missing on previous windows versions.

As far as I am aware, this is because the API `DwmExtendFrameIntoClientArea` causes any shadow and rounding to disappear from a window regardless of its classes or configuration.

I was working on a different project when I noticed this and found this [blog](https://kubyshkin.name/posts/win32-window-custom-title-bar-caption/) that had already described an alternative solution for frameless windows that seems to require less work and does not need to use `DwmExtendFrameIntoClientArea`.

I've implemented this into a fork of wails and I'm interested in hearing some feedback and whether or not something like this should be made into a pull request. 

[Link to fork/commit](https://github.com/maxkunes/wails/commit/7d57a1af53e5fc6d44037ad66e8a114430f7130f)

stffabi:
This is a breaking change for all users of the flag.

Fix #1147

### DIFF
--- a/v2/internal/frontend/desktop/windows/window.go
+++ b/v2/internal/frontend/desktop/windows/window.go
@@ -151,7 +151,7 @@ func (w *Window) WndProc(msg uint32, wparam, lparam uintptr) uintptr {
 			// This Option is not affected by returning 0 in WM_NCCALCSIZE.
 			// As a result we have hidden the titlebar but still have the default window frame styling.
 			// See: https://docs.microsoft.com/en-us/windows/win32/api/dwmapi/nf-dwmapi-dwmextendframeintoclientarea#remarks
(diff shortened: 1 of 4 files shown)


leaanthony (approved):
This is 💯 Chinese docs too? 🔥🔥🔥

stffabi:
Well, deepl.com was my friend 😄. Hopefully @misitebao will have a look into the Chinese docs and improve it.

maxkunes:
Looks good to me. Thanks @stffabi !

Debdut:
I don't have rounded corners but on Mac? Is this a bug or am I missing something?

Screen Shot 2022-08-30 at 1 53 10 PM

PR #1158, opened 2022-02-16
[v2] Consolidate processRequest, improve logging and error handling

### LINKED ISSUE #1055: V2 panics using Svelte template
Hello, having some issues trying to use the Svelte template. Thanks in advance and I really love what you are doing!

**Description**
When launching `npm run dev` in a second  terminal as per the directions at `https://github.com/raitonoberu/wails-svelte-template`, Wails panics as follows:

```
panic: Unexpected host for request on wails:// scheme
goroutine 7 [running]:
github.com/wailsapp/wails/v2/internal/frontend/desktop/darwin.(*Frontend).processRequest(0xc00019e840, 0xc0001e6210)
	/Users/billmatlock/go/pkg/mod/github.com/wailsapp/wails/v2@v2.0.0-beta.27/internal/frontend/desktop/darwin/frontend.go:288 +0x2a5
github.com/wailsapp/wails/v2/internal/frontend/desktop/darwin.(*Frontend).startRequestProcessor(0x0)
	/Users/billmatlock/go/pkg/mod/github.com/wailsapp/wails/v2@v2.0.0-beta.27/internal/frontend/desktop/darwin/frontend.go:105 +0x48
created by github.com/wailsapp/wails/v2/internal/frontend/desktop/darwin.NewFrontend
	/Users/billmatlock/go/pkg/mod/github.com/wailsapp/wails/v2@v2.0.0-beta.27/internal/frontend/desktop/darwin/frontend.go:92 +0x2b1
````
	
	
This also occurs when using the built-in Svelte template installed via: `wails init -n myapp -t svelte -g`

**To Reproduce**
I'm installing this via either of the following methods: 
`wails init -n myapp -t https://github.com/raitonoberu/wails-svelte-template -g`

OR

`wails init -n myapp -t svelte -g`

Then going into `/myapp/frontend` and running:
`npm install`

Then opening a terminal in `/myapp` and running:
`wails dev`

Then in another terminal window in `/myapp/frontend` and running:
`npm run dev`

At this point, running `npm run dev` causes a panic in the first window

**Expected behaviour**
No errors. The Wails app should run and monitor changes in Go files, while the command `npm run dev` watches for changes in web files and rebuilds as necessary.

**System Details**

```
Node: 17.3.0


wails doctor
Wails CLI v2.0.0-beta.27


System
------
OS:		MacOS
Version: 	11.6.2
ID:		199506
Go Version:	go1.17.3
Platform:	darwin
Architecture:	amd64

xcode command line tools 	N/A		Installed	2384
npm 				N/A		Installed	8.3.0
*upx 				N/A		Available	

* - Optional Dependency

Diagnosis
---------
```

**Additional context**
If I do `npm run dev` first, then in another terminal run `wails dev` everything is fine until I try changing something in a web asset. When Svelte rebuilds, Go panics in the same manner.

### PR DESCRIPTION
**Pull request by stffabi**
Closes #1055

### DIFF
--- a/v2/internal/frontend/desktop/common/process_request.go
+++ b/v2/internal/frontend/desktop/common/process_request.go
@@ -1,0 +1,55 @@
+package common
+
+import (
+	"fmt"
+	"net/http"
+	"os"
+	"strings"
+
+	"github.com/wailsapp/wails/v2/internal/frontend/assetserver"
+)
+
+type RequestRespone struct {
+	Body       []byte
+	MimeType   string
+	StatusCode int
+}
+
+func (r RequestRespone) StatusText() string {
+	return http.StatusText(r.StatusCode)
+}
+
+func (r RequestRespone) String() string {
+	return fmt.Sprintf("Body: '%s', StatusCode: %d", string(r.Body), r.StatusCode)
+}
+
+func ProcessRequest(uri string, assets *assetserver.DesktopAssetServer, expectedScheme string, expectedHosts ...string) (RequestRespone, error) {
+	// Translate URI to file
+	file, err := translateUriToFile(uri, expectedScheme, expectedHosts...)
+	if err != nil {
+		if err == ErrUnexpectedHost {
+			body := fmt.Sprintf("expected host one of \"%s\"", strings.Join(expectedHosts, ","))
+			return textResponse(body, http.StatusInternalServerError), err
+		}
+
+		return RequestRespone{StatusCode: http.StatusInternalServerError}, err
+	}
+
+	content, mimeType, err := assets.Load(file)
+	if err != nil {
+		statusCode := http.StatusInternalServerError
+		if os.IsNotExist(err) {
+			statusCode = http.StatusNotFound
+		}
+		return RequestRespone{StatusCode: statusCode}, err
+	}
+
+	return RequestRespone{Body: content, MimeType: mimeType, StatusCode: http.StatusOK}, nil
+}
+
+func textResponse(body string, statusCode int) RequestRespone {
+	if body == "" {
+		return RequestRespone{StatusCode: statusCode}
+	}
+	return RequestRespone{Body: []byte(body), MimeType: "text/plain;charset=UTF-8", StatusCode: statusCode}
+}
--- a/v2/internal/frontend/desktop/common/uri_translate.go
+++ b/v2/internal/frontend/desktop/common/uri_translate.go
@@ -1,20 +1,34 @@
 package common
 
-import "net/url"
+import (
+	"fmt"
+	"net/url"
+)
 
-func TranslateUriToFile(uri string, expectedScheme string, expectedHost string) (file string, match bool, err error) {
+var ErrUnexpectedScheme = fmt.Errorf("unexpected scheme")
+var ErrUnexpectedHost = fmt.Errorf("unexpected host")
+
+func translateUriToFile(uri string, expectedScheme string, expectedHosts ...string) (file string, err error) {
 	url, err := url.Parse(uri)
 	if err != nil {
-		return "", false, err
+		return "", err
 	}
 
-	if url.Scheme != expectedScheme || url.Host != expectedHost {
-		return "", false, nil
+	if url.Scheme != expectedScheme {
+		return "", ErrUnexpectedScheme
 	}
 
-	filePath := url.Path
-	if filePath == "" {
-		filePath = "/"
+	for _, expectedHost := range expectedHosts {
+		if url.Host != expectedHost {
+			continue
+		}
+
+		filePath := url.Path
+		if filePath == "" {
+			filePath = "/"
+		}
+		return filePath, nil
 	}
-	return filePath, true, nil
+
+	return "", ErrUnexpectedHost
 }
--- a/v2/internal/frontend/desktop/darwin/frontend.go
+++ b/v2/internal/frontend/desktop/darwin/frontend.go
@@ -16,10 +16,8 @@ import "C"
 import (
 	"context"
 	"encoding/json"
-	"fmt"
 	"html/template"
 	"log"
-	"os"
 	"strconv"
 	"unsafe"
 
@@ -277,40 +275,21 @@ func (f *Frontend) ExecJS(js string) {
 func (f *Frontend) processRequest(r *request) {
 	uri := C.GoString(r.url)
 
-	var _contents []byte
-	var _mimetype string
-
-	// Translate URI to file
-	file, match, err := common.TranslateUriToFile(uri, "wails", "wails")
-	if err == nil {
-		if !match {
-			// This should never happen on darwin, because we get only called for wails://
-			panic("Unexpected host for request on wails:// scheme")
-		}
-
-		// Load file from asset store
-		_contents, _mimetype, err = f.assets.Load(file)
-	}
-
-	statusCode := 200
+	res, err := common.ProcessRequest(uri, f.assets, "wails", "wails")
 	if err != nil {
-		if os.IsNotExist(err) {
-			statusCode = 404
-		} else {
-			err = fmt.Errorf("Error processing request %s: %w", uri, err)
-			f.logger.Error(err.Error())
-			statusCode = 500
-		}
+		f.logger.Error("Error processing request '%s': %s (HttpResponse=%s)", uri, err, res)
 	}
 
-	var data unsafe.Pointer
-	if _contents != nil {
-		data = unsafe.Pointer(&_contents[0])
+	var content unsafe.Pointer
+	var contentLen int
+	if _contents := res.Body; _contents != nil {
+		content = unsafe.Pointer(&_contents[0])
+		contentLen = len(_contents)
 	}
-	mimetype := C.CString(_mimetype)
+	mimetype := C.CString(res.MimeType)
 	defer C.free(unsafe.Pointer(mimetype))
 
-	C.ProcessURLResponse(r.ctx, r.url, C.int(statusCode), mimetype, data, C.int(len(_contents)))
+	C.ProcessURLResponse(r.ctx, r.url, C.int(res.StatusCode), mimetype, content, C.int(contentLen))
 }
 
 //func (f *Frontend) processSystemEvent(message string) {
--- a/v2/internal/frontend/desktop/linux/frontend.go
+++ b/v2/internal/frontend/desktop/linux/frontend.go
@@ -14,8 +14,8 @@ import "C"
 import (
 	"context"
 	"encoding/json"
-	"fmt"
 	"log"
+	"net/http"
 	"os"
 	"strconv"
 	"text/template"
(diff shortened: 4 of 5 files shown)

### CODE REVIEW

**Comment by stffabi** [comment]
As mentioned in the slack thread here's a proposal to consolidate and improve process request across all platforms. I've tested it on Windows and Darwin, would be great if @leaanthony or @tmclane could test it on linux.

File: v2/internal/frontend/desktop/common/uri_translate.go [hunk 1]
```diff
@@ -1,20 +1,34 @@
...
-	if url.Scheme != expectedScheme || url.Host != expectedHost {
-		return "", false, nil
+	if url.Scheme != expectedScheme {
+		return "", ErrUnexpectedScheme
 	}
 
-	filePath := url.Path
-	if filePath == "" {
-		filePath = "/"
+	for _, expectedHost := range expectedHosts {
```
leaanthony, line 21: On some platforms, I've seen a nil host come through. Is there a reason to consider the host at all?

stffabi: On Linux and Macos the callback is only called for the wails:// scheme, so we could map all hosts to the asset fs if we want to.
On Windows the callback is called for all requests and the assets are under file://wails. Currently we let handle all other file:// requests by the default handler by just returning from the callback. So thats to detect this case, but we could also detect that case in the windows callback itself.

stffabi: I think for the case mentioned on the slack thread it would get more information for the user if he knows that the host was wrong and a 500 is returned than just a 404.
In this implementation the log entry specificly mentions the host was wrong.

leaanthony: I was holding off until Linux beta was released then it slipped off the radar. Will try testing today.

leaanthony:
@bh90210 fancy testing this?

bh90210:
heh heh I was in the middle of doing this :-p
I just reproduced the original problem.. test/review incoming

bh90210 (commented):
First of all it works :smiley:.

I am very pro loading with error on terminal than crushing in panic.
I do have some comments for here and there but the biggest one, is there an obvious justification of `process_request.go` over going for better error handling in `func TranslateUriToFile(uri string, expectedScheme string, expectedHost string) (file string, match bool, err error)` and `func (f *Frontend) processRequest(request unsafe.Pointer)`?

stffabi:
It's to remove code duplication. Basically all 3 platforms did the same, calling TranslateUri then calling into the assets to load the file, then mapping NotExists errors to a 404 http response or other errors to a 500 error.
Having the common ProcessRequest is to have only one location which basically maps from the Request to the Http response in one place. So the platform frontends only handle the platform specific code from getting the url string, passing it to the common ProcessRequest, and then only handle mapping back the response to the platform specific code.
I think this greatly improves code quality and ensures all platforms behave in the same way. If other errors want to be mapped to http statuscodes this only has to be done in one place in the future.

v2/internal/frontend/desktop/common/uri_translate.go
```diff
@@ -1,20 +1,34 @@
...
 
-import "net/url"
+import (
+	"fmt"
+	"net/url"
+)
 
-func TranslateUriToFile(uri string, expectedScheme string, expectedHost string) (file string, match bool, err error) {
+var ErrUnexpectedScheme = fmt.Errorf("unexpected scheme")
+var ErrUnexpectedHost = fmt.Errorf("unexpected host")
```
bh90210, line 9: more of a stylistic suggestion coming from the rest of the code base
```
// ErrUnexpectedScheme some explanation, ie when does this error can arise.
// Link to github issue maybe?
const ErrUnexpectedScheme = "unexpected scheme"
// ErrUnexpectedHost same.
const ErrUnexpectedHost = "unexpected host"
```
and `fmt.Error(ErrUnexpectedScheme)` when returning.

v2/internal/frontend/desktop/common/process_request.go
```diff
@@ -0,0 +1,55 @@
...
+func (r RequestRespone) String() string {
+	return fmt.Sprintf("Body: '%s', StatusCode: %d", string(r.Body), r.StatusCode)
+}
+
+func ProcessRequest(uri string, assets *assetserver.DesktopAssetServer, expectedScheme string, expectedHosts ...string) (RequestRespone, error) {
+	// Translate URI to file
+	file, err := translateUriToFile(uri, expectedScheme, expectedHosts...)
+	if err != nil {
+		if err == ErrUnexpectedHost {
+			body := fmt.Sprintf("expected host one of \"%s\"", strings.Join(expectedHosts, ","))
```
bh90210, line 31:
```
ERR | Error processing request 'wails://helo.myfonts.net/count/3e7d84': unexpected host (HttpResponse=BODY: 'expected host one of ",null"', StatusCode: 500)
```
Feedback given to the user is a little confusing here?

v2/internal/frontend/desktop/common/process_request.go
```diff
@@ -0,0 +1,55 @@
...
+
+func (r RequestRespone) StatusText() string {
+	return http.StatusText(r.StatusCode)
+}
+
+func (r RequestRespone) String() string {
+	return fmt.Sprintf("Body: '%s', StatusCode: %d", string(r.Body), r.StatusCode)
+}
+
+func ProcessRequest(uri string, assets *assetserver.DesktopAssetServer, expectedScheme string, expectedHosts ...string) (RequestRespone, error) {
```
bh90210, line 26: Afterthought: some documentation here it would be nice.

v2/internal/frontend/desktop/common/process_request.go
```diff
@@ -0,0 +1,55 @@
...
+	// Translate URI to file
+	file, err := translateUriToFile(uri, expectedScheme, expectedHosts...)
+	if err != nil {
+		if err == ErrUnexpectedHost {
+			body := fmt.Sprintf("expected host one of \"%s\"", strings.Join(expectedHosts, ","))
+			return textResponse(body, http.StatusInternalServerError), err
+		}
+
+		return RequestRespone{StatusCode: http.StatusInternalServerError}, err
```
bh90210, line 35: I would have thought that this should be 400/Bad Request? From the point of view of the backend server..

v2/internal/frontend/desktop/common/process_request.go
leaanthony: How did you manage to get that output?

stffabi: Any suggestions what you would like to see in the output?

v2/internal/frontend/desktop/common/uri_translate.go
stffabi: Shouldn't we try to keep well established patterns, especially patterns that are widely used in the go standard lib? Cause that's how gophers are used too and makes it easier to get familiar with the code base. Defining string constants and returning always a new error would make it impossible to use `errors.Is` or `errors.As` and especially it would also not be possible to check if the error has been wrapped and is somewhere in the error chain. Simple comparisons would need to be done by first getting the error string and comparing on that one. AFAIK I've never seen such a pattern in the go standard lib.
I've also seen that the PR isn't up to date with my latest version, I've replaced the error comparisons with `errors.Is` but haven't yet pushed it because I didn't know how @leaanthony wanted the proceed after his first comment.

As for adding comments I agree with you.

v2/internal/frontend/desktop/common/process_request.go
stffabi: Yeah, that makes absolutely sense.

v2/internal/frontend/desktop/common/process_request.go
bh90210: Well it just happens! As Lea said on a comment above it just returns null.
I don't know what the intention was originally (show multiple hostnames that one of them if the one user should use?). Looking at it, it is not very clear why there would be multiple hostnames either, so I don't have a suggestion really.
What do you see on Mac/Win? Are we trying to let the user know that some of their assets can not load because they point to the wrong hostname?

stffabi: On MacOS an Windows you see only one suggestion because there's only one hostname that comes through for the assets, which is "wails". Normally users don't have to decide on a hostname in any way, they just have to correctly use relative paths.
Well the two hostnames special case on Linux was already there and I didn't change that. As for that I just output both possibilities, to be fair I'm currently not using Wails on Linux so I have no insights on why there are coming requests with "null" as a hostname.

leaanthony: Yeah, I've seen this happen and I'm not sure why. I added code to just "allow" nil hosts and treat it as internal. I think it's fine as is.

v2/internal/frontend/desktop/common/uri_translate.go
leaanthony: @stffabi - push changes, all good. I think this is 99% there and ready to merge.

v2/internal/frontend/desktop/common/process_request.go
bh90210: @leaanthony to reproduce I ran `wails init -n projectname` and added this line to `frontend/src/main.css`
```css
@import url("//helo.myfonts.net/count/3e7d84");
```
which was the original problem. Maybe in the error message hint that probably an incorrect path somewhere in the frontend might be the issue so users have some lead?

tmclane:
I would recommend putting a 404 in the output when it can't be found with the full path of the URI being searched for such that the developer can locate that string in the front end code. Anything else they'll never be able to find.
The wails prefix won't make sense to them. This is an odd case since the string being searched for was actually invalid and caused an invalid local search.

stffabi:
> I would recommend putting a 404 in the output when it can't be found with the full path of the URI being searched for such that the developer can locate that string in the front end code. Anything else they'll never be able to find. The wails prefix won't make sense to them. This is an odd case since the string being searched for was actually invalid and caused an invalid local search.

Well I see no way of getting the original URI in reliable way because if someone uses `//helo.myfonts.net/count/3e7d84` that will get translated to `wails://helo.myfonts.net/count/3e7d84` by the webview. So we get as request URI `wails://helo.myfonts.net/count/3e7d84` in ProcessRequest. There would be only the possibility to strip away "wails:" in the case of an unexpected host has been supplied. I wouldn't count this as a reliable way to determine the original URI. 

Showing the URI provided from the webview  (which contains "wails://") would allow the user to open the developer tools and to correlate the Wails error message with the developer tools, which also shows `wails://helo.myfonts.net/count/3e7d84`.

Furthermore I see no benefit in using a 404 than using a 400 (BadRequest) in this case. I as a developer like to have more insight whats going on, and for me this case is not only just a FileNotFound. So I would prefer to see something other than 404.

leaanthony:
How will it be possible to use `//` urls in a project? Windows appears to be ok with it. Just wondering if we should try stripping `wails:` and trying to fetch that data for the frontend...

stffabi:
> How will it be possible to use `//` urls in a project? Windows appears to be ok with it

That could be done like @bh90210 mentioned or how the slack thread opener had it. It's also possible on Darwin, that's what the thread opener used.

> Just wondering if we should try stripping `wails:` and trying to fetch that data for the frontend...

Well then we could simply ignore the "Unexpected host part" check and that's it. And we are back again at the discussion of your first comment 9 days ago, where there was no final dicision.

Well from my side this PR did cost me too much time discussing things, e.g. about well established patterns and other things that could be read out of the code of the PR. Especially if it doesn't solve any real problem for our company.
So I'll leave the PR open with "allow edits from maintainers"  so it could get improved, changed or I'm also fine if the consense is to abandon it and doing something other.

leaanthony:
I think it's basically there. Sorry you felt it took too much time. I'm mostly to blame for that 🙏

leaanthony:
I'm merging this as is. We'll deal with the `//` separately. Amazing work @stffabi and thanks 🙏

Issue #1165, opened 2022-02-20
[v2]An error occurs when creating a new project with a template generated by the Generate Template command


misitebao:
**Description**
Using the template generated by the `wails generate template` command to create a new project will cause errors in the generated project file.

**To Reproduce**
Steps to reproduce the behaviour:
1. Run `wails generate template wails-template-demo`
2. Run `wails init -n Wails-Demo -t ./wails-template-demo`
3. Run `wails dev`

**Expected behaviour**
It is hoped that the steps of the third step can be optimized so that the program can run the build process correctly.

**Screenshots**
This is the file generated after running the `wails dev` command.
![image](https://user-images.githubusercontent.com/28185258/154869357-a3533dce-2021-4baf-adbc-8ded3779ff60.png)
![动画](https://user-images.githubusercontent.com/28185258/155225649-3b81cb30-d0d2-4d63-afa1-fbe5e1c61258.gif)

**System Details**
```shell
$ wails doctor
Wails CLI v2.0.0-beta.30


Scanning system - Please wait (this may take a long time)...Done.

System
------
OS:             Windows 10 Pro
Version:        2009 (Build: 19044)
ID:             21H2
Go Version:     go1.17.6
Platform:       windows
Architecture:   amd64

Dependency      Package Name    Status          Version
----------      ------------    ------          -------
WebView2        N/A             Installed       98.0.1108.56
npm             N/A             Installed       8.5.1
*upx            N/A             Available

* - Optional Dependency

Diagnosis
---------
Your system is ready for Wails development!
Optional package(s) installation details:
  - upx : Available at https://upx.github.io/

If Wails is useful to you or your company, please consider sponsoring the project:
https://github.com/sponsors/leaanthony
```

**Additional context**
As to why this error occurs, I'm guessing because running `wails dev` causes the `frontend:install` command in `wails.json` to be run, which is supposed to be in `frontend` directory operation, probably because `wails` does not limit the directory, so this command is run in the project root directory, so unnecessary files (`package-lock.json` and `node_modules` directories) are generated.

There are two suggestions for a solution to this problem:
1. Optimize the `wails dev` and `wails build` commands so that they can correctly identify the directory where the front-end project is located instead of the `package.json` file in all directories. For example, you can set a field to specify the directory where the front-end project is located.
2. The project build script files are uniformly moved to the `scripts` directory in the root directory.

bh90210:
I can not reproduce the problem on linux.
I do
1. `wails generate template -name test`
[...]

PR #1166, opened 2022-02-21
feat(cli):optimized template generation

misitebao:
The template generation command has been optimized and has been tested in the local win10 environment.


leaanthony:
Thanks for this! Are the scripts for the person building a template or for projects that are generated?

misitebao:
> Thanks for this! Are the scripts for the person building a template or for projects that are generated?

These scripts are used to generate the project, their role is to replace the `package.json` file in the `base` directory, and at the same time solve the problems mentioned in [https://github.com/wailsapp/wails/issues/1165](https://github.com/wailsapp/wails/issues/1165).

leaanthony:
I'm ok adding the scripts but they should be mentioned on the main docs site. Are you comfortable adding that?

misitebao:
@leaanthony 
I don’t think the script file must exist for the project. The template project can choose whether to need this or not. We add the script to the basic template just to give developers a suggestion, not to require all templates to have scripts. Developers can also choose to manually Install dependencies and execute build commands.

leaanthony:
Yeah ok, I'm alright with that.

Issue #1155, opened 2022-02-15
Single source of truth for manifest data


ValentinTrinque:
**Is your feature request related to a problem? Please describe.**
Maintaining OS specific "manifest" (`darwin/Info.plist`, `windows/<project>.exe.manifest`, `windows/info.json`) is a bit tedious, especially because they contain a lot of duplication. And updating them is error prone.

**Describe the solution you'd like**
Having a single source of truth, such as `wails.json`, that will be used to populate the content of these manifests.

This will require the `ProjectData` to be used to generate a default configuration for each OS. If a manifest already exists, and that the common properties are defined then the manifest values takes precedence. Basically, a merge.

So, during packaging:
1. No manifest -> auto-generation from `wails.json`
2. Existing manifest -> merge with priority to the manifest values.

**Describe alternatives you've considered**
None.

**Additional context**
N/A

_Of course, if I can help getting this done..._ :)

leaanthony:
It's a good suggestion. Will happily accept a PR for this 👍

PR #1184, opened 2022-02-25
[v2] NSIS installer support for Windows

### LINKED ISSUE #1135: NSIS Support for Windows
Create an installer option for Windows builds.

stffabi:
- Brings support for NSIS installer building for Windows
- Adds support for build-level build-hooks
- Populate manifest and app info files from `wails.json' by leveraging go templating.

Closes #1135, #1155


leaanthony:
@jponc

v2/pkg/commands/build/build.go
```diff
@@ -213,8 +214,63 @@ func Build(options *Options) (string, error) {
...
+			continue
+		}
+		args[i] = newArg
+	}
+
+	if options.Verbosity == VERBOSE {
+		outputLogger.Println("%s", strings.Join(args, " "))
+	}
+
+	_, stderr, err := shell.RunCommand(options.BuildDirectory, args[0], args[1:]...)
```
leaanthony, line 269: This might throw an error if the build hook is a single command?

v2/pkg/buildassets/build/windows/installer/wails_tools.nsh
```diff
@@ -0,0 +1,155 @@
...
+# Install webview2 by launching the bootstrapper
+# See https://docs.microsoft.com/en-us/microsoft-edge/webview2/concepts/distribution#online-only-deployment
+!macro wails.webview2runtime
+    !ifndef WAILS_INSTALL_WEBVIEW_DETAILPRINT
+        !define WAILS_INSTALL_WEBVIEW_DETAILPRINT "Installing: WebView2 Runtime"
+    !endif
+
+    SetRegView 64
+	# If this key exists and is not empty then webview2 is already installed
+	ReadRegStr $0 HKLM "SOFTWARE\WOW6432Node\Microsoft\EdgeUpdate\Clients\{F3017226-FE2A-4295-8BDF-00C3A9A7E4C5}" "pv"
```
leaanthony, line 140: There are 2 keys to check for installation depending on whether it was installed by an admin or not:
HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Microsoft\EdgeUpdate\Clients\{F3017226-FE2A-4295-8BDF-00C3A9A7E4C5}

HKEY_CURRENT_USER\Software\Microsoft\EdgeUpdate\Clients\{F3017226-FE2A-4295-8BDF-00C3A9A7E4C5}

v2/pkg/buildassets/buildassets.go
```diff
@@ -1,67 +1,145 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	data := &assetData{
+		Name: projectData.Name,
+		Info: projectData.Info,
+	}
+
+	data.Internal.Webview2RuntimeSetupPath = filepath.Join(fs.RelativePath("../../"), "internal", "webview2runtime", "MicrosoftEdgeWebview2Setup.exe")
```
leaanthony, line 123: We're moving away from assuming the Wails CLI is installed as a Go module so we don't want to reference files relative to the source. I think we should copy this setup file into a temp directory. I've added a [method](https://github.com/wailsapp/wails/blob/master/v2/internal/webview2runtime/webview2runtime.go#L175) to the webview2runtime package to do this. Perhaps we could use `os.TempDir()`?

leaanthony (changes requested):
Thanks for this - it's awesome stuff! Just a couple of comments to discuss and I think we're good to go.

v2/pkg/commands/build/build.go
stffabi: That's fine, subslicing with the len of the slice always returns an empty slice. Becauce it's checked that the hook is never "", the slice always contains at least one element.
https://go.dev/play/p/h9m2XQ8YiJI

v2/pkg/buildassets/build/windows/installer/wails_tools.nsh
stffabi: As the installer per default is run elevated it makes sense to only check against the admin key. If we wouldn't do that, an installer which is run elevated and installs the app for all users, the app would only work for that specific user and not for all user. Or does an installation by a non admin user work for other users on the same machine?

This check is also just a fastpath, if it doesn't work the bootstrapper will detect the runtime is installed and won't do anything. So it would be just a waste of time and probably bandwidth during the installation.

I think the best would be to do the following:
- If the installer is run elevated just check against the admin key.
- If the installer is not run elevated, first check against the admin key and then the current user key.

leaanthony: The reason I bring it up is that running it elevated was a temporary work around as MS did a breaking change (they created a user install key but didn't publish it). Installing the runtime with user privileges is the best scenario as it'll work in all environments (eg corporate). I reverted the "admin install" change today as we don't need it any more (we detect installation by asking `webview2loader.dll` what version is installed. I agree with your summary at the end.

v2/pkg/buildassets/buildassets.go
leaanthony: Yeah, I thought that initially but then it'd be easy to check in with your project. The reason I suggested temp dir was because I assumed it would be deleted anyway.

v2/pkg/buildassets/build/windows/installer/wails_tools.nsh
stffabi: Yeah, I agree with that for the embedded install case. For the installer it's slightly different, and for corporate environments the installation must almost always be performed by an admin.

So I'll go ahead and change the NSIS tools accordingly.

v2/pkg/buildassets/buildassets.go
leaanthony: Given everything you've said, I'd suggest dumping it in the build dir and add a .gitignore is fine 👍

stffabi: Great, I'll update the PR on monday

v2/pkg/buildassets/build/windows/installer/wails_tools.nsh
stffabi: I didn't find a reliable way to determine the execution level of the installer that satisfied me. Therefore I've introduced a new define `!define REQUEST_EXECUTION_LEVEL "admin"` that user can use to change the execution level. This define is then taken into account in the `wails_tools.nsh` to also check the user specific key if needed.

v2/pkg/buildassets/buildassets.go
stffabi: .gitignore entries have been added.

leaanthony:
Thanks for all the hard work on this 🙏

stffabi:
Awesome, thanks so much for merging.

PR #1200, opened 2022-03-02
Fix linux unexpected signal errors

ianmjones:
As seen in discussion https://github.com/wailsapp/wails/discussions/1156, I've noticed a few unexpected signal errors and panics when running a Wails app in NixOS.

This PR fixes the errors I've seen:

1. When quitting the app it would always fatal error because the window reference was released before destroy was called on it.
2. There was often an error related to setting the title on the window due to a race condition.
3. Clicking a blank/background area in the window caused a crash when drag move was initiated.

### DIFF
--- a/v2/internal/frontend/desktop/linux/window.go
+++ b/v2/internal/frontend/desktop/linux/window.go
@@ -192,27 +192,18 @@ void loadIndex(void* webview) {
 }
 
 typedef struct DragOptions {
-	void *webview;
-	GtkWindow* mainwindow;
+	void* webview;
+	void* mainwindow;
 } DragOptions;
 
-static void startDrag(gpointer data)
-{
+void startDrag(gpointer* data) {
 	DragOptions* options = (DragOptions*)data;
 
-   // Ignore non-toplevel widgets
-   GtkWidget *window = gtk_widget_get_toplevel(GTK_WIDGET(options->webview));
-   if (!GTK_IS_WINDOW(window)) return;
+	// Ignore non-toplevel widgets
+	GtkWidget *window = gtk_widget_get_toplevel(GTK_WIDGET(options->webview));
+	if (!GTK_IS_WINDOW(window)) return;
 
-   gtk_window_begin_move_drag(options->mainwindow, 1, xroot, yroot, dragTime);
-	free(data);
-}
-
-static void StartDrag(void *webview, GtkWindow* mainwindow) {
-	DragOptions* data = malloc(sizeof(DragOptions));
-	data->webview = webview;
-	data->mainwindow = mainwindow;
-	ExecuteOnMainThread(startDrag, (gpointer)data);
+	gtk_window_begin_move_drag(options->mainwindow, 1, xroot, yroot, dragTime);
 }
 
 typedef struct JSCallback {
@@ -731,7 +722,11 @@ func (w *Window) ExecJS(js string) {
 }
 
 func (w *Window) StartDrag() {
-	C.StartDrag(w.webview, w.asGTKWindow())
+	data := C.DragOptions{
+		webview:    w.webview,
+		mainwindow: w.gtkWindow,
+	}
+	C.ExecuteOnMainThread(C.startDrag, C.gpointer(&data))
 }
 
 func (w *Window) Quit() {
--- a/v2/internal/frontend/desktop/linux/window.go
+++ b/v2/internal/frontend/desktop/linux/window.go
@@ -413,25 +413,6 @@ void setRGBA(gpointer* data) {
 	webkit_web_view_set_background_color(WEBKIT_WEB_VIEW(options->webview), &colour);
 }
 
-typedef struct SetTitleArgs {
-	GtkWindow* window;
-	char* title;
-} SetTitleArgs;
-
-void setTitle(gpointer data) {
-	SetTitleArgs* args = (SetTitleArgs*)data;
-	gtk_window_set_title(args->window, args->title);
-	free((void*)args->title);
-	free((void*)data);
-}
-
-void SetTitle(GtkWindow* window, char* title) {
-	SetTitleArgs* args = malloc(sizeof(SetTitleArgs));
-	args->window = window;
-	args->title = title;
-	ExecuteOnMainThread(setTitle, (gpointer)args);
-}
-
 typedef struct SetPositionArgs {
 	int x;
 	int y;
@@ -736,7 +717,9 @@ func (w *Window) SetDecorated(frameless bool) {
 }
 
 func (w *Window) SetTitle(title string) {
-	C.SetTitle(w.asGTKWindow(), C.CString(title))
+	newTitle := C.CString(title)
+	defer C.free(unsafe.Pointer(newTitle))
+	C.gtk_window_set_title(w.asGTKWindow(), newTitle)
 }
 
 func (w *Window) ExecJS(js string) {
--- a/v2/internal/frontend/desktop/linux/window.go
+++ b/v2/internal/frontend/desktop/linux/window.go
@@ -197,13 +197,15 @@ typedef struct DragOptions {
 	GtkWindow* mainwindow;
 } DragOptions;
 
-static gboolean startDrag(gpointer data)
-{
+static gboolean startDrag(gpointer data) {
 	DragOptions* options = (DragOptions*)data;
 
 	// Ignore non-toplevel widgets
 	GtkWidget *window = gtk_widget_get_toplevel(GTK_WIDGET(options->webview));
-	if (!GTK_IS_WINDOW(window)) return G_SOURCE_REMOVE;
+	if (!GTK_IS_WINDOW(window)) {
+		free(data);
+		return G_SOURCE_REMOVE;
+	}
 
 	gtk_window_begin_move_drag(options->mainwindow, 1, xroot, yroot, dragTime);
 	free(data);
--- a/v2/internal/frontend/desktop/linux/window.go
+++ b/v2/internal/frontend/desktop/linux/window.go
@@ -57,8 +57,7 @@ GdkRectangle getCurrentMonitorGeometry(GtkWindow *window) {
     return result;
 }
 
-void Center(gpointer data)
-{
+gboolean Center(gpointer data) {
 	GtkWindow *window = (GtkWindow*)data;
 
     // Get the geometry of the monitor
@@ -73,6 +72,8 @@ void Center(gpointer data)
 
     // Place the window at the center of the monitor
     gtk_window_move(window, newX, newY);
+
+    return G_SOURCE_REMOVE;
 }
 
 int IsFullscreen(GtkWidget *widget) {
@@ -192,18 +193,29 @@ void loadIndex(void* webview) {
 }
 
 typedef struct DragOptions {
-	void* webview;
-	void* mainwindow;
+	void *webview;
+	GtkWindow* mainwindow;
 } DragOptions;
 
-void startDrag(gpointer* data) {
+static gboolean startDrag(gpointer data)
+{
 	DragOptions* options = (DragOptions*)data;
 
 	// Ignore non-toplevel widgets
 	GtkWidget *window = gtk_widget_get_toplevel(GTK_WIDGET(options->webview));
-	if (!GTK_IS_WINDOW(window)) return;
+	if (!GTK_IS_WINDOW(window)) return G_SOURCE_REMOVE;
 
 	gtk_window_begin_move_drag(options->mainwindow, 1, xroot, yroot, dragTime);
+	free(data);
+
+	return G_SOURCE_REMOVE;
+}
+
+static void StartDrag(void *webview, GtkWindow* mainwindow) {
+	DragOptions* data = malloc(sizeof(DragOptions));
+	data->webview = webview;
+	data->mainwindow = mainwindow;
+	ExecuteOnMainThread(startDrag, (gpointer)data);
 }
 
 typedef struct JSCallback {
@@ -211,7 +223,7 @@ typedef struct JSCallback {
     char* script;
 } JSCallback;
 
-int executeJS(gpointer data) {
+gboolean executeJS(gpointer data) {
     struct JSCallback *js = data;
     webkit_web_view_run_javascript(js->webview, js->script, NULL, NULL, NULL);
     free(js->script);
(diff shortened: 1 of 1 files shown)


v2/internal/frontend/desktop/linux/window.go
```diff
@@ -736,7 +708,9 @@ func (w *Window) SetDecorated(frameless bool) {
 }
 
 func (w *Window) SetTitle(title string) {
-	C.SetTitle(w.asGTKWindow(), C.CString(title))
+	newTitle := C.CString(title)
+	defer C.free(unsafe.Pointer(newTitle))
+	C.gtk_window_set_title(w.asGTKWindow(), newTitle)
```
leaanthony, line 713: I think this operation needs to be on the main thread. If not, it can bork on ARM64 arch. The original code created the C.CString which was freed at the end of the `SetTitle` C method, which is what this change does. What was the original error you were seeing?

v2/internal/frontend/desktop/linux/window.go
```diff
@@ -609,8 +581,8 @@ func (w *Window) UnFullscreen() {
 }
 
 func (w *Window) Destroy() {
-	C.g_object_unref(C.gpointer(w.gtkWindow))
 	C.gtk_widget_destroy(w.asGTKWidget())
+	C.g_object_unref(C.gpointer(w.gtkWindow))
```
leaanthony, line 585: Nice catch!

v2/internal/frontend/desktop/linux/window.go
```diff
@@ -748,7 +722,11 @@ func (w *Window) ExecJS(js string) {
 }
 
 func (w *Window) StartDrag() {
-	C.StartDrag(w.webview, w.asGTKWindow())
+	data := C.DragOptions{
+		webview:    w.webview,
+		mainwindow: w.gtkWindow,
+	}
+	C.ExecuteOnMainThread(C.startDrag, C.gpointer(&data))
```
leaanthony, line 729: The reason this was being constructed in C was to ensure that the DragOptions were still valid on the main thread. I'm not 100% sure on when Go will free this memory and I'm curious how this fixes the original issue!

leaanthony (commented):
Thanks so much for this Ian! There's a few talking points here and I'll need to do some testing on ARM64 👍

v2/internal/frontend/desktop/linux/window.go
ianmjones: It's a bit weird, but the original error (end of this comment) complains about `gtk_widget_get_toplevel` not being able to assert that the given widget is a `GtkWidget`, usually. Sometimes it complained about a null pointer instead.

I tried all kinds of things to try and verify that the path from the "drag" message coming through `processMessage` had good data and valid types, but still it would fail on click.

In the end I used the simpler setup that is used by `SetRGBA` that seems to have variables with consistent access to the expected webview and mainwindow references.

For me, letting Go control and garbage collect the struct with references to the window and webview makes sense and is simpler for me to understand, but this is not an area I have a lot of experience in.

```
ian@ians-apollo:~/Projects/wails-v2-test $ build/bin/wails-v2-test
Overriding existing handler for signal 10. Set JSC_SIGNAL_FOR_GC if you want WebKit to use a different signal

(wails-v2-test:31285): Gtk-CRITICAL **: 22:20:24.689: gtk_widget_get_toplevel: assertion 'GTK_IS_WIDGET (widget)' failed

(wails-v2-test:31285): Gtk-CRITICAL **: 22:20:24.689: gtk_widget_get_toplevel: assertion 'GTK_IS_WIDGET (widget)' failed
fatal error: unexpected signal during runtime execution
[signal SIGSEGV: segmentation violation code=0x1 addr=0x7f18ae662 pc=0x7f333f293e96]

runtime stack:
runtime.throw({0x6733dc, 0x7f333af668af})
        /nix/store/76qbr3hy9adncz7j149wx9s34s0namyw-go-1.17.6/share/go/src/runtime/panic.go:1198 +0x71
runtime.sigpanic()
        /nix/store/76qbr3hy9adncz7j149wx9s34s0namyw-go-1.17.6/share/go/src/runtime/signal_unix.go:719 +0x396

goroutine 1 [syscall]:
runtime.cgocall(0x5ff1b0, 0xc00019fe80)
        /nix/store/76qbr3hy9adncz7j149wx9s34s0namyw-go-1.17.6/share/go/src/runtime/cgocall.go:156 +0x5c fp=0xc00019fe58 sp=0xc00019fe20 pc=0x40b0dc
github.com/wailsapp/wails/v2/internal/frontend/desktop/linux._Cfunc_gtk_main()
        _cgo_gotypes.go:979 +0x45 fp=0xc00019fe80 sp=0xc00019fe58 pc=0x5ecf65
github.com/wailsapp/wails/v2/internal/frontend/desktop/linux.(*Window).Run(0xc000192230)
        /home/ian/Projects/github.com/wailsapp/wails/v2/internal/frontend/desktop/linux/window.go:718 +0x8c fp=0xc00019fe98 sp=0xc00019fe80 pc=0x5f506c
github.com/wailsapp/wails/v2/internal/frontend/desktop/linux.(*Frontend).Run(0xc0001921c0, {0x6da120, 0xc00007ff80})
        /home/ian/Projects/github.com/wailsapp/wails/v2/internal/frontend/desktop/linux/frontend.go:130 +0xbd fp=0xc00019fed8 sp=0xc00019fe98 pc=0x5ef59d
github.com/wailsapp/wails/v2/internal/appng.(*App).Run(0xc00007c320)
        /home/ian/Projects/github.com/wailsapp/wails/v2/internal/appng/app_production.go:37 +0x31 fp=0xc00019ff10 sp=0xc00019fed8 pc=0x5fced1
github.com/wailsapp/wails/v2.Run(0xc0000001a0)
        /home/ian/Projects/github.com/wailsapp/wails/v2/wails.go:24 +0x33 fp=0xc00019ff28 sp=0xc00019ff10 pc=0x5fd633
main.main()
        /home/ian/Projects/wails-v2-test/main.go:26 +0x3af fp=0xc00019ff80 sp=0xc00019ff28 pc=0x5fdaaf
runtime.main()
        /nix/store/76qbr3hy9adncz7j149wx9s34s0namyw-go-1.17.6/share/go/src/runtime/proc.go:255 +0x227 fp=0xc00019ffe0 sp=0xc00019ff80 pc=0x43d0e7
runtime.goexit()
        /nix/store/76qbr3hy9adncz7j149wx9s34s0namyw-go-1.17.6/share/go/src/runtime/asm_amd64.s:1581 +0x1 fp=0xc00019ffe8 sp=0xc00019ffe0 pc=0x469c81

goroutine 6 [chan receive]:
github.com/wailsapp/wails/v2/internal/frontend/desktop/linux.(*Frontend).startMessageProcessor(0x0)
        /home/ian/Projects/github.com/wailsapp/wails/v2/internal/frontend/desktop/linux/frontend.go:111 +0x59
created by github.com/wailsapp/wails/v2/internal/frontend/desktop/linux.NewFrontend
        /home/ian/Projects/github.com/wailsapp/wails/v2/internal/frontend/desktop/linux/frontend.go:96 +0x367

goroutine 7 [chan receive]:
github.com/wailsapp/wails/v2/internal/frontend/desktop/linux.(*Frontend).startRequestProcessor(0x0)
        /home/ian/Projects/github.com/wailsapp/wails/v2/internal/frontend/desktop/linux/frontend.go:280 +0x5a
created by github.com/wailsapp/wails/v2/internal/frontend/desktop/linux.NewFrontend
        /home/ian/Projects/github.com/wailsapp/wails/v2/internal/frontend/desktop/linux/frontend.go:97 +0x3a9
ian@ians-apollo:~/Projects/wails-v2-test $
```

v2/internal/frontend/desktop/linux/window.go
ianmjones: From tracing through and testing `runtime.WindowSetTitle`from both JS and Go, I couldn't find a reason to hop across to the main thread. This code worked consistently both at startup and runtime.

However, this is not my area of expertise, so if forcing main thread use is a must, especially for ARM64, I'm game to try and reintroduce it.

The errors I'm seeing at startup are...

```
(wails-v2-test:62686): Gtk-CRITICAL **: 22:55:36.314: gtk_window_set_title: assertion 'GTK_IS_WINDOW (window)' failed
fatal error: unexpected signal during runtime execution
[signal SIGSEGV: segmentation violation code=0x1 addr=0xfffffffffffffffb pc=0x7f9f5e73ea95]
```

or

```
fatal error: unexpected signal during runtime execution
[signal SIGBUS: bus error code=0x80 addr=0x0 pc=0x7f2a0dbdd77a]
```

The failed assertion by `gtk_window_set_title` made me think there was a race condition due to `SetTitle` being called before the `gtk_main()` call. Commenting out the `result.SetTitle(appoptions.Title)`call in `NewWindow` stopped the errors on startup. Like with the `StartDrag` fix below, I tried various ways of guarding against the window reference not be what it should be, but found stripping right back seemed to do the job.

ianmjones:
@leaanthony So, I had a play on my Raspberry Pi 400 this morning to test out ARCH64, there's good new and bad news...

The good news is that while I got crashes on startup when using a `master` branch checkout in `go.mod`, using this PR's branch got me a working app...

Screenshot from 2022-03-03 09-44-33

The bad news is that clicking the Wails logo or any other area of the app other than input or button effectively caused a lock-up, and made everything unusable. So this likely means that your fears about the `StartDrag` function's memory management are likely spot on.

I'll switch this PR to draft mode and give it a bit more work.

v2/internal/frontend/desktop/linux/window.go
stffabi: > The reason this was being constructed in C was to ensure that the DragOptions were still valid on the main thread. I'm not 100% sure on when Go will free this memory and I'm curious how this fixes the original issue!

IIRC putting the data on the Go-Stack might be problematic, it's sort of a similar problem which was in the go-webview2 code. CGO only makes sure this data remains there and doesn't get moved as long as the CGO call is running. After the CGO call returns the memory will get reused, especially after the Go method which called CGO has been returned. If the Main-Thread callback is getting called after this has happened, there might be other data stored at that pointer.

stffabi:
Regarding the `startDrag` and `setTitle` panics, I think I might have found the problem. According to the documentation of [g_idle_add](https://docs.gtk.org/glib/func.idle_add.html) which is used for `ExecuteOnMainThread` and the [SourceFunc-ReturnValue](https://docs.gtk.org/glib/callback.SourceFunc.html#return-value) one should return `false` in the callback method to get it removed from the message loop. Otherwise the callback method is called continuously as long as the return value is not `false`. Currently the callbacks `startDrag` and also `setTitle` use void as return type, AFAIK it's not specified what the compiler really returns here if the return value is used as a `bool startDrag()`. So it seems like for ARM64 and or newer versions of the compiler a non zero value is returned, which leads to continuously calling the method. As a result memory is freed multiple times in those callbacks.

Using current master branch with the following changed code for `startDrag` and `setTitle` did fix the panics for me on Fedora34-ARM64.

```c
static bool startDrag(gpointer data)
{
   DragOptions* options = (DragOptions*)data;

   // Ignore non-toplevel widgets
   GtkWidget *window = gtk_widget_get_toplevel(GTK_WIDGET(options->webview));
   if (!GTK_IS_WINDOW(window)) return false;

   gtk_window_begin_move_drag(options->mainwindow, 1, xroot, yroot, dragTime);
   free(data);

   return false;
}

bool setTitle(gpointer data) {
  SetTitleArgs* args = (SetTitleArgs*)data;
  gtk_window_set_title(args->window, args->title);
  free((void*)args->title);
  free((void*)data);
  return false;
}
```

Maybe @ianmjones you could give that a try on your raspberry to see if that also fixes the problem for you? That would be really awesome.

stffabi:
And thanks @ianmjones so much for putting your time into this.

leaanthony:
Really nice catch @stffabi. I did notice that, for some reason, amd64 compiles seemed to operate a lot better with not being on main thread and it made no sense to me. That's why there was effort moving everything over to the main threas. Hadn't seen any issues with ubuntu ARM after this but what you said about compiler defaults makes sense. Just goes to show how important testing on different systems is!

Thank you both for your time on this 🙏

ianmjones:
@stffabi Thanks for the info regarding `g_add_idle`, I implemented your set of test changes and they worked fine on adm64 with Arch Linux and NixOS, and on arm64 with Ubuntu.

So I've gone ahead and applied those changes for all the `ExecuteOnMainThread` callback functions, using the return type and constant the docs you linked to recommended.

I've also fixed up the couple of instances where `ExecuteOnMainThread` was calling gtk functions directly that don't adhere to the expected interface.

While testing that last change with runtime calls to `WindowFullscreen` I came across a bug on arm64 with or without my changes that doesn't manifest on amd64. As it's an existing problem I'll raise that as a separate issue for investigation.

Thank you @leaanthony and @stffabi for your reviews and guidance, very much appreciated.

leaanthony:
Amazing work Ian! Thanks for this. ARM64 is really shaping up!

ianmjones:
> Amazing work Ian! Thanks for this. ARM64 is really shaping up!

You're welcome.

> There's an unrelated issue I commented on but I can add that in after merge if you like

Hmm, can't find the comment you mention, can you point me to it?

v2/internal/frontend/desktop/linux/window.go
```diff
@@ -196,16 +197,18 @@ typedef struct DragOptions {
...
+static gboolean startDrag(gpointer data)
 {
 	DragOptions* options = (DragOptions*)data;
 
-   // Ignore non-toplevel widgets
-   GtkWidget *window = gtk_widget_get_toplevel(GTK_WIDGET(options->webview));
-   if (!GTK_IS_WINDOW(window)) return;
+	// Ignore non-toplevel widgets
+	GtkWidget *window = gtk_widget_get_toplevel(GTK_WIDGET(options->webview));
+	if (!GTK_IS_WINDOW(window)) return G_SOURCE_REMOVE;
```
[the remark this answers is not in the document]
leaanthony: ☝️ was in "pending"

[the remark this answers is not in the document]
ianmjones: I've plugged that leak!

leaanthony:
Amazing PR! Thank you both 🙏

Issue #1210, opened 2022-03-04
Web workers?


ghost:
Since wails uses file:// protocol (i think) using web workers is not possible

is there any way to come around this?

for example the monaco editor needs a web worker to work properly :)

leaanthony:
This is only true on Windows. I would rather it didn't use file but it appeared impossible to use a custom protocol in Webview2 for http

stffabi:
@leaanthony what would be possible is to move the asset server e.g. to "https://wails.localhost" instead of "file://wails/". The callback which we receive from the Webview2 is for all requests so we can also serve requests for https. I've created a draft PR for the purpose of testing, so @Failip420 could give it a try. I've just changed it and the vanilla template worked, I've not further tested it and I've not taken a look into all implications for it.

leaanthony:
Holy moly I didn't expect you to just do it 😀 Thanks! @Failip420 let us know how you get on.
@stffabi is this a better approach for the other platforms too? Are there potential CORS problems with this approach?

Failip420:
Yeah nice :) Just in case you could add an option in wails.json to choose between file:// or https etc 😊

Failip420:
I'm not sure I did download the new with https and compiled it and ran using wails.exe dev with the new binary and it still is using file?

Do you have already compiled version for windows @stffabi ?

<img width="467" alt="image" src="https://user-images.githubusercontent.com/2877483/156858753-8adbf097-e6cd-4437-a978-92f1219aa97a.png">

Failip420:
Yeah no now for sure I did compile the PR @stffabi made and installed it with go install and that, It is still the same error as the screenshot.

leaanthony:
Make sure you're following the guide https://wails.io/docs/guides/bleeding-edge

Failip420:
@leaanthony Ahh yes haha forgot to do that. Well it seems to work just fine now.
No errors or anything so far :) and it seems to work.

leaanthony:
This seems like it works. Will merge and close 👍 Thanks @stffabi!

stffabi:
Awesome thanks both of you to test it 🚀 and thanks @leaanthony for merging it.

strogonoff:
> Make sure you're following the guide https://wails.io/docs/guides/bleeding-edge

FYI, this link is now broken.

leaanthony:
It's now: https://wails.io/docs/guides/local-development

PR #1211, opened 2022-03-04
[v2, windows] Move assets from file://wails to https://wails.localhost

stffabi:
This should allow web workers to be used

The purpose of this draft PR is to see if web workers would work as requested in #1210. I've just changed it and the vanilla template worked, I've not further tested it and I've not taken a look into all implications of this change. So the PR is open for discussions.


leaanthony:
I've checked that you can load remote resources when using this scheme and you can. I think this is good to go!

PR #1214, opened 2022-03-05
refactor(cli): optimize default templates

misitebao:
- Optimize the base template used by `wails generate template`
- Optimize built-in default `vanilla` templates and add scripts
- Refactor `svelte` template with Vite and add script

The template generated by the `base` template keeps the same UI and function as the `vanilla` and `svelte` templates. All updates are only tested on windows.


v2/cmd/wails/internal/commands/initialise/templates/templates/svelte/wails.tmpl.json
```diff
@@ -1,9 +1,12 @@
 {
   "name": "{{.ProjectName}}",
   "outputfilename": "{{.BinaryName}}",
+  "assetdir": "frontend/dist",
```
stffabi, line 4: Wails tries to infer the `assetdir` from the embed.FS out of main.go, which works fine as long as a user doesn't move the embed.FS out of main.go. We have seen several reports during the last few weeks where people had a working `wails dev` and `wails build` didn't, which was caused by having only adapted the `assetdir` and not changing the embed.FS.
Therefore I would suggest to remove the `assetdir` entry here, to avoid those inconsistencies.

v2/cmd/wails/internal/commands/initialise/templates/templates/vanilla/wails.tmpl.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "{{.ProjectName}}",
   "outputfilename": "{{.BinaryName}}",
+  "assetdir": "frontend/src",
```
stffabi, line 4: Same applies here.

stffabi (changes requested):
Awesome @misitebao, thank for all the hard work you've put into this.

I'm not an frontend expert, so I have only left a comment regarding the `assetdir` entry in wails.json.

v2/cmd/wails/internal/commands/initialise/templates/templates/svelte/wails.tmpl.json
misitebao: It means that the value of this field will be automatically obtained from the Go code without the need to fill in manually, right?

stffabi: Yeah, exactly.

v2/cmd/wails/internal/commands/generate/template/base/frontend/dist/index.html
```diff
@@ -1,18 +1,20 @@
...
-    <input class="input" id="name" type="text" autocomplete="off">
-    <button class="btn" onclick="greet()">Greet</button>
-  </div>
-
-  <script src="/main.js"></script>
-</body>
-
-</html>
\ No newline at end of file
+  <body data-wails-drag>
```
leaanthony, line 9: We should remove all the drag attributes from the default template. People have reported issues with click events not realising this attribute will swallow them.

v2/cmd/wails/internal/commands/generate/template/base/frontend/dist/index.html
```diff
@@ -1,18 +1,20 @@
...
+    <div id="app" class="app">
+      <div class="logo"></div>
+      <div class="result" id="result">Please enter your name below 👇</div>
+      <div class="input-box" id="input" data-wails-no-drag>
```
leaanthony, line 13: Let's remove no drag too 👍

leaanthony (changes requested):
Let's take out the drag attributes.

v2/cmd/wails/internal/commands/generate/template/base/frontend/dist/index.html
misitebao: If this property is removed, how can the user drag the window in borderless mode?

leaanthony: By adding it back in? 😄

v2/cmd/wails/internal/commands/initialise/templates/templates/svelte/frontend/README.md
```diff
@@ -1,105 +1,48 @@
...
+# Svelte + Vite
 
----
+This template should help get you started developing with Svelte in Vite.
 
-# svelte app
+## Recommended IDE Setup
 
-This is a project template for [Svelte](https://svelte.dev) apps. It lives at https://github.com/sveltejs/template.
+[VSCode](https://code.visualstudio.com/) + [Svelte](https://marketplace.visualstudio.com/items?itemName=svelte.svelte-vscode).
```
leaanthony, line 7:
```suggestion
  - [Goland](https://www.jetbrains.com/go/) + [Svelte](https://plugins.jetbrains.com/plugin/12375-svelte)
  - [VSCode](https://code.visualstudio.com/) + [Svelte](https://marketplace.visualstudio.com/items?itemName=svelte.svelte-vscode).
```

v2/cmd/wails/internal/commands/initialise/templates/templates/svelte/frontend/package-lock.tmpl.json
```diff
@@ -1,3638 +1,1016 @@
 {
   "name": "{{.ProjectName}}",
-  "version": "1.0.0",
-  "lockfileVersion": 1,
+  "version": "0.0.0",
```
leaanthony, line 3: I'm not sure we need this file? People might not trust it.

misitebao: This should be decided by the user, and some projects will use the front-end part as a sub-module (for example: https://github.com/letheanVPN/desktop). Front-end developers and back-end developers maintain it separately, I think it exists If necessary, I usually modify it myself to add version tags.

v2/cmd/wails/internal/commands/initialise/templates/templates/svelte/frontend/README.md
misitebao: Sorry, I forgot it, I just overwritten it, do I need to go back to the previous version now?

PR #1215, opened 2022-03-05
feat(website): sync documents and add content

misitebao:
- Add a directory to the website
- Update site navigation menu
- Update site tagline(Emphasis on cross-platform)
- Sync the changelog of the latest version to the website


misitebao:
@leaanthony 
Do you see if there is anything that needs to be modified or added?

leaanthony:
This is great - thanks. I'm wondering if we need appendix right now? Maybe add it when we have content? What do you think?

misitebao:
@leaanthony 
The `appendix` directory has been temporarily removed, it will not show up in the documentation.

leaanthony:
LGTM

PR #1229, opened 2022-03-08
Feature/1175 default index page

leaanthony:
Retry loading `index.html` every 500ms for 5s. Default to inbuilt template after this.


misitebao:
1. Build the latest `CLI` based on the current branch
2. Execute `wails init -n wails-svelte -t svelte`
3. Modify `go.mod` of `wails-svelte`
4. Execute `wails dev`
5. Try to modify the front-end source code, the program will automatically refresh, until the end, everything is normal.
6. **Manually delete the `index.html` file, the window content will not change**.
7. Exit the program and repeat step 5
8. Manually modify the name of the `index.html` file. **At this time, the window will be stuck, and the function of dragging the window will fail**. After a while, the window will display the default page. At this point, the drag function returns to normal.

leaanthony:
I'm ok with this. If someone wants to delete index.html and start dragging the window around then I guess there's a price to pay 😀 the other 99.9% use case is covered. We could probably work around the lock by using a goroutine, but I'm ok not doing that for now too

misitebao:
@leaanthony 
It can be merged into the master branch for now, and it can be optimized later, because no one will deliberately delete files manually.

stffabi:
I've just found a little problem since this PR. Using a vanilla project production build run on Windows, it seems like it tries to load a favicon.ico which is not there. Therefore since this PR, reloading comes in and tries to load it for the next 5 seconds and fully blocks the application.
Because it's a production build, there's never a chance that this file will exist. Shouldn't we restrict this reloading only to `dev` builds?

leaanthony:
Yeah - I was a bit tired last night and thought that's what it was :-) Absolutely that's what it should be.

leaanthony:
@stffabi how about this?
https://github.com/wailsapp/wails/pull/1241

Issue #945, opened 2021-11-16
Models  are not generated in Models.ts for output types


felixbecker:
#####################################################
     **If you have a technical issue, please do not open a bug this way!** 
                      Please use the `wails issue` command!
          If you do not do this then the issue may be closed automatically.                      

 NOTE: If your bug is related to Windows, make sure you read 
 the [Windows Developer Guide](https://wails.app/guides/windows/)
#####################################################

**Description**
As discussed on the discussion board here: https://github.com/wailsapp/wails/discussions/828#discussioncomment-1643354 
it looks like the models are not generated for output types. 
Also if one access a list of objects via JavaScript it just returns plain objects. Methods on the JavaScript Objects will not be returned. Also shown in the discussion above.

**To Reproduce**
Steps to reproduce the behaviour:
https://github.com/wailsapp/wails/discussions/828#discussioncomment-1643354

**Expected behaviour**
Models should be generated. And Methods on the Models should also be available by JavaScript.

**Screenshots**
[If applicable, add screenshots to help explain your problem.](https://github.com/wailsapp/wails/discussions/828#discussioncomment-1643354)

**System Details**
Please provide your platform, GO version and variables, etc

**Additional context**

- [ ] This issue is for Windows and I have read the [Windows Developer Guide](https://wails.app/guides/windows/)

leaanthony:
Getting the models generated isn't too hard but getting the Go methods in Javascript will be a lot harder. What would that even look like? Do you have an example?

leaanthony:
@felixbecker I've pushed an update to master that supports return params, but also slices of structs too :-)

felixbecker:
Hi @leaanthony: thanks for your quick response. 
Unfortunately it is not generating the models for me.
Here is what I did. 

- I cloned the current master
- cd' in the v2 folder and `go build cmd/wails/main.go``
- with this binary I created the following project https://github.com/felixbecker/wailsmodelgenerationtest
- I added a `dataservice.go` file which contains a a method that returns all todo types 

Model.ts is still empty. In case I do something wrong ... let me know.

leaanthony:
Make sure your project's `go.mod` has the replace line pointing to the code you checked out 👍

leaanthony:
@felixbecker - how did you get on?

felixbecker:
I am not able to get it running. I uncommented the replace statement in the `go.mod`-file.
When I build or try to run it ... I get the following error:
```
Installing frontend dependencies: Done.
[...]

PR #1231, opened 2022-03-08
import models on binding definition

adalessa:
From the conversation on the issue #945 the generation of the file more specific here https://github.com/wailsapp/wails/issues/945#issuecomment-999378082 to avoid errors on the bindings definitions is necessary to reference the models. In order to import all the generated models it is necessary to us the `as` and give a namespace and of course it is necessary to add it to the use of the types.


leaanthony:
Thanks for doing this! LGTM!

PR #1238, opened 2022-03-10
Build added win32 compatibility

fengweiqiang:
need to depend on https://github.com/leaanthony/go-webview2/pull/8#issue-1164702663


leaanthony (approved):
Thanks for looking at this. I'm not sure we'll officially support 32bit but we can unofficially have it work 👍

PR #1240, opened 2022-03-10
ensure binary resources can be served

### LINKED ISSUE #1235: [v2] Linux - processRequest can't deal with binary resources properly
**Description**
Binary resources from embed.FS on Linux (such as images, fonts or pdfs) or any resource with embedded nulls can get truncated resulting in errors or serving incomplete resources. 

**To Reproduce**

1. Add a binary resource (such as a PDF file) to your 'assets' directory
2. build your app with 'wails build'
3. Retrieve the file from Javascript (for instance using pdfjsLib.getDocument from pdf.js)
4. An exception is produced because file is truncated

**Expected behaviour**
Resources with embedded binary nulls should be served correctly and not result in content truncation.

**Additional context**
The problem occurs in the `processRequest()` function in `v2/internal/frontend/desktop/linux/frontend.go` in the following snippet:

```go
        cContent := C.CString(string(res.Body))
	defer C.free(unsafe.Pointer(cContent))
	cMimeType := C.CString(res.MimeType)
	defer C.free(unsafe.Pointer(cMimeType))
	cLen := C.long(C.strlen(cContent))
	stream := C.g_memory_input_stream_new_from_data(
		unsafe.Pointer(C.g_strdup(cContent)),
		cLen,
		(*[0]byte)(C.g_free))
	C.webkit_uri_scheme_request_finish(req, stream, cLen, cMimeType)
	C.g_object_unref(C.gpointer(stream))
```

Apart from the fact that a lot of unnecessary copying seems to be to done in `C.CString(string(res.Body))` and then later `C.g_strdup`, the code uses `C.strlen(cContent)` which will return the length of `cContent` up to the first embedded binary null character. This works fine for resources without embedded binary nulls but not for arbitrary content. `g_strdup` suffers from the same problem. The fix should be pretty simple - I'll add a pull request with a proposed fix.

napalu:
fixes #1235

### DIFF
--- a/v2/internal/frontend/desktop/linux/frontend.go
+++ b/v2/internal/frontend/desktop/linux/frontend.go
@@ -306,13 +306,19 @@ func (f *Frontend) processRequest(request unsafe.Pointer) {
 		return
 	}
 
-	cContent := C.CString(string(res.Body))
-	defer C.free(unsafe.Pointer(cContent))
+	var cContent unsafe.Pointer
+	bodyLen := len(res.Body)
+	if bodyLen > 0 {
+		cContent = C.malloc(C.ulong(bodyLen))
+		C.memcpy(cContent, unsafe.Pointer(&res.Body[0]), C.size_t(bodyLen))
+	}
+
 	cMimeType := C.CString(res.MimeType)
 	defer C.free(unsafe.Pointer(cMimeType))
-	cLen := C.long(C.strlen(cContent))
+
+	cLen := C.long(len(res.Body))
 	stream := C.g_memory_input_stream_new_from_data(
-		unsafe.Pointer(C.g_strdup(cContent)),
+		cContent,
 		cLen,
 		(*[0]byte)(C.g_free))
 	C.webkit_uri_scheme_request_finish(req, stream, cLen, cMimeType)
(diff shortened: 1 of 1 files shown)


v2/internal/frontend/desktop/linux/frontend.go
```diff
@@ -306,13 +306,19 @@ func (f *Frontend) processRequest(request unsafe.Pointer) {
...
 	cMimeType := C.CString(res.MimeType)
 	defer C.free(unsafe.Pointer(cMimeType))
-	cLen := C.long(C.strlen(cContent))
+
+	cLen := C.long(len(res.Body))
 	stream := C.g_memory_input_stream_new_from_data(
-		unsafe.Pointer(C.g_strdup(cContent)),
+		cContent,
 		cLen,
 		(*[0]byte)(C.g_free))
```
leaanthony, line 323: Am wondering if this free might blow up when the body length is 0?

leaanthony:
Thanks for finding this bug and submitting the PR. Just one comment but looks good 👍

v2/internal/frontend/desktop/linux/frontend.go
stffabi: Since cContent is now allocated with `C.malloc` wouldn't it be more consistent to use `C.free` here instead of `C.g_free`?

leaanthony:
Awesome stuff! Let's go!

PR #1241, opened 2022-03-10
Only retry loading assets when loading from disk

leaanthony:
Not tested!


stffabi (approved):
LGTM, just tested on Windows. Works as expected

stffabi:
Thanks a lot @leaanthony

PR #1245, opened 2022-03-10
Ensure Linux IsFullScreen uses GDK_WINDOW_STATE_FULLSCREEN bitmask appropriately.

### LINKED ISSUE #1244: [v2,Linux] Calling runtime.WindowUnfullscreen from a button does not work
**Description**

On Linux, when the JavaScript `runtime.WindowUnfullscreen()` function is used via a button or any other trigger on the **active** window, nothing happens.

**Attempted Fixes**

I have a fix for this incoming! :smile:

ianmjones:
Resolves #1244 

The GDK_WINDOW_STATE_FULLSCREEN bitmask was not being used properly in `C.IsFullScreen`, and because other states such as GDK_WINDOW_STATE_FOCUSED could also be in play, `Window.IsFullScreen` needs to check for non-zero rather than one as the `C.IsFullScreen` result.


leaanthony:
Amazing! Thanks! 🙏🙏

Issue #1232, opened 2022-03-08
V2 - watcher keep running after closing app


BillBuilt:
I've just notice some strange behavior. So when I use `wails dev` to launch the app using the new `frontend:dev:watcher` config option, if I close the app the normal way, or use `CTRL-C` to kill it, a node watcher stays running in the background. I was able to see it in activity monitor (Mac OS) with `node` as the process name. When launching `wails dev` this way I see 2 `node` processes spin up, both showing CPU usage. After the building is done and the app launches, 1 remains running and I believe this is the one causing this behavior. At this point if I close the app, this `node` process remains active. Running `npm run dev` manually only spawns 1 `node` instance and it closes properly using `CTRL-C`.

Im using Vite and `npm run dev` = `vite build --watch`

_Originally posted by @BillBuilt in https://github.com/wailsapp/wails/issues/1132#issuecomment-1062198620_

leaanthony:
Hey @BillBuilt - is this still an issue now?

BillBuilt:
Hey @leaanthony, sorry, I guess I thought this would be closed with the merge. But, so far so good on my end so I will close this now. Thanks for the reminder.

PR #1247, opened 2022-03-11
Fix, or at least a workaround, for (#1232)

BillBuilt:
My first ever open source contribution. I hope I did it correctly ;-) There may be a more elegant solution but this is what ended up working under most circumstances ( that I was able to try out )


leaanthony:
Thanks Bill! Will take a look this weekend 🙏

Issue #1246, opened 2022-03-11
[v2] Linux - OnDomReady is not implemented


napalu:
**Description**
OnDomReady is not implemented on linux.

**Expected behaviour**
OnDomReady should be called after the frontend dom has loaded

PR #1249, opened 2022-03-12
add support for 'DomReady' callback on linux

napalu:
cosmetic: use bodyLen instead of len() in processRequest()

### DIFF
--- a/v2/internal/frontend/desktop/linux/frontend.go
+++ b/v2/internal/frontend/desktop/linux/frontend.go
@@ -226,6 +226,13 @@ func (f *Frontend) Notify(name string, data ...interface{}) {
 }
 
 func (f *Frontend) processMessage(message string) {
+	if message == "DomReady" {
+		if f.frontendOptions.OnDomReady != nil {
+			f.frontendOptions.OnDomReady(f.ctx)
+		}
+		return
+	}
+
 	if message == "drag" {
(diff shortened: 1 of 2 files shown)


napalu:
A small nitpick: the term "DomReady" is in my opinion misleading because it suggests and equivalence to javascript's 'DOMContentLoaded' whereas 'DomReady' is emitted when navigation has finished loading. Maybe the description and name of the go callback should be changed to reflect this?

napalu:
fixes #1246

leaanthony:
> A small nitpick: the term "DomReady" is in my opinion misleading because it suggests an equivalence to javascript's 'DOMContentLoaded' whereas 'DomReady' is emitted when navigation has finished loading. Maybe the description and name of the go callback should be changed to reflect this?

That's a valid point. Originally it was to indicate that elements of the page were ready, including the runtime but I agree the name is now misleading. What would you call it?

napalu:
Something like `resourcesLoaded` might be more appropriate - currently the  comment to describe `domReady` after `wails init` is run reads: `domReady is called after the front-end dom has been loaded`.

I suppose we could leave the name as-is and just alter the comment to read `domReady` is called after front-end resources have been loaded` - what do you think?

leaanthony:
Updated comments in templates + website. Thanks for all your contributions 🙏

PR #1252, opened 2022-03-13
Warn when running wails outside of GOPATH

### LINKED ISSUE #1148: [v2] Add short warning when wails binary is not in PATH

When I started working with Wails, the installed wails binary was not in the PATH. I started wails with its absolute path, as I do with other Go applications as well.

The Wails documentation is clear about that. I just missed that part. But without having Wails in PATH, it will exit without any error. Even verbose mode doesn't print information. It finishes after "wails generate module"

The documentation is clear and correct. It would be nice to get an additional error message, that Wails binary was not found.

**Additional context**

image

leaanthony:
Fixes #1148


leaanthony:
Tested on Windows

Issue #1253, opened 2022-03-13
[v2, windows] Add flag to indicate whether to use https://


stefpap:
**Is your feature request related to a problem? Please describe.**
The newly merged #1211 PR broke my app because with I cannot load http content anymore, since the current page is considered to be `https://wails.localhost/`
![image](https://user-images.githubusercontent.com/22637722/158064306-ece71344-0b62-4a0a-a6b6-ada301cec6dc.png)

**Describe the solution you'd like**
Do you think it would be possible to add a parameter to indicate `http`, `https` or `file://`, or just a bool flag something like isHTTP ?

stffabi:
Yeah, that's a problem. Maybe we should switch to http. Because in the near future as soon as Webview2 supports custom schemes there's the plan to use wails:// to have it standardised on all platforms. MacOS and Linux already use that custom scheme.
Especially https makes it now inconsistent, because it works on MacOS and Linux but not on Windows.
So I would propose to switch to http.

stefpap:
Would switching to http break the web workers? If not, that would be an easy change

stffabi:
AFAIK web workers should still work, but we would need to check it.

stefpap:
Created a draft PR for this #1255

leaanthony:
@failip420 any chance you could check if this PR breaks your web workers app?

Failip420:
> @Failip420 any chance you could check if this PR breaks your web workers app?

It still seems to work.
No errors or anything so far :)

leaanthony:
That would need testing across all platforms. I'm ok with this PR in the short term and them switch to custom protocol once it's merged in webviews. I believe that'll be in a month

stffabi:
Yeah absolutely my intention was not to use http:// on all platforms, only just for windows until we can use wails://.
With a consistent behaviour I meant that all platforms allow accessing unsecure content.

stefpap:
I'll close the issue since the PR got merged. Thanks everyone for helping out!

PR #1255, opened 2022-03-13
[v2, windows] Windows switch scheme: https -> http

stefpap:
This PR addresses issue #1253 .

TLDR: Basically switch to http to allow requests for insecure content


stffabi:
Awesome, thanks for the PR 🚀. We just need to make sure WebWorkers still work, then we are good to go. 👍

PR #1257, opened 2022-03-14
Ensure Focus is regained by Webview2 when tabbing

### LINKED ISSUE #1254: [V2,Window] Can't auto focus to Web window/document after switch window.
**Description**
When use "Alt-Tab" to switch window, Wails don't auto focus to web content (window/document) after switch back to app.
```js
window.addEventListener("focus",...);
document.addEventListener("focus",...);
```
Both of them not triggered.

**To Reproduce**

1. Open App
2. Alt-Tab switch to another window.
3. Alt-Tab back to App
4. Don't focus to App until mouse click to content

**Expected behaviour**
I thought `window` will be focus when anytime switch to app.

**System Details**
image

leaanthony:
Fixes #1254 
Tested focusing input, alt-tab, alt-tab = focus back
Also tested focus event in JS.

### DIFF
--- a/v2/go.mod
+++ b/v2/go.mod
@@ -22,7 +22,7 @@ require (
 	github.com/leaanthony/debme v1.2.1
 	github.com/leaanthony/go-ansi-parser v1.0.1
 	github.com/leaanthony/go-common-file-dialog v1.0.3
-	github.com/leaanthony/go-webview2 v1.0.3-0.20220125181555-00f20a61c17d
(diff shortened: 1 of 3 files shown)


RensTillmann:
Was wondering about this "fix", because for me `ALT+TAB` doesn't regain focus of the window. Only losing focus works if it was already focussed. Any thoughts? There is another open issue regarding this same problem here: #3783

PR #1272, opened 2022-03-20
v2: linux: add icon to linux specific app option and load it during w…

bearsh:
…indow creation

Signed-off-by: Martin Gysel <me@bearsh.org>

### DIFF
--- a/v2/internal/frontend/desktop/linux/window.go
+++ b/v2/internal/frontend/desktop/linux/window.go
@@ -549,6 +549,19 @@ void DisableContextMenu(void* webview) {
 	g_signal_connect(WEBKIT_WEB_VIEW(webview), "context-menu", G_CALLBACK(disableContextMenu), NULL);
 }
 
+void SetWindowIcon(GtkWindow* window, const guchar* buf, gsize len) {
+	GdkPixbufLoader* loader = gdk_pixbuf_loader_new();
+	if (!loader) {
+		return;
+	}
+	if (gdk_pixbuf_loader_write(loader, buf, len, NULL) && gdk_pixbuf_loader_close(loader, NULL)) {
+		GdkPixbuf* pixbuf = gdk_pixbuf_loader_get_pixbuf(loader);
+		if (pixbuf) {
+			gtk_window_set_icon(window, pixbuf);
+		}
+	}
+	g_object_unref(loader);
+}
 
 */
 import "C"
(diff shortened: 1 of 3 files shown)


v2/internal/frontend/desktop/linux/window.go
```diff
@@ -549,6 +549,20 @@ void DisableContextMenu(void* webview) {
...
+	}
+	if (!gdk_pixbuf_loader_write(loader, buf, len, NULL) && !gdk_pixbuf_loader_close(loader, NULL)) {
+		return;
+	}
+	GdkPixbuf* pixbuf = gdk_pixbuf_loader_get_pixbuf(loader);
+	if (!pixbuf) {
+		return;
+	}
+	gtk_window_set_icon(window, pixbuf);
+}
```
bearsh, line 565: freeing the loader cause a 'double free or corruption' error and a SIGABRT, still for me it seems logical to free it after usage... anyway even if it leaks some memory, onetime...

leaanthony: Yeah that's weird :-)

bearsh: do you see a problem with that?

bearsh: thanks for the hint

v2/internal/frontend/desktop/linux/window.go
```diff
@@ -635,6 +648,7 @@ func NewWindow(appoptions *options.App, debug bool) *Window {
 	result.SetTitle(appoptions.Title)
 	result.SetMinSize(appoptions.MinWidth, appoptions.MinHeight)
 	result.SetMaxSize(appoptions.MaxWidth, appoptions.MaxHeight)
+	result.SetWindowIcon(appoptions.Linux.Icon)
```
leaanthony, line 651:
```suggestion
	if appoptions.Linux != nil {
		if appoptions.Linux.Icon != nil {
			result.SetWindowIcon(appoptions.Linux.Icon)
		}
	}
```

leaanthony:
This doesn't appear to work using Ubuntu 2021.10. It uses a default icon for both taskbar and window list:
image
image

What systems did you test it on?

v2/internal/frontend/desktop/linux/window.go
bearsh: is the second comparison really needed as the length of the slice gets checked in the SetWindowIcon() method?

bearsh:
> What systems did you test it on?

KDE neon (Ubuntu 20.04)

v2/internal/frontend/desktop/linux/window.go
leaanthony: Yeah, this will throw a nil error for apps that don't have a linux specific config but run on linux. It was the first thing that happened when testing this PR :-)

leaanthony:
Could you please update the docs to add this Linux specific config? I think we need to say that it's just a hint and not all window managers support it, EG Wayland.

v2/internal/frontend/desktop/linux/window.go
bearsh: sure about the first `if`, I mean the second one: `if appoptions.Linux.Icon != nil {` as this is already tested in `SetWindowIcon`

bearsh:
will do so, as soon as time permits

bearsh:
doc added

Issue #1149, opened 2022-02-14
[v2] Windows: Title bar in color of system theme (dark, light)


TheFranconianCoder:
**Is your feature request related to a problem? Please describe.**
I use Windows with dark system theme. The title bar of an Wails applications stays light.

**Describe the solution you'd like**
On dark theme, the title bar should be dark as well. Ideally that switch happens automatically, even when the application runs.

stffabi:
This seems to involve quite a bit of WinAPI and using undocumented DWM flags and APIs. It also seems to even depend on the Windows version running. I was unable to find any documentation from MS itself how to do this.

Here's the electron implementation [dark_mode.cc](https://github.com/electron/electron/blob/main/shell/browser/win/dark_mode.cc)

Seems MS did something similar in their new terminal https://github.com/microsoft/terminal/commit/bc7eb9611030aed3204aff4e662c318cbf9143a6#diff-e26a93b2aa9fea92ebf24336c4fe6412L19-L22 before they switched to a function named `Microsoft::Console::Internal::Theming::TrySetDarkMode(hWnd)`. But I was unable to find the implementation of that, they mention [here](https://github.com/microsoft/terminal/blob/10824ccf7494cc9befac7ab06db02e1dd3258775/src/inc/conint.h#L10-L14)

>    This module defines an abstraction for calling certain OS support
    functionality that we can't make completely public due to obligations
    we have with our partner teams about disclosure of their API surface.

sebastian0x62:
I found one proof of concept for different windows versions from 2020: https://github.com/ysc3839/win32-darkmode

Then I checked TAO which is used with Tauri. Funny enough, they reference the repository as well and made their own improvements: https://github.com/tauri-apps/tao/blob/dev/src/platform_impl/windows/dark_mode.rs

stffabi:
Seems like WebView2 doesn't have DarkMode compatibility as of now. So e.g. scrollbars in the webview will remain in light theme.

https://github.com/MicrosoftEdge/WebView2Feedback/issues/1689

sebastian0x62:
That would of course be consistent, but independent to the title bar issue.

To be honest, I use almost always custom styled scrollbars. So I didn't realize with Tauri.

leaanthony:
For Mac, this is easy. For Windows it would seem not so much. I'm ok with supporting and media queries can deal with scrollbars. I'm not even sure it's possible with linux

stffabi:
[...]

PR #1291, opened 2022-03-28
[Feature/1149] Dark mode:  functions for manually switching theme

sebastian0x62:


sebastian0x62:
Proposal for #1149. Changing theme at runtime.

leaanthony:
Thanks for opening this. I'm open to this change, but we need to do a few more things for it to be complete:

  - There needs to be stubs in the Linux and Mac frontends or else it won't compile any more
  - We need to add the methods to the JS runtime
  - The website documentation needs updating with the new methods

Thoughts?

sebastian0x62:
Sure, when the functions work for you, I will check your mentioned items.

leaanthony:
They work for me 👍

sebastian0x62:
@leaanthony  I'm currently checking the changes. Two things:

- Is there any preferred way for such stubs for other systems, which are not supported for specific functions?
- I used `runtime.LockOSThread()` like in other functions in this context. I'm wondering if this is needed, when we could have one call in an `init()`. But really not sure about that.

stffabi:
> - I used `runtime.LockOSThread()` like in other functions in this context. I'm wondering if this is needed, when we could have one call in an `init()`. But really not sure about that.

Those are needed especially when the methods get called from a JS call from the frontend, which are executed in a separate goroutine. But we should release the OSThread after the call has been finished. IIRC otherwise we create new OSThreads which might not be needed. I have a PR lying around for that...

leaanthony:
The LockOSThreads are there to make sure the method runs on the main thread. I've been meaning to check if it's necessary. Try without and see what happens 😀

Stubs should just return if it isn't supported.

Thanks! 🙏

stffabi:
LockOSThreads might not be needed in all cases, if e.g. BeginInvoke is used on Windows to switch to the GUI Thread they aren't needed. Because the callback makes sure the OsThread won't switch.

sebastian0x62:
Tried to find all parts for JS API. Added some documentation. Had to change the calls on Go side, or it wouldn't change when triggered via JS and window is inactive.

leaanthony:
Looks good! We can deal with LockOSThreads in a different PR 👍

leaanthony:
Something has gone very wrong with this pr

sebastian0x62:
> Something has gone very wrong with this pr

Merged master into my branch because of conflicts to resolve them. :( My branch is still working.

leaanthony:
Weird, I don't know if it was a github fail but I saw 280 files on the PR 😮‍💨

leaanthony:
Nice job 👍

Issue #1292, opened 2022-03-28
Information/docs for making installers?


gardc:
**Is your feature request related to a problem? Please describe.**
I noticed there is support for making and packaging installers for macOS and Windows for Wails, but cannot find any information about it. How can we make a macOS install dmg and a installer for Windows with Wails?
I tried running `wails build -nsis` on Windows with NSIS installed, but received an error that an `makensis` command is not found. Is it possible this is written on and for Linux/macOS only?

**Describe the solution you'd like**
I'd love some documentation on how to build installers for Windows and macOS with Wails on wails.io. Any information commented on this issue would also be greatly appreciated :).

stffabi:
Thanks for creating this issue.

At the moment Wails only supports creating Windows installers, creating them is supported on all platforms. Could you please paste the output of `wails doctor`? I suspect the nsis installation couldn't be found, is `makensis` included in your `PATH`?

Regarding docs there's missing the content about installers. Any help on that docs topic would be awesome :)

stffabi:
@gardc Great that you got it working.

> I can help write docs for this and code signing after I've implemented it for my app!

That would be absolutely awesome, thanks so much.

gardc:
Added docs for both NSIS installer and code signing on GitHub Actions in PR #1301 
Feel free to check that the documentation looks good! :)

leaanthony:
Nice work! Closing 👍

masontmorris:
Hi, I've done what OP did and added the NSIS install directory to my `PATH`, along with the /Bin folder after that didn't work, and I'm still running into the same issue. I've made sure to restart my shell with every attempt and have tried both Bash and PowerShell. Any help would be appreciated!

aalekh12:
can yous 

> Hi, I've done what OP did and added the NSIS install directory to my `PATH`, along with the /Bin folder after that didn't work, and I'm still running into the same issue. I've made sure to restart my shell with every attempt and have tried both Bash and PowerShell. Any help would be appreciated!
> 
> Update: `wails build -nsis` works for me now in a separate PowerShell window! I had been trying in VS Code's integrated terminal, not sure why that didn't work but this is a good enough solution for me!

can you please share the screenshot how you have set your path for nsis

aalekh12:
because i am still facing issue getting same error makensis not found

PR #1301, opened 2022-03-30
Added docs for code-signing and NSIS installer

gardc:
Hopefully this includes useful documentation for code signing Wails binaries on GitHub Actions and NSIS installers.
Closes issue #1292


leaanthony:
Thanks so much for spending the time to write this up. Docs can be so time consuming so this is greatly appreciated!

PR #1303, opened 2022-03-31
cleanup Version

tmclane:
Looks like the an invalid version string got merged to master.
This just removes it.


leaanthony:
Nice one Travis! 👍

PR #1307, opened 2022-04-03
Do not duplicate imports

adalessa:
With the last change for the typescript generation, when you have methods returning the same structure it duplicates the imports. failing the typescript complilation. Here I collect them and write them after removing the duplicated ones.


leaanthony (approved):

leaanthony:
Thanks for this!

PR #1308, opened 2022-04-03
Add missing method declaration

adalessa:
I found when trying to use the documented method `EventsOff` that I was having an error due that is not declare. I have tested adding it here and works for me. Since the file does not contains the header saying is generated I think is ok. But if it is not please let me know how is the correct way to handle it.

Thanks.

### DIFF
--- a/v2/internal/frontend/runtime/wrapper/events.js
+++ b/v2/internal/frontend/runtime/wrapper/events.js
@@ -35,6 +35,16 @@ export function EventsOn(eventName, callback) {
 }
 
 /**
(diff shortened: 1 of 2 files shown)


leaanthony:
Nice catch! Just noticed it's missing from the runtime wrapper too. https://github.com/wailsapp/wails/blob/master/v2/internal/frontend/runtime/wrapper/events.js

adalessa:
I have added it on the wrapper as well.

leaanthony:
Thanks for this 🙏

PR #1311, opened 2022-04-04
feat(website): convert documents that don't require version management to individual pages

misitebao:
Removed documents that do not need version control, added documents and navigation.

Preview: https://wails-lqiqmov4m-misitebao.vercel.app/


leaanthony:
Awesome! I wasn't sure where the extra docs were and am wondering if we should have a static link in the sidebar for them? Thoughts?

misitebao:
@leaanthony 
I had this idea. Yesterday I tried to keep the menu in the sidebar and link to the new page when I removed the document, but I can't add custom tags when using auto-generated, I'm testing adding the sidebar manually .

PR #1314, opened 2022-04-05
feat(website): add page links to the sidebar

misitebao:
Add page to sidebar.

Note: Subsequent addition or deletion of page sidebars will not be automatically generated, and the sidebar configuration needs to be manually updated.

Preview: https://wails-8xbo2i8bh-misitebao.vercel.app/


leaanthony:
Manual updates aren't great. I'll just forget 😂 do you think it matters that these pages were being duplicated each release?

leaanthony:
After looking at the 2 options. I think it's ok to keep the sidebar autogenerated without the links and have the links in the About section as it currently is. Having the sidebar links disappear feels a bit odd after clicking them. At least clicking in "about" it is expected to be different. Thoughts?

misitebao:
Because it's a "page" and not a "document", only documents will have sidebars.

PR #1322, opened 2022-04-09
[v2, windows] Suppress resizing during minimize of a frameless window

### LINKED ISSUE #1319: [V2] Issue Minimizing and Restoring with Frameless Window
### Description

When using frameless window, minimizing and restoring take more time for the contents in window to reload. See gifs below:

### To Reproduce

N/A

### Expected behaviour

Should be the same on both cases (Frameless and No Frameless)

### Screenshots

With ``Frameless`` (Noted the delay?)
test1
Without ``Frameless``:
test2

### Attempted Fixes

_No response_

### System Details

```shell

System
------
OS:             Windows 10 Enterprise LTSC 2019
Version:        1809 (Build: 17763)
ID:
Go Version:     go1.17.6
Platform:       windows
Architecture:   amd64

WebView2        N/A             Installed       100.0.1185.29
npm             N/A             Installed       6.14.13
*upx            N/A             Installed       upx 3.96
*nsis           N/A             Available

* - Optional Dependency

Diagnosis
---------
```

### Additional context

_No response_

stffabi:
If the window is frameless and we are minimizing, then we need to suppress
the Resize on the WebView2. If we don't do this, restoring does not work as
expected and first restores with some wrong size during the restore animation
and only fully renders when the animation is done.
This highly depends on the content in the WebView.
Fixes #1319

### DIFF
--- a/v2/internal/frontend/desktop/windows/frontend.go
+++ b/v2/internal/frontend/desktop/windows/frontend.go
@@ -143,6 +143,17 @@ func (f *Frontend) Run(ctx context.Context) error {
 	f.mainWindow.notifyParentWindowPositionChanged = f.chromium.NotifyParentWindowPositionChanged
 
 	mainWindow.OnSize().Bind(func(arg *winc.Event) {
+		if f.frontendOptions.Frameless {
+			// If the window is frameless and we are minimizing, then we need to suppress the Resize on the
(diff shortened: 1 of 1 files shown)


stffabi:
@leaanthony I think we are good to go. Seems like this solves the issue

PR #1325, opened 2022-04-11
[v2] Add support for AssetsHandler

stffabi:
AssetsHandler is a http.Handler delegate, which gets called
as a fallback for all Non-GET requests and for GET requests
for which the Assets didn’t find the file.

Known Limitations on Linux:
- All requests are GET requests
- No request headers
- No request body
- No response status code, only StatusOK will be returned
- No response headers

Known Limitations on Windows:
-  Request body is leaking memory. Seems to be a bug in
    WebView2, investigation ongoing.

Most of these limitations on Linux will be fixed in the future with
adding support for Webkit2Gtk 2.36.0+.

NOTE: When used in combination with a Frontend DevServer there might be limitations, eg. Vite serves the index.html
on every path, that does not contain a file extension.


v2/internal/frontend/desktop/linux/frontend.go
```diff
@@ -93,7 +82,10 @@ func NewFrontend(ctx context.Context, appoptions *options.App, myLogger *logger.
 		}
 		result.assets = assets
 
-		go result.startRequestProcessor()
+		// Start 10 processors to handle requests in parallel
+		for i := 0; i < 10; i++ {
```
leaanthony, line 86: 👍

website/docs/reference/options.mdx
```diff
@@ -316,7 +347,7 @@ Defines how the window should present itself at startup.
 | --------------- | --- | --- | --- |
 | Fullscreen      | ✅  | ✅   | ✅ |
 | Maximised       | ✅  | ✅   | ✅ |
-| Minimised       | ✅  |     | ✅ |
+| Minimised       | ✅  | ❌   | ✅ |
```
leaanthony, line 350: 🙏

leaanthony (approved):
Amazing work here! Should we link to the webview2 issue for the memory leak and mention it in the docs? I'm happy to do that if you have the link.

stffabi:
Thanks for merging 🙏

PR #1336, opened 2022-04-16
Add Elm template :)

benjamin-thomas:


leaanthony:
Fantastic! Got a lot of love for Elm!

PR #1359, opened 2022-04-28
[v2] Validate devServer property to be of the correct form

stffabi:


leaanthony:
Thanks for doing this 🙏

PR #1369, opened 2022-05-02
fix runtime.js events

polikow:
Calls on these functions were causing "OnMultiple is not defined" errors


leaanthony:
That's so weird. I fixed this, but most have been on a branch. Thanks for the pickup 🙏

PR #1372, opened 2022-05-03
Fix docs formatting

antimatter96:
I am not sure what happened to the last line [done via Github UI, might be \r\n issue]


leaanthony:
Thanks for picking this up!

PR #1373, opened 2022-05-03
[v2] Improve multi platform builds

stffabi:
- Make sure to reset the default Arch of the build options and not keep the previous one.
- Expand to the full platform in outputBinaries so e.g. `-platform windows` can be used with `-nsis`


leaanthony:
If the path of the output binaries has changed, do you think we need to update the docs?

stffabi:
Thanks for merging.

The PR should not affect the path of the output binaries. The full platform expansion is only used as key of the map from platform to the output path. This map is only used for NSIS to find the binaries.

But maybe I'm missing something.

Issue #1321, opened 2022-04-09
[V2] SIGSEV Gorilla Websocket on linux amd64 - works on windows.


efitzkiwi:
### Description

Gorilla websocket causes SIGSEV on application start in Ubuntu 20.04 (amd64 binary). This error does not happen on windows .exe files built by wails. Tested on both bleeding edge and latest stable beta.

There are two other test functions that can be tested to prove the websocket is the problem

### To Reproduce

clone https://github.com/KillerKiwiJuice/wailstestbug
cd into frontend, `npm i`
ensure code in `recv_poly_data2` is not commented out (or to make the app run, comment it out)
`wails build -platform linux/amd64`
`./build/bin/databridge`

### Expected behaviour

App should run without SIGSEV

### Screenshots

![image](https://user-images.githubusercontent.com/9599423/162548925-ac755696-bba9-4c98-a5d9-ce2e35f9ae9b.png)

### Attempted Fixes

Tried arm64 build

### System Details

```shell
System
------
OS:             Ubuntu
Version:        20.04
ID:             ubuntu
Go Version:     go1.17.1
Platform:       linux
Architecture:   amd64

Wails
------
Version:                v2.0.0-beta.34
Package Manager:        apt

Dependency      Package Name            Status          Version
----------      ------------            ------          -------
*docker         docker.io               Installed       20.10.14
gcc             build-essential         Installed       12.8ubuntu1.1
libgtk-3        libgtk-3-dev            Installed       3.24.20-0ubuntu1.1
libwebkit       libwebkit2gtk-4.0-dev   Installed       2.34.6-0ubuntu0.20.04.1
npm             npm                     Installed       6.14.4+ds-1ubuntu2
*nsis           nsis                    Available       3.05-2
pkg-config      pkg-config              Installed       0.29.1

* - Optional Dependency

Diagnosis
---------
Your system is ready for Wails development!
Optional package(s) installation details: 
  - nsis: sudo apt install nsis
```

### Additional context

Full error stack
```
2022/04/08 20:19:11 connecting to wss://testnet-explorer.binance.org/ws/block
fatal error: unexpected signal during runtime execution
[signal SIGSEGV: segmentation violation code=0x1 addr=0x100000200 pc=0x7f45228b57fa]

runtime stack:
runtime.throw({0x7649e7, 0x5e33cf5ed123d300})
        /usr/local/go/src/runtime/panic.go:1198 +0x71
runtime.sigpanic()
[...]

PR #1392, opened 2022-05-11
[v2, linux] Improve switching to main thread for callbacks

stffabi:
Make sure no pointers to the stack are passed to g_idle_add,
because at the time the callback gets executed on the
main thread, the pointer might be invalid.
Go might have reused the stack or grown the stack and the
pointer is invalid.

The concept used on Windows has been ported to Linux.

Potential fix for #1321

### DIFF
--- a/v2/internal/frontend/desktop/linux/invoke.go
+++ b/v2/internal/frontend/desktop/linux/invoke.go
@@ -1,0 +1,78 @@
+//go:build linux
+// +build linux
+
+package linux
+
+/*
+#cgo linux pkg-config: gtk+-3.0
+
+#include <stdio.h>
+#include "gtk/gtk.h"
+
+extern gboolean invokeCallbacks(void *);
+
+static inline void triggerInvokesOnMainThread() {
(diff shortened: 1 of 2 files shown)


leaanthony:
Thanks for looking at this 👍 It looks fine to me. One thing I am starting to see is duplication between the different platform frontends and am wondering if we should think about moving some things to a common base... or whether that will bite us in the end :-)

stffabi:
Thanks for looking into the PR. 

> One thing I am starting to see is duplication between the different platform frontends and am wondering if we should think about moving some things to a common base... or whether that will bite us in the end :-)

Yeah absolutely would be great to reduce those duplications 👍 .

leaanthony:
What do we need to get this out of draft? 😄

stffabi:
Not much, basically just changing the return values to void for those native methods. I'll try to clean that up during the next couple of days

leaanthony:
Looks great! Thanks for doing this. Let's see if this helps #1321

leaanthony (approved):

PR #1405, opened 2022-05-18
fix typo bindings.js to ipc.js

rayshoo:
As a result of examining the document and the html of the actual application, it is determined that ipc.js is correct and sends a PR.


stffabi (approved):
Nice find, thanks for fixing 👍.

stffabi:
I've just seen that the PR only updates the docs of version `beta.35`. Would it be possible for you to also fix the other `frontend.mdx` files of `beta.34` and of the next version?

rayshoo:
sure. I'll do it. I will close this PR and send a new PR with an update of three versions of the document.

PR #1356, opened 2022-04-26
Fix for TS/JS generation on methods with no returns

leaanthony:
Fixes #1355 

@ianmjones - a quick fix, quickly tested 👍

Issue #1355, opened 2022-04-26
Function with no return exported from app.go creates App.js & App.d.ts entries that return a Promise


ianmjones:
### Description

In `master` (moved on from `2.0.0-beta.34`) we now have the ability in the frontend JS to `import` functions from `wailsjs/go/main/App.js`, which awesome! :tada: 

For a Go function like follows...

```go
// Ping expects pong or another supplied response from the dbus service.
func (a *App) Ping(expected string) (string, error) {
	return a.serv.Ping(expected)
}
```

... App.js has ...

```js
export function Ping(arg1) {
  return window['go']['main']['App']['Ping'](arg1);
}
```

... and the companion App.d.ts has the expected ...

```ts
export function Ping(arg1:string):Promise<string|Error>;
```

However, for a Go function that has no return ...

```go
// ShowError displays an error message in a desktop dialog.
func (a *App) ShowError(title string, message string) {
	_, _ = runtime.MessageDialog(a.ctx, runtime.MessageDialogOptions{
		Type:    runtime.ErrorDialog,
		Title:   title,
		Message: message,
	})
}
```

... the App.js looks like ...

```js
export function ShowError(arg1, arg2) {
  return window['go']['main']['App']['ShowError'](arg1, arg2);
}
```

... with companion App.d.ts like ...

```ts
export function ShowError(arg1:string,arg2:string):Promise<void>;
```

This means that the function that never returns anything usable has to have a `Promise` handled when called, which feels a bit weird to me.

### To Reproduce

See description.

### Expected behaviour

If an exported Go function bound to the App instance has no return, I would not expect the App.d.ts to specify that a Promise is being returned, and ideally App.js would not use a return at all.

In the generated `wailsjs/runtime/runtime.js` some exported functions do not use a return:

```js
export function LogPrint(message) {
    window.runtime.LogPrint(message);
}
```

The companion `runtime.d.ts` is as expected, and does not specify a return type (well, just `void`):

```ts
// [LogPrint](https://wails.io/docs/reference/runtime/log#logprint)
// logs the given message as a raw message
export function LogPrint(message: string): void;
```

It would be great if `App.go` functions that do not return anything could also have a `void` return in the `App.d.ts` and no actual return in `App.js`.

### System Details

This is probably not very helpful as it's from a custom build of wails on NixOS from commit `75ee36f8ebf484bd91cf2591b00af7ce6cd4e21a`, but here it is anyway...

```shell
Wails CLI v2.0.0-beta.34


[...]

PR #1410, opened 2022-05-19
Always use return in wrapped method

### LINKED ISSUE #1408: [V2] void function need to return Promise

``` js
export function A() {
  window['go']['main']['App']['A']();
}
export function B() {
  window['go']['main']['App']['B']();
}
```

If the function does not have a return value, the generated js code has no return value, and the "Promise" object is not exposed, so the scenario of executing function B after function A is complete is impossible.

``` js
await A()
await B()
```
OR
``` js
A().then(()=>{
  B()
})
```

Changed so that the generated js function returns a result even if the binding target (Go function) is void

``` js
export function A() {
  return window['go']['main']['App']['A']();
}
export function B() {
  return window['go']['main']['App']['B']();
}
```

_No response_

### Additional context

_No response_

leaanthony:
Fixes #1408

### DIFF
--- a/v2/internal/binding/generate.go
+++ b/v2/internal/binding/generate.go
@@ -44,11 +44,7 @@ func (b *Bindings) GenerateGoBindings(baseDir string) error {
 				argsString := args.Join(", ")
 				jsoutput.WriteString(fmt.Sprintf("\nexport function %s(%s) {", methodName, argsString))
 				jsoutput.WriteString("\n")
-				if methodDetails.OutputCount() == 0 {
-					jsoutput.WriteString(fmt.Sprintf("  window['go']['%s']['%s']['%s'](%s);", packageName, structName, methodName, argsString))
-				} else {
-					jsoutput.WriteString(fmt.Sprintf("  return window['go']['%s']['%s']['%s'](%s);", packageName, structName, methodName, argsString))
-				}
+				jsoutput.WriteString(fmt.Sprintf("  return window['go']['%s']['%s']['%s'](%s);", packageName, structName, methodName, argsString))
 				jsoutput.WriteString("\n")
 				jsoutput.WriteString(fmt.Sprintf("}"))
 				jsoutput.WriteString("\n")


ianmjones (approved):
LGTM :+1: 

I was worried that it would undo the changes in #1356, but by returning void and not a void promise we get the synchronous functionality that #1408 wants back, without breaking #1355.

Nice one!

leaanthony:
Now that you've said that, I'm wondering if this does satisfy both scenarios. One wants to know when the method ends and the other doesn't care. The TS can't satisfy both (or can it 🤔 could we return `<Promise<void>|void>`)

I'm wondering whether method calls should always be async, and events should be used for "fire and forget" scenarios. That was the original intent of both and perhaps that should have been the solution to the original request. Thoughts?

ianmjones:
> I'm wondering whether method calls should always be async...

You mean, always return a promise again?

If so, then I personally don't think that's a good idea, but I'm happy to be convinced otherwise.

I feel like it's a nice set up at the moment with if there is something to be returned, and the last return item is an `error`, then you get a promise with the return item, and `catch` the error.

If there is only an `error` in the return, I take it there's still a void promise with a possible catch?

If there is nothing to return, I personally think the previous async nature was fine as there's no result to process, and therefore nothing to `await` for (or chain with `then`). However, this new setup where you can `await` if you want to is fine too.

> ..., and events should be used for "fire and forget" scenarios.

I haven't looked at the events side of things yet, but I assumed they had no expectation of a return and so were fire and forget. I assumed they were for "this thing just happened with these details", meaning you could not change the outcome of what triggered the event, only react to it.

> That was the original intent of both and perhaps that should have been the solution to the original request. Thoughts?

I hope I interpreted your intents correctly.

ianmjones:
I've slept on it, and I do wonder whether events should be synchronous or not. Could you have events that expect a response?

One of my favourite things about WordPress is its hook system (this may be an unpopular opinion). Fundamentally there are two ways to fire off a hook, as an `action` or `filter`, both are synchronous.

An [action](https://developer.wordpress.org/reference/functions/do_action/) just says "this thing is will or did happen", and has no return.

```php
do_action( 'example_action', $arg1, $arg2 );
```

While at first glance you may think, ok, this could just be async then as there is nothing returned and the emitter doesn't care. In reality, an action gives other parts of the system a chance to prepare for the event that will happen whether they like it or not. Or deal with the results of an action, before the framework continues with post action stuff.

These actions in WP are like the events in Wails at the moment as they do not expect anything back when emitted.

A [filter](https://developer.wordpress.org/reference/functions/apply_filters/) gives none or many filter handlers a chance to alter a value before it is used.

```php
$value = apply_filters( 'example_filter', 'filter me', $arg1, $arg2 );
```

These are clearly synchronous as the emitter is waiting for all filter handlers to finish taking a look at the value to see if they want to change it before it then uses the value.

Filters are probably the most used type of hook, and quite often used for "hey, should I proceed with this action" kinds of things...

```php
$cancel = apply_filters('about_to_do_thing', false, $context );
```

Wails doesn't have this kind of event, yet (maybe v3?).

So on reflection, it does seem like events should be emitted synchronously as the receiver may or may not want to prepare for whatever the event signifies is about to happen, or has happened. There's nothing to stop the event receiver from taking note of the event and then asynchronously firing off a process to deal with it, e.g. logging, or some long running task.

Sorry for the long post, but you did ask for thoughts! :joy:

leaanthony:
Am away for a week so interested in all thoughts on this!

PR #1414, opened 2022-05-21
Add docs on wails generate module

TechplexEngineer:
For users who wish to keep the wailsjs directory out of source control, the `wails generate module` command is valuable in a CI environment where wails build does not automatically generate the needed javascript module.


leaanthony:
Thanks for this!

PR #1419, opened 2022-05-22
Add missing className to input in React and Preact templates

edwardbrowncross:
CSS styling is not being applied to the input field in the hello world templates for React and Preact because the input tag is missing the class identifier present in the templates for the other frameworks.


leaanthony (changes requested):
Thanks for taking the time to do this 🙏 The templates are generated using the generate.go script (I can't link as I'm on my phone). Next time we regenerate we will lose this change. If you could update the templates that exist in `generate/assets` then we will keep the change next time we generate. Thanks

stffabi (approved):
LGTM, thanks @edwardbrowncross for these fixes.

PR #1430, opened 2022-05-30
Create gitignore when using g option

jaesung9507:


leaanthony (changes requested):
Thanks for opening this. Ioutil is deprecated but there's an equivalent method in the Io package. Otherwise LGTM

stffabi:
Please also make sure to use `filepath.Join` to join the path, otherwise it might fail on Windows.

leaanthony:
@stffabi I may be reading it wrong but it looks like forward slash is the format used for all platforms? https://git-scm.com/docs/gitignore

stffabi:
Yeah, the comment was about the os.WriteFile argument and not for the content i mn .gitignore.
Sorry I wasn't clear enough about that in my comment. 😔

stffabi:
@jaesung9507 Sorry for not being clear enough in my comment and for wasting your time with those filepath.Join in the .gitignore content.

leaanthony:
Ah all good 😃 I should have thought of that 😉

jaesung9507:
@leaanthony @stffabi 

Thank you for your kind code review.

stffabi (approved):
LGTM, thanks for this PR 👍

leaanthony (approved):

PR #1436, opened 2022-06-02
Add showcase entry for October + update homepage carousel entry for October

marcus-crane:
It's basically what it says on the tin!


leaanthony:
Nice one!

PR #1441, opened 2022-06-08
[v2, windows] Unlock OSThread after native calls have been finished

stffabi:
This makes sure the OSThread can be reused by other go
routines after a native call has been finished. Otherwise the
OSThread will be destroyed as soon as the goroutine has
finished.


leaanthony:
Nice one!

leaanthony (approved):
LGTM!

Issue #1439, opened 2022-06-07
runtime to set window topmost  or normal


chenxiao1990:
### Describe the solution you'd like

Runtime can dynamic to set window topmost  or  normal。
Now only can set window top by init .
wails.Run(&options.App{
AlwaysOnTop:      true,
})

PR #1442, opened 2022-06-08
feature: add runtime func WindowSetAlwaysOnTop

chenxiao1990:
This PR resolves feature request #1439.
Add windows 、linux platform WindowSetAlwaysOnTop function . 
Darwin platform no environment , Not implemented. 
JS runtime interface Not implemented.


leaanthony (commented):
Thanks for opening this. A few things: 
  - there's already functions for top level windows that we call during startup so we should reuse those
  - It looks like the calls to set and unset the top window are the same
  - We'd need to add a JavaScript method for completeness
  - The docs will need updating with the new runtime method

It's a good start though! 👍

chenxiao1990:
> Thanks for opening this. A few things:
> 
> * there's already functions for top level windows that we call during startup so we should reuse those
> * It looks like the calls to set and unset the top window are the same
> * We'd need to add a JavaScript method for completeness
> * The docs will need updating with the new runtime method
> 
> It's a good start though! 👍



add a JavaScript method for completeness
The docs updating with the new runtime method

chenxiao1990:
Yes, i think a single api call WindowSetAlwaysOnTop(bool)  is a good idea.
I will modify it.

leaanthony:
LGTM!

PR #1448, opened 2022-06-12
Feature: set browser path

### LINKED ISSUE #1446: Ability to specify path to WebView2

I'm building an application using wails v2, what I want to do is to download webview2 in a separate bootstrapper that will download all dependencies of the application (including webview2, if not installed), and then use downloaded (not installed in the system) version.

I think it's not very good for user experience to pop up messages and force them to download a webview or silently download it without any visual feedback.

Also, this could be helpful when debugging issues with a specific version of webview2.

Basically just add a new field in struct `windows.Options` and pass it to `createCoreWebView2EnvironmentWithOptions`

```
options := &windows.Options {
    WebviewBrowserPath: "webview_downloaded/",
}
```

_No response_

### Additional context

_No response_

NanoNik:
Resolves #1446 

This PR adds new webview2 strategy: manual

It won't download or notify user about webview2 installation status.
You must manually specify a path to a webview2 runtime in `windows.Options` structure.

If the specified directory does not exist, it will try to use webview2 installed in the system.


leaanthony:
Thanks for raising this PR and updating the documentation. @stffabi any thoughts?

stffabi (changes requested):
Thank you such much @NanoNik for putting your time into this PR. I only have some minor comments which I think should be addressed before it's getting merged. Overall it looks good to me.

v2/internal/frontend/desktop/windows/go-webview2/pkg/edge/chromium.go
```diff
@@ -84,7 +86,16 @@ func (e *Chromium) Embed(hwnd uintptr) bool {
 		dataPath = filepath.Join(os.Getenv("AppData"), currentExeName)
 	}
 
-	res, err := createCoreWebView2EnvironmentWithOptions(nil, windows.StringToUTF16Ptr(dataPath), 0, e.envCompleted)
+	var browserPathPtr *uint16 = nil
+	if e.BrowserPath != "" {
+		if _, err := os.Stat(e.BrowserPath); !errors.Is(err, os.ErrNotExist) {
+			browserPathPtr = windows.StringToUTF16Ptr(e.BrowserPath)
+		} else {
+			log.Printf("Browser path %s does not exist, will use webview2 installed in the system", e.BrowserPath)
```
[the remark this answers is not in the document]
NanoNik: Changed, now it just prints log and returns false

website/docs/reference/options.mdx
```diff
@@ -430,6 +431,14 @@ Type: string
 
 This defines the path where the WebView2 stores the user data. If empty `%APPDATA%\[BinaryName.exe]` will be used.
 
+### WebviewBrowserPath
+
+Name: WebviewBrowserPath
+
+Type: string
+
+This defines the path to a directory with WebView2 executable files and libraries. If empty, webview2 installed in the system will be used.
```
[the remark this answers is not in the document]
NanoNik: Good point, added

v2/internal/frontend/desktop/windows/go-webview2/pkg/edge/chromium.go
```diff
@@ -84,7 +86,16 @@ func (e *Chromium) Embed(hwnd uintptr) bool {
...
+		}
+	}
+
+	res, err := createCoreWebView2EnvironmentWithOptions(browserPathPtr, windows.StringToUTF16Ptr(dataPath), 0, e.envCompleted)
```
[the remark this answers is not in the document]
NanoNik: Fixed, I renamed `GetInstalledVersion` to `GetWebviewVersion` as it's now can return version from specified path

leaanthony:
Would it be possible to use `UTF16FromString` rather than `StringToUTF16Ptr` as that call has been deprecated. Thanks 👍

v2/internal/wv2installer/manual.go
```diff
@@ -0,0 +1,11 @@
+//go:build wv2runtime.manual
+// +build wv2runtime.manual
+
+package wv2installer
+
+import "github.com/wailsapp/wails/v2/pkg/options/windows"
+
+func doInstallationStrategy(installStatus installationStatus, messages *windows.Messages) error {
+	// fallback for manually specifying webview2
+	return nil
```
stffabi, line 10: During the implementation of PR #1456 I took a closer look into the installing process and now I'm unsure if we really need this manual strategy, or at least we should never return `nil` here.

Currently the `doInstallationStrategy` is called if no webview could be found or if the webview is outdated (does not meet the minimum version requirement). In the fixed version case, `no webview` (e.g. wrong folder) will be catched during the `chromium.Embed`. If the webview is outdated and by returning nil here, the creation of the webview will succeed but later on it could lead to strange panics like `AccessViolationException` because COM interfaces are used that aren't part of the outdated version.
On the other hand having a fixed version with any strategy other than `error` might be confusing. If e.g. the fixed version is outdated and the default strategy is used, the user will see a download confirmation dialog. But that will just install/update the system wide webview and it will never fix the problem.

What would you think about the following @leaanthony and @NanoNik?
 > If a fixed version is used and it would lead to a `doInstallationStrategy` call, then just error with an appropriate message similar to the `error` strategy. It won't matter which strategy has been selected during the build, it will just be ignored.

leaanthony: Given that insight, perhaps we could get some more context on the problem this PR is trying to solve.

stffabi: I think the PR is perfectly fine and it's a great feature to allow to distribute the Webview2 runtime with the app itself and to only use that runtime.

My comment was referring to if we need another Webview2 installer strategy in `wails build`:
```
 -webview2 string
    	WebView2 installer strategy: download,embed,browser,error. (default "download")
```

stffabi: Maybe we could just move the implementation of the error strategy into the common part?
Then the error strategy could simply be used as a trampoline to that and we have the implementatiom available independent from the build tags.

stffabi (approved):
LGTM. Thanks for all your effort you have put into this 🚀

leaanthony:
Please fix merge conflicts and we should be good 👍

Issue #1057, opened 2022-01-03
Is there any way to set zoom level?


e8panse:
How to achieve the same functionality as web browser, to have native zoom level?
In chromium there is <webview>.setZoomLevel(level) function.

If I just simply use: `document.body.style.zoom = 1.5;` on my frontend side it breaks my internal modules (monaco editor)

Is there any way to do this?

leaanthony:
Hi. Not currently.

pierrejoye:
I will give a PR a try using Get/Put IsZoomControlEnabled 

Idea would be:

- Add an App.run option for it 
- Add API to get/set it
- if there is an API for it, add API to get/set the zoom factor  (could not find so far :)

Thoughts?

leaanthony:
I don't have any thoughts on this. Why not try it out and see how you get on? If you are doing an `App.run` option then you can make it Windows only. If you want to add it to the runtime then it will need to support Linux and MacOS too.

pierrejoye:
> I don't have any thoughts on this. Why not try it out and see how you get on? If you are doing an `App.run` option then you can make it Windows only. If you want to add it to the runtime then it will need to support Linux and MacOS too.

Will do the option today :)

leaanthony:
Merged!

janstuemmel:
why is this only for windows?

PR #1463, opened 2022-06-18
#1057 expose ZoomFactor get/set and add the respective windows only options

pierrejoye:
As discussed in the ticket. Here is a 1st version. 

Do we have tests in general? I can add the relevant ones if any needed here (while not sure how to test the underlying COM APIs calls :)


leaanthony:
Thanks for this! Tests are super hard and time consuming, but if you'd like to get the ball rolling I'm open to that! What did you have in mind?

v2/cmd/wails/internal/commands/generate/template/base/main.tmpl.go
```diff
@@ -55,6 +55,7 @@ func main() {
 			DisableWindowIcon:    false,
 			// DisableFramelessWindowDecorations: false,
 			WebviewUserDataPath: "",
+			ZoomFactor: 1.0,
```
leaanthony, line 58: I'll regenerate the templates to include this.

v2/internal/frontend/desktop/windows/frontend.go
```diff
@@ -366,7 +366,14 @@ func (f *Frontend) setupChromium() {
 	if err != nil {
 		log.Fatal(err)
 	}
-	err = settings.PutIsZoomControlEnabled(false)
+
+
+	if opts := f.frontendOptions.Windows; opts != nil && opts.ZoomFactor > 0.0 {
+		log.Print(fmt.Sprintf("Zoom: %d", opts.ZoomFactor))
```
leaanthony, line 372: If the default is `1.0` it will print this each time. Did you mean to remove it after testing?

v2/internal/frontend/desktop/windows/frontend.go
```diff
@@ -366,7 +366,14 @@ func (f *Frontend) setupChromium() {
...
+		chromium.PutZoomFactor(opts.ZoomFactor)
+	}
+
+	err = settings.PutIsZoomControlEnabled(true)
 	if err != nil {
 		log.Fatal(err)
 	}
```
leaanthony, line 379: Should we put this section within the test for a zoom factor in case a value is not set?

leaanthony (changes requested):
Thanks or doing this. It's looking good! Just a couple of comments/questions. The only thing that would be good to add is the docs to the [options.mdx](https://github.com/wailsapp/wails/blob/master/website/docs/reference/options.mdx) file in the `website` directory 👍

v2/internal/frontend/desktop/windows/go-webview2/pkg/edge/ICoreWebView2Controller.go
```diff
@@ -1,5 +1,6 @@
 package edge
 
+import "C"
```
leaanthony, line 3: This is not required and will force the user to have gcc installed.

v2/cmd/wails/internal/commands/generate/template/base/main.tmpl.go
pierrejoye: Oh right, and for IsZoomControlEnabled :)
Thinking about it, I may add a note (and test) if one wants to apply a zoom but does allow runtime change by the users (f.e. using a config sets using a menu or other). That would be complete then, hopefully :)

pierrejoye:
> Thanks for this! Tests are super hard and time consuming, but if you'd like to get the ball rolling I'm open to that! What did you have in mind?

The easy part are the wails specific parts, generations etc. As they are almost all self contained. Unit tests for each functions are relatively straightforward for most. If not, we may need to refactor some to allow easy testing of each specific function (ie. if a function does too many things, can be splitted in separate ones without changing the public APIs).

For the integration tests (all OSes system APIs, be webviews or other), kind of tricky. Maybe writing fake functions for the C APIs. For COM, not sure, writing a fake COM sounds cumbersome. I need to think about that part :)

v2/internal/frontend/desktop/windows/frontend.go
pierrejoye: oh, indeed, sorry :)

Will remove it

v2/internal/frontend/desktop/windows/frontend.go
pierrejoye: the IsZoomControlEnabled allows the user to change the zoom. The ZoomFactor initializes the webview2 with a specific scale. The default zoomFactor is to 1.0 (by our code).

IsZoomControlEnabled can be true without setting a ZoomFactor. In that case it will be 1.0 and the user can change it at wish.

It is also possible to do not allow the user to change the zoomFactor but sets one using the config (f.e. for a kiosk like app, to fit a screen nicely).

As you pointed out, this is what I still have to add to this PR, in the documentation. It is not complicated but with no documentation it can be confusing :) Thoughts?

leaanthony: Should this be `true` or `options.IsZoomControlEnabled`?

v2/internal/frontend/desktop/windows/go-webview2/pkg/edge/ICoreWebView2Controller.go
```diff
@@ -130,3 +131,36 @@ func (i *ICoreWebView2Controller) NotifyParentWindowPositionChanged() error {
 	}
 	return nil
 }
+
+func Float64fromBits(b uint64) float64 {
```
stffabi, line 135: Would it be possible to use math.Float64frombits and math.Float64bits instead?

v2/cmd/wails/internal/commands/generate/template/base/main.tmpl.go
pierrejoye: Added to the mdx :)

v2/internal/frontend/desktop/windows/frontend.go
pierrejoye: removed

v2/internal/frontend/desktop/windows/frontend.go
pierrejoye: Fixed he logic and add the option as parameter to the COM API call, both ZoomFactor.

v2/internal/frontend/desktop/windows/go-webview2/pkg/edge/ICoreWebView2Controller.go
pierrejoye: removed

v2/internal/frontend/desktop/windows/go-webview2/pkg/edge/ICoreWebView2Controller.go
pierrejoye: Oh thanks, I did not know about them, they are now used

pierrejoye:
Thanks for the great feedback and suggestion :) I will see which part I can add tests in the next couple of days.

v2/internal/frontend/desktop/windows/frontend.go
```diff
@@ -366,10 +366,19 @@ func (f *Frontend) setupChromium() {
...
-	err = settings.PutIsZoomControlEnabled(false)
-	if err != nil {
-		log.Fatal(err)
+
+
+	if opts := f.frontendOptions.Windows; opts != nil  {
+		if  opts.ZoomFactor > 0.0 {
```
leaanthony, line 372: Should this be `opts.ZoomFactor > 0.0 && opts.ZoomFactor != 1.0` to ignore when the default value is set?

leaanthony:
Any updates on this @pierrejoye? I wouldn't want this blocked because of trying to work out the hard problem of testing 👍 Let me know how we can move it forward 👍

leaanthony:
@pierrejoye any news on this?

pierrejoye:
hello :)

sorry for the late reply, I did not have the time to work on the test.

that patch is used in prod here since I made the PR though  :)

leaanthony:
Hey there 👋 We'll look at this soon 👍

leaanthony (approved):

stffabi (approved):
LGTM 👍 , thanks for this.

PR #1466, opened 2022-06-19
Set Window Background colour if provided + debounce redraw option

leaanthony:
Sets window background colour if provided.
Added `ResizeDebounceMS` for specifying the delay before redraw when resizing.

This PR provides options to decrease the "flicker" when resizing the app on Windows.


stffabi (approved):
LGTM, great we now have a flicker free resizing on Windows 🚀.

I've also added a commit to workaround the flickering in Frameless mode.

PR #1473, opened 2022-06-20
Improve theme change at runtime

leaanthony:
Fixes runtime theme changes

### DIFF
--- a/v2/internal/frontend/desktop/windows/theme.go
+++ b/v2/internal/frontend/desktop/windows/theme.go
@@ -7,6 +7,12 @@ import (
 
 func (w *Window) updateTheme() {
 
+	// Don't redraw theme if nothing has changed
+	if !w.themeChanged {
+		return
+	}
+	w.themeChanged = false
+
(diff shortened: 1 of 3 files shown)


v2/internal/frontend/desktop/windows/theme.go
```diff
@@ -7,32 +7,37 @@ import (
...
-		if winOptions.Theme == windows.Light {
-			isDarkMode = false
-		}
 	}
-
-	win32.SetTheme(w.Handle(), isDarkMode)
-
 	// Custom theme
 	if win32.SupportsCustomThemes() && customTheme != nil {
 		if w.isActive {
```
stffabi, line 43: Don't we need to also set `w.themeChanged = true` when `w.isActive` gets changed during `w32.WM_ACTIVATE`?
I think we might loose the adaption of the active/inactive color when customThemes are active.

stffabi (commented):
Nice work, that definitely improve the theme changing 🚀 .

I just added some comments about possibly loosing the adaption to system changes, like active/inactive and changing the system theme.

PR #1488, opened 2022-06-23
Better signal handling

leaanthony:
Noticed `ctrl-c` isn't handled and leads to ungraceful shutdown.

### DIFF
--- a/v2/internal/signal/signal.go
+++ b/v2/internal/signal/signal.go
@@ -1,73 +1,43 @@
 package signal
 
 import (
-	"context"
 	"os"
 	gosignal "os/signal"
 	"sync"
 	"syscall"
-
-	"github.com/wailsapp/wails/v2/internal/logger"
-	"github.com/wailsapp/wails/v2/internal/servicebus"
 )
 
-// Manager manages signals such as CTRL-C
-type Manager struct {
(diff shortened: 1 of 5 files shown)


v2/internal/signal/signal.go
```diff
@@ -1,73 +1,43 @@
...
 // Start the Signal Manager
-func (m *Manager) Start() {
+func Start() {
 
 	// Hook into interrupts
-	gosignal.Notify(m.signalchannel, os.Interrupt, syscall.SIGTERM, syscall.SIGINT)
+	gosignal.Notify(signalChannel, os.Interrupt, syscall.SIGTERM, syscall.SIGINT)
 
-	m.wg.Add(1)
+	var wg sync.WaitGroup
```
stffabi, line 27: It seems like we are not calling `wg.Wait()` anywhere. Are we maybe missing that or could we just remove the waitgroup completely?

stffabi (approved):
Nice improvement 👍. LGTM, just a minor comment if we need the waitgroup.

v2/internal/signal/signal.go
leaanthony: Looks like it was left in after the simplification. I'll remove it

PR #1489, opened 2022-06-24
Chore/react 18 create root

tomanagle:
- Updates the react & react-ts frontends to use `createRoot` instead of `ReactDOM.render` as ReactDOM.render is no longer supported in React 18 

Docs: https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#updates-to-client-rendering-apis

### DIFF
--- a/v2/cmd/wails/internal/commands/initialise/templates/generate/assets/preact/frontend/src/main.jsx
+++ b/v2/cmd/wails/internal/commands/initialise/templates/generate/assets/preact/frontend/src/main.jsx
@@ -1,5 +1,14 @@
-import {render} from 'preact';
-import {App} from './app';
-import './style.css';
+import React from 'react'
+import { createRoot } from 'react-dom/client'
+import './style.css'
+import App from './App'
 
-render(<App/>, document.getElementById('app'));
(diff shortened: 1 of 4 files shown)


sidwebworks (approved):
Looks good :smile:

leaanthony:
Thanks for doing this! Is there any chance you could add them to the directories under `generate`? My fear is they'll get overwritten next time we generate the templates again. It should be fairly obvious what's happening in the generate directory (those files are copied after a default vite generation) but if you need more info, I'm happy to help! 🙏

tomanagle:
> Thanks for doing this! Is there any chance you could add them to the directories under `generate`? My fear is they'll get overwritten next time we generate the templates again. It should be fairly obvious what's happening in the generate directory (those files are copied after a default vite generation) but if you need more info, I'm happy to help! 🙏

Yup, I'll take a look around and update the PR :)

tomanagle:
I updated those templates (hopefully in the right place).

Also realised my linter was changing more lines than required, so fixed that up too

leaanthony (approved):

leaanthony:
Thanks for this!

Issue #1491, opened 2022-06-24
Can't make it work with SolidJS-Start as it doesn't produce `index.html`


nikivdev:
### Description

Not sure if bug or a feature request. But I have a Wails repo setup here:

https://github.com/learn-anything/learn-anything/tree/main/app

My front end is built from https://github.com/solidjs/solid-start

I modified wails.json to run my `pnpm run ..` commands: https://github.com/learn-anything/learn-anything/blob/main/app/wails.json

I assume the issue is that with Wails-React setup, I need an `index.html` file like [in here](https://github.com/nikitavoloboev/test/blob/main/wails-react/frontend/index.html). That supposedly should call a [file similar to this but for SolidJS](https://github.com/nikitavoloboev/test/blob/main/wails-react/frontend/src/main.tsx). The thing is that Solid-Start uses Vite for bundling. And my output dir (called netlify) doesn't give me any .html files. 😿

<img width="190" alt="image" src="https://user-images.githubusercontent.com/6391776/175515250-59544a59-c3a9-4a68-98cc-863c10e70083.png">

Not sure how to go about solving this.

### To Reproduce

1. Clone https://github.com/learn-anything/learn-anything 
2. `cd app`
3. `wails dev`

See this:

<img width="735" alt="image" src="https://user-images.githubusercontent.com/6391776/175514747-96a4d5b8-7d12-4005-b66e-ad2def2909d1.png">

### Expected behaviour

It works similar to other Svelte/React templates.

### System Details

```shell
wails doctor
Wails CLI v2.0.0-beta.37


Scanning system - Please wait (this may take a long time)...Done.

System
------
OS:		MacOS
Version: 	12.4
ID:		199506
Go Version:	go1.19beta1
Platform:	darwin
Architecture:	arm64

Wails
------
Version: 	v2.0.0-beta.37

Dependency			Package Name	Status		Version
----------			------------	------		-------
xcode command line tools 	N/A		Installed	2395
npm 				N/A		Installed	8.5.5
*upx 				N/A		Available	
*nsis 				N/A		Available	

* - Optional Dependency

Diagnosis
---------
Your system is ready for Wails development!
Optional package(s) installation details:
  - upx : Available at https://upx.github.io/
  - nsis : Available at https://nsis.sourceforge.io/Download
```

leaanthony:
[...]

PR #1492, opened 2022-06-24
Add Solid JS template

sidwebworks:
I read this issue #1491 and decided to create a Solid JS template for Wails.


leaanthony:
Nice! I noticed that step 2 in your instructions involves a manual step. If you set `frontend:build` and `frontend:install` it should happen automatically. https://wails.io/docs/reference/project-config

sidwebworks:
@leaanthony thanks buddy, I will update the template README

PR #1494, opened 2022-06-25
Fix lit-ts template

Orijhins:
This pull fixes a bug where a newly generated lit-ts template would have incorrect syntax in the fronted/src/my-element.ts file


leaanthony (commented):
Interesting. The template is the default vite template. Do you know if it's an issue there too?

Thanks for raising this 🙏

leaanthony:
Rather than update the template itself, could you add the file under the generate directory so it isn't over written next time we regenerate? Thanks!

Orijhins:
I didn't see any problem with Vite so far. If they had this same problem in their template, it seems to have been fixed:
https://stackblitz.com/edit/vitejs-vite-vmnt6c?file=src%2Fmy-element.ts&terminal=dev

leaanthony:
It may be a recent fix. I'll check why the generated code isn't as expected 👍

Orijhins:
I fixed it with the latest commit, but it is indeed interesting

v2/cmd/wails/internal/commands/initialise/templates/templates/lit-ts/frontend/src/my-element.ts
```diff
@@ -68,32 +68,31 @@ export class MyElement extends LitElement {
...
-{
-    let thisName = this.shadowRoot.getElementById('name').value
-    Greet(thisName).then(result => {
-        this.resultText = result
-    });
-}
+    greet() {
+        let thisName = (this.shadowRoot.getElementById('name') as HTMLInputElement)?.value;
+        if (thisName) {
+            Greet(thisName || 'UAAA').then(result => {
```
leaanthony, line 78: 'UAAA' ?

v2/cmd/wails/internal/commands/initialise/templates/generate/assets/lit-ts/frontend/src/my-element.ts
```diff
@@ -68,32 +68,31 @@ export class MyElement extends LitElement {
...
-{
-    let thisName = this.shadowRoot.getElementById('name').value
-    Greet(thisName).then(result => {
-        this.resultText = result
-    });
-}
+    greet() {
+        let thisName = (this.shadowRoot.getElementById('name') as HTMLInputElement)?.value;
+        if (thisName) {
+            Greet(thisName || 'UAAA').then(result => {
```
leaanthony, line 78: 'UAAA'?

v2/cmd/wails/internal/commands/initialise/templates/templates/lit-ts/frontend/src/my-element.ts
Orijhins: My bad, it was a bad null check ! Value and innerText returned null until I applie the HTMLInputElement type assertion. Fixed with latest commit

v2/cmd/wails/internal/commands/initialise/templates/generate/assets/lit-ts/frontend/src/my-element.ts
Orijhins: My bad, it was a bad null check ! Value and innerText returned null until I applie the HTMLInputElement type assertion. Fixed with latest commit

PR #1506, opened 2022-06-29
Rename WindowSetRGBA -> WindowSetBackgroundColour

leaanthony:
The name of this method is ambiguous and unhelpful. Also the signature for WindowSetRGBA was incorrect for Javascript. This PR fixes all of that and gives it a proper name.


leaanthony:
Looks like a huge PR but most of the changes are the same `runtime.d.ts` and `runtime.js` files in the templates

stffabi (approved):
LGTM 👍

leaanthony:
Thanks for double checking this. It had been bothering me since forever.

PR #1534, opened 2022-07-05
Allow -appargs flag to pass flags to binary.

### LINKED ISSUE #1533: wails dev -appargs does not allow for args that are flags
### Description

When using `wails dev` and you need to pass arguments to the app, wails provides an `-appargs` flag which can take a quoted string of args to be passed to the binary.

However, if those quoted args are flags, such as follows:

`wails dev -appargs '-hello'`

... the app never launches, and includes an error signifying that the arg is not recognised:

```
flag provided but not defined: -hello
Usage of /home/ian/Projects/scratch/devargs/build/bin/devargs-dev-linux-amd64:
  -assetdir string
        Directory to serve assets
  -loglevel string
        Loglevel to use - Trace, Debug, Info, Warning, Error (default "debug")
```

### To Reproduce

1. Create new wails project.
2. Run `wails dev -appargs "hello"`
3. Everything works as expected.
4. Quit app.
5. Run `wails dev -appargs "-hello"`
6. App never shows.

### Expected behaviour

Args given to `-appargs` that are flags are passed onto binary when using `wails dev`.

### Screenshots

_No response_

### Attempted Fixes

PR incoming! :smile: 

### System Details

```shell
Wails CLI v2.0.0-beta.38


System
------
OS:             NixOS
Version:        22.11
ID:             nixos
Go Version:     go1.17.11
Platform:       linux
Architecture:   amd64

Wails
------
Version:        v2.0.0-beta.38

Dependency      Package Name    Status  Version
----------      ------------    ------  -------

Diagnosis
---------
```

### Additional context

_No response_

ianmjones:
Resolves #1533 

This change allows `-appargs` to pass flags to the binary when using `wails dev`.

This is needed because if you do not set "assetdir" in `wails.json` then the `App` created by `wails dev` tries to check the command line flags for `-assetdir` and a few more flags, and then fails when it finds the app specific flag being passed on the command line which is unknown to the flagset being used in the dev `CreateApp`.

The only way I could figure out to achieve this was to use a distinct FlagSet and set it to continue on error. I think this is ok as it only applies to running via `wails dev`, but wouldn't be too upset if someone has a better solution! :smile: 

There's a bonus typo fix in the help message for `-appargs` and gofmt fixes included for good measure too.

### DIFF
--- a/v2/cmd/wails/internal/commands/dev/dev.go
+++ b/v2/cmd/wails/internal/commands/dev/dev.go
@@ -104,7 +104,7 @@ func AddSubcommand(app *clir.Cli, w io.Writer) error {
 	command.IntFlag("debounce", "The amount of time to wait to trigger a reload on change", &flags.debounceMS)
 	command.StringFlag("devserver", "The address of the wails dev server", &flags.devServer)
 	command.StringFlag("frontenddevserverurl", "The url of the external frontend dev server to use", &flags.frontendDevServerURL)
-	command.StringFlag("appargs", "arguments to pass to the underlying app (quoted and space searated)", &flags.appargs)
+	command.StringFlag("appargs", "arguments to pass to the underlying app (quoted and space separated)", &flags.appargs)
 	command.BoolFlag("save", "Save given flags as defaults", &flags.saveConfig)
 	command.BoolFlag("race", "Build with Go's race detector", &flags.raceDetector)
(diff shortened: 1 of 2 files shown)


leaanthony:
Awesome Ian 👍 I've no better suggestions right now so let's go for it

Issue #1554, opened 2022-07-13
[v2] Windows: build with race flag crash with fatal error on windows


hminbdxxxx:
### Description

All dependencies are installed. Use `wails init -n example` and `wails build -debug -race`, launch the exe and it will show `fatal error: checkptr: pointer arithmetic computed bad pointer value`.
I tried use `wails build -race` instead and the exe showed nothing.

### To Reproduce

1. wails init -n example
2. wails build -debug -race
3. launch the exe

### Expected behaviour

exe and race detactor should work

### System Details

```shell
Wails CLI v2.0.0-beta.38


Scanning system - Please wait (this may take a long time)...Done.

System
------
OS:             Windows 10 Pro
Version:        2009 (Build: 19044)
ID:             21H2
Go Version:     go1.18
Platform:       windows
Architecture:   amd64

Wails
------
Version:        v2.0.0-beta.38

Dependency      Package Name    Status          Version
----------      ------------    ------          -------
WebView2        N/A             Installed       103.0.1264.49
npm             N/A             Installed       8.3.1
*upx            N/A             Available
*nsis           N/A             Available

* - Optional Dependency

Diagnosis
---------
Your system is ready for Wails development!
Optional package(s) installation details:
  - upx : Available at https://upx.github.io/
  - nsis : Available at https://nsis.sourceforge.io/Download
```

### Additional context

full stack trace
```
PS D:\projects\example> .\build\bin\example.exe
fatal error: checkptr: pointer arithmetic computed bad pointer value

goroutine 1 [running, locked to thread]:
runtime.throw({0x10a3ebe?, 0x8?})
        D:/Go/src/runtime/panic.go:992 +0x76 fp=0xc00007fb00 sp=0xc00007fad0 pc=0x9d8156
runtime.checkptrArithmetic(0xff80000000?, {0x0?, 0x225303028f0?, 0xd0?})
        D:/Go/src/runtime/checkptr.go:52 +0xbb fp=0xc00007fb30 sp=0xc00007fb00 pc=0x9a82bb
github.com/wailsapp/wails/v2/internal/frontend/desktop/windows/winc/w32.MakeIntResource(...)
        C:/Users/ThinkPad/go/pkg/mod/github.com/wailsapp/wails/v2@v2.0.0-beta.38/internal/frontend/desktop/windows/winc/w32/utils.go:41
[...]

PR #1556, opened 2022-07-13
Remove usage of unsafe.Pointer in winc

leaanthony:
Compiling with `-race` causes a standard windows app to crash. This fixes that. Potential fix for #1554


leaanthony:
Not sure this is the *correct* fix, but it does fix the problem.

stffabi:
So I think we now return the wrong value.

leaanthony:
@stffabi - yeah, I understand the problem a bit more now. It appears that the use of unsafe pointer and casting to a `*uint16` is how most projects on GitHub do this. Any thoughts?

stffabi:
I think we would need to work with a `uintptr` type instead of a `*uint16`. Which would need some changes on the caller side but I think that might be the only way to satisfy the race detector.

@leaanthony If you would like I could add a proposal to this PR.

leaanthony:
Thanks @stffabi ! I did try changing to uintptr and got reasonably far but it didn't appear to change anything. I was probably doing it wrong

stffabi:
> Thanks @stffabi ! I did try changing to uintptr and got reasonably far but it didn't appear to change anything. I was probably doing it wrong

You have still seen the race error or the missing icon?

stffabi:
I've added a proposal, with this the icon will be displayed correctly. I wasn't able to test the `--race` because it seems like it's not supported on arm64 and I only have Windows ARM64.

leaanthony:
I've just pushed a fix for `EnumProc` which was failing in a similar way.

stffabi:
I think there might be still one edge case, the &monitorContainer pointer should be passed as `unsafe.Pointer` and not uintptr here:

```go
succeeded := w32.EnumDisplayMonitors(dc, nil, syscall.NewCallback(EnumProc), uintptr(unsafe.Pointer(&monitorContainer)))
```

Otherwise go might not now that this pointer is escaping to the syscall and when growing the stack the pointer might get invalid.

leaanthony:
Good catch. I've pushed a commit that I think fixes what you raised. Please review @stffabi  👍

stffabi (approved):
LGTM

stffabi:
Just for doc purposes, if one needs to just have the checkptr active without `go race` (which needs CGO) one can just use the following flags. 

```sh
go build -gcflags "all=-N -l -d=checkptr"
```

Issue #1517, opened 2022-07-01
Cannot set size of window on Ubuntu 22.04 (heigh and width is 0)


skamensky:
### Description

When running wails, both in dev and build mode on Ubuntu, the height and width are zero. I ran the same exact code on MacOS and it works. 

It's stuck at (seemingly) 0 width and 0 height. If you pass it the `Fullscreen` argument in app initialization, or call `runtime.WindowFullscreen(ctx)` from go, you still start off with 0 widht and 0 height, but it becomes resizable via the GUI.

OS Version:

```
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04 LTS
Release:	22.04
Codename:	jammy
```

### To Reproduce

Initialize project: `wails init  -n wails_issue_repro -t react-ts`
Run `wails dev` or run `wails build` and execute build executable

### Expected behaviour

Window respects size of initial app config and is programatically resizable.

### Screenshots

![image](https://user-images.githubusercontent.com/19151369/176964506-866b2569-d0b9-4930-8045-a1480b157c77.png)

### Attempted Fixes

Search through github and did not find any fixes or same issues

### System Details

```shell
~/tmp/wails-issue-repro/wails_issue_repro$  wails doctor
Wails CLI v2.0.0-beta.38


Scanning system - Please wait (this may take a long time)...Done.

System
------
OS:		Ubuntu
Version: 	22.04
ID:		ubuntu
Go Version:	go1.18.1
Platform:	linux
Architecture:	amd64

Wails
------
Version: 		v2.0.0-beta.38
Package Manager: 	apt

Dependency	Package Name		Status		Version
----------	------------		------		-------
*docker		docker.io		Installed	20.10.17
gcc		build-essential		Installed	11.2.0
libgtk-3	libgtk-3-dev		Installed	3.24.33-1ubuntu2
libwebkit	libwebkit2gtk-4.0-dev	Installed	2.36.3-0ubuntu0.22.04.1
npm		npm			Installed	8.5.1~ds-1
*nsis		nsis			Installed	v3.08-2
pkg-config	pkg-config		Installed	0.29.2

* - Optional Dependency

Diagnosis
---------
Your system is ready for Wails development!


If Wails is useful to you or your company, please consider sponsoring the project:
https://github.com/sponsors/leaanthony
```

skamensky:
After posting this I saw maintainers asking whether issues occur when using the vanilla template. I tried again with the vanilla template and I'm still seeing the same issue.

[...]

PR #1557, opened 2022-07-13
Use scaling when setting min/max window

leaanthony:
Scaling is only considered when setting the window to fullscreen. This PR considers the scale when setting min/max width/height.

Potential fix for #1517.


acheong08:
This seems to be stable, when will it be merged?

leaanthony:
Waiting for @skamensky to confirm about appoptions Fullscreen: https://github.com/wailsapp/wails/issues/1517#issuecomment-1186135498

leaanthony:
@skamensky happy for this to be merged?

skamensky:
From my I've seen it's a partial fix without any regressesions. I'll check layer today to see if there's something I can do to make it a complete fix. But if not, I see no reason not to merge it.

leaanthony:
It was a full fix on my system. Startup options worked as expected

Issue #1588, opened 2022-07-18
[V2] Mac: Cannot show menus correctly


cxt90730:
### Description

Cannot show menus correctly on mac. Location and label display is not what I expect. I don't know if the document needs to be updated?  (e.g. runtime.Quit now needs ctx of lifecycle, but document of `Menus` do not show this.)

### To Reproduce

Mycode:

```
        AppMenu := menu.NewMenu()
	FileMenu := AppMenu.AddSubmenu("File")
	FileMenu.AddText("WindowSetTitle", keys.CmdOrCtrl("1"), func(_ *menu.CallbackData) {
		runtime.WindowSetTitle(ctx, time.Now().Local().String())
	})
	FileMenu.AddSeparator()
	FileMenu.AddText("WindowCenter", keys.CmdOrCtrl("2"), func(_ *menu.CallbackData) {
		runtime.WindowCenter(ctx)
	})
	FileMenu.AddSeparator()
	FileMenu.AddText("WindowSetLightTheme", keys.CmdOrCtrl("3"), func(_ *menu.CallbackData) {
		runtime.WindowSetLightTheme(ctx)
	})
	FileMenu.AddText("WindowSetDarkTheme", keys.CmdOrCtrl("4"), func(_ *menu.CallbackData) {
		runtime.WindowSetDarkTheme(ctx)
	})
	FileMenu.AddSeparator()
	FileMenu.AddText("WindowGetSize", keys.CmdOrCtrl("5"), func(_ *menu.CallbackData) {
		fmt.Println(runtime.WindowGetSize(ctx))
	})
	if os_runtime.GOOS == "darwin" {
		AppMenu.Append(menu.AppMenu())
		AppMenu.Append(menu.EditMenu()) // on macos platform, we should append EditMenu to enable Cmd+C,Cmd+V,Cmd+Z... shortcut
	}
	runtime.MenuSetApplicationMenu(ctx, AppMenu)
```

![Uploading image.png…]()

### Expected behaviour

display correct

### System Details

```shell
unnecessary...
```

cxt90730:
![image](https://user-images.githubusercontent.com/11382804/179446719-f037dbd1-d677-4f94-a1ce-8a5c4934fb04.png)

cxt90730:
emmm, I need to add these code before:

```
         AppMenu := menu.NewMenu()
	if os_runtime.GOOS == "darwin" {
		AppMenu.Append(menu.AppMenu())
	}
```

[...]

PR #1598, opened 2022-07-19
Update macOS examples and outdated runtime docs

acheong08:
Reference https://github.com/wailsapp/wails/issues/1588


acheong08:
Conflicts need to be fixed...

Issue #1583, opened 2022-07-17
A runtime method to manage application focus state


purplesmoke05:
### Is your feature request related to a problem? Please describe.

Could we have an API managing the application focus state such as active/inactive.

When WindowHide() is called in mac environment, only hiding window and the application is still active.

### Describe the solution you'd like

A runtime method to make the application inactive would be appreciated.

leaanthony:
Thanks for the request. Which Cocoa APIs were you specifically thinking of?

purplesmoke05:
Thanks for getting back to me.
I am not familiar with Cocoa at all, but ```NSApplication:hide()``` looks close to what I was thinking.

https://developer.apple.com/documentation/appkit/nsapplication/1428733-hide/

Or just being able to invoke closing window would be great.

leaanthony:
Yeah interesting. We're moving the current window out using a different method. This method will hide all windows. Perhaps we should call it `WindowHideAll`? Could also do a `WindowShowAll`.

leaanthony:
@purplesmoke05 - Please try the linked PR. It adds `runtime.Hide()` and `runtime.Show()` which work at the application level. On Mac, this uses a `hide:` and `unhide:` so it affects the application, not just the main window.

purplesmoke05:
Awesome.
This is exactly the behavior I was looking for.

PR #1599, opened 2022-07-19
Add Show() and Hide() to runtime to show/hide application

leaanthony:
Potential fix for #1583 
It should be noted that for Windows and Linux, this will simply call WindowShow and WindowHide.
Mac has the concept of hiding applications which is different from hiding the window. That's what this API does.


stffabi (approved):
LGTM 👍

Issue #1610, opened 2022-07-19
Golang build tags not working with wails build


monopolly:
### Description

Hi guys. 
I am stuck.
Wails cli build don't see a packages build's tags for other platforms like "+windows"
But go build works fine. Could you advice me how to solve it?
I am newbie in desktop apps, so maybe I do something wrong?

### To Reproduce

- Init wails project
- Add module: go get github.com/akrylysov/pogreb@v0.10.1
- Run: GOOS=windows GOARCH=amd64 wails build -clean -platform windows
- Got error: ../../../../../pkg/mod/github.com/akrylysov/pogreb@v0.10.1/fs/os_mmap_windows.go:23:13: undeclared name maxMmapSize for array length

### Expected behaviour

Its not an error

### Attempted Fixes

I tried different tags for wails build but no success.

But clean go build works fine
GOOS=windows GOARCH=amd64 go build -tags desktop,production -ldflags "-w -s -H windowsgui" -o example.exe

### System Details

```shell
System
------
OS:		MacOS
Version: 	12.4
ID:		199506
Go Version:	go1.18
Platform:	darwin
Architecture:	amd64

Wails
------
Version: 	v2.0.0-beta.38

Dependency			Package Name	Status		Version
----------			------------	------		-------
xcode command line tools 	N/A		Installed	2395
npm 				N/A		Installed	8.11.0
*upx 				N/A		Available	
*nsis 				N/A		Installed	v3.08

* - Optional Dependency

Diagnosis
---------
Your system is ready for Wails development!
Optional package(s) installation details: 
  - upx : Available at https://upx.github.io/
```

leaanthony:
Thanks for opening this @monopolly. You found a genuine bug, though I'm not sure we really saw it as a bug as `-platform` would still work as expected (In your case, `-platform windows/amd64` with no `GOARCH` would work fine). That being said, I think it's reasonable to allow the env vars to override the platform. The attached PR does this. Please test using [this guide](https://wails.io/docs/guides/bleeding-edge). Let me know how you get on👍

leaanthony:
This is now merged. $GOOS will be used by default if no platform specified. $GOARCH will be used if no arch specified. If these aren't given then runime.GOOS and runtime.GOARCH are used 👍

PR #1618, opened 2022-07-20
Override platform with GOOS and GOARCH

leaanthony:
Whilst `-platform` worked as expected, the `GOOS` and `GOARCH` flags were not overriding the platform.
Fix for #1610


v2/cmd/wails/internal/commands/build/build.go
```diff
@@ -162,7 +162,24 @@ func AddBuildSubcommand(app *clir.Cli, w io.Writer) {
...
 			modeString = "Debug"
 		}
+		splitPlatform := strings.Split(platform, "/")
+		targetPlatform := os.Getenv("GOOS")
+		targetArch := os.Getenv("GOARCH")
+		if targetPlatform == "" {
+			targetPlatform = splitPlatform[0]
+		}
+		if targetArch == "" {
+			targetArch = runtime.GOOS
```
stffabi, line 172: Shouldn't this be: 
```suggestion
			targetArch = runtime.GOARCH
```

v2/cmd/wails/internal/commands/build/build.go
```diff
@@ -162,7 +162,24 @@ func AddBuildSubcommand(app *clir.Cli, w io.Writer) {
...
+			if len(splitPlatform) > 1 {
+				targetArch = splitPlatform[1]
+			} else {
+				if system.IsAppleSilicon {
+					targetArch = "arm64"
+				}
+			}
+		}
 
+		platform := targetPlatform + "/" + targetArch
```
stffabi, line 182: Doesn't this conflict with the multi target building where we could specify multiple platforms separated by `,`? On Line 184 we split the platform by `,` but since we rebuild the platform here we loose all additional defined platforms.

stffabi (changes requested):
I suspect we loose the ability to do multi platform builds in a correct way.

Should we either allow setting the GOOS/GOARCH env vars or the platform string, but not both?

v2/cmd/wails/internal/commands/build/build.go
leaanthony: Yes, you do, but I'm going to assume the person compiling is overriding the platform or arch in this way for a reason. We could print a warning?

stffabi: Yeah but it seems like even when not overriding with thr env vars, one can't build multiple platforms anymore.

leaanthony: Yeah, that's an issue, but I'm suggesting reverting all of this and just setting the default platform right at the start to be `$GOOS/$GOARCH` if specified or `runtime.GOOS/runtime.GOARCH` if not.

stffabi: That sounds great 👍

v2/cmd/wails/internal/commands/build/build.go
```diff
@@ -49,13 +49,19 @@ func AddBuildSubcommand(app *clir.Cli, w io.Writer) {
...
+	defaultPlatform := os.Getenv("GOOS")
+	if defaultPlatform == "" {
+		defaultPlatform = runtime.GOOS
 	}
+	defaultArch := os.Getenv("GOARCH")
+	if defaultArch == "" {
+		if system.IsAppleSilicon {
+			defaultArch = "arm64"
+		} else {
+			defaultArch += runtime.GOARCH
```
stffabi, line 61: Just a minor adjustment 
```suggestion
			defaultArch = runtime.GOARCH
```

v2/cmd/wails/internal/commands/build/build.go
```diff
@@ -269,14 +279,7 @@ func AddBuildSubcommand(app *clir.Cli, w io.Writer) {
...
 			buildOptions.Platform = platformSplit[0]
-			buildOptions.Arch = runtime.GOARCH
-			if system.IsAppleSilicon {
-				buildOptions.Arch = "arm64"
-			}
-			if len(platformSplit) == 2 {
-				buildOptions.Arch = platformSplit[1]
-			}
-
+			buildOptions.Arch = platformSplit[1]
```
stffabi, line 282: We should still initialise `buildOptions.Arch = defaultArch` and only set the second array item if it exists. Otherwise people might see a `index out of range` if they specify a platform without arch.

v2/cmd/wails/internal/commands/build/build.go
leaanthony: That's weird, I thought I caught that :/

v2/cmd/wails/internal/commands/build/build.go
leaanthony: Fixed. Copilot is freaky accurate...

stffabi: Never used it so far.

stffabi (approved):
LGTM 👍

leaanthony:
Thanks heaps for reviewing. Great catches! 🔥

PR #1631, opened 2022-07-21
add go for syntax highlight

leoliuasia:


leoliuasia (commented):
add go for markdown syntax highlight

PR #1632, opened 2022-07-21
Add -s support for wails dev

### LINKED ISSUE #1623: Wails dev mode do not compile front-end command line options

Can wails add a function command option that the development mode does not automatically compile the front end? This can speed up the test speed of go. It takes a long time for the wails dev department to recompile the front end every time

demo:
wails dev --without-build-frontend

_No response_

### Additional context

_No response_

leaanthony:
Fixes #1623


stffabi:
Just crossed my mind, couldn't we skip building the frontend in dev mode if we use an external dev server?

Since #1615 we install all dependencies for the frontend/devserver outside of the build. So we could always skip building the frontend in that case, since we only proxy to the external dev server.

leaanthony:
That's a valid point. I think we should keep this flag for when someone doesn't use a Dev server but we should skip for when they do 👍

stffabi:
Yeah, absolutely.

leaanthony:
@stffabi - looks like you added this scenario in March: https://github.com/wailsapp/wails/blob/master/v2/cmd/wails/internal/commands/dev/dev.go#L689-L691

If you're happy this covers that scenario then I think this is ready for merging 👍

stffabi:
> @stffabi - looks like you added this scenario in March: https://github.com/wailsapp/wails/blob/master/v2/cmd/wails/internal/commands/dev/dev.go#L689-L691

That will only skip triggering a reload of the frontend when frontend files get changed. It won't prevent building the frontend when a rebuild due to changed go files is triggered.

I think the following should do what we want:
```diff
- IgnoreFrontend: false,
+ IgnoreFrontend: flags.skipFrontend || flags.frontendDevServerURL != "",
```

leaanthony:
Perfect! I'll get that in tonight.

PR #1639, opened 2022-07-23
Deprecate Fullscreen appoption

acheong08:
Add comments and notes to recommend `WindowStartState: options.Fullscreen` instead


leaanthony:
Thanks for this. Looks like it's mixed up with another PR. Any chance of removing those files or opening a new PR?

PR #1644, opened 2022-07-24
Generalize manual compile steps [Documentation]

acheong08:
`go build -tags desktop,production -ldflags "-w -s -H windowsgui"` returns an error on MacOS/Linux.
It is referenced in Wails CLI that manually only `-w -s` are passed by default.


leaanthony:
Yeah, nice! Thank you 🙏

PR #1651, opened 2022-07-25
[dev] Ignore empty install command, we also do ignore it during the build step

stffabi:
Otherwise we can't use `wails dev` for projects where `wails build` works, e.g. on the plain template.


leaanthony:
Amazing stuff. Thanks!

PR #1652, opened 2022-07-25
Feature/add support for server and hybrid

tmclane:
This is a first pass implementation of a Wails target type of `server` and `hybrid`.

    wails build -outputType hybrid
    wails build -outputType server

Both server and hybrid accept an optional `options.Server` argument to define the host and port on which to listen.
The intent was to allow the `hybrid` to enable / disable the server endpoint at runtime but this is not implemented at this time.

## `Server`
- only enables the dev server on the specified host and port (does not open a local window)

## `Hybrid`
- Always opens a local desktop window 
- Enables a remote access option using the devserver functionality on the specified host/port.

Default port is : `3112`

### DIFF
--- a/v2/cmd/wails/internal/commands/build/build.go
+++ b/v2/cmd/wails/internal/commands/build/build.go
@@ -146,6 +146,13 @@ func AddBuildSubcommand(app *clir.Cli, w io.Writer) {
 			}
 		}
 
+		// Validate experimental
+		if slicer.String([]string{"hybrid", "server"}).Contains(outputType) {
+			if !slicer.String(userTags).Contains("exp") {
+				return fmt.Errorf("output type '%s' requires the '-tags exp'", outputType)
+			}
+		}
+
 		// Webview2 installer strategy (download by default)
 		wv2rtstrategy := ""
 		webview2 = strings.ToLower(webview2)
--- a/v2/cmd/wails/internal/commands/build/build.go
+++ b/v2/cmd/wails/internal/commands/build/build.go
@@ -1,0 +1,432 @@
+package build
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"text/tabwriter"
+	"time"
+
+	"github.com/wailsapp/wails/v2/internal/colour"
+	"github.com/wailsapp/wails/v2/internal/project"
+	"github.com/wailsapp/wails/v2/internal/system"
+
+	"github.com/wailsapp/wails/v2/cmd/wails/internal"
+	"github.com/wailsapp/wails/v2/internal/gomod"
+
+	"github.com/leaanthony/clir"
+	"github.com/leaanthony/slicer"
+	"github.com/wailsapp/wails/v2/pkg/clilogger"
+	"github.com/wailsapp/wails/v2/pkg/commands/build"
+)
+
+// AddBuildSubcommand adds the `build` command for the Wails application
+func AddBuildSubcommand(app *clir.Cli, w io.Writer) {
+	command := app.NewSubCommand("build", "Builds the application")
+
+	outputType := "desktop"
+	validTargetTypes := slicer.String([]string{"desktop", "hybrid", "server"})
+
+	command.StringFlag("outputType", fmt.Sprintf("Type of binary %s", validTargetTypes.AsSlice()), &outputType)
+
+	// Setup noPackage flag
+	noPackage := false
+	command.BoolFlag("noPackage", "Skips platform specific packaging", &noPackage)
+
+	compilerCommand := "go"
+	command.StringFlag("compiler", "Use a different go compiler to build, eg go1.15beta1", &compilerCommand)
+
+	skipModTidy := false
+	command.BoolFlag("m", "Skip mod tidy before compile", &skipModTidy)
+
+	compress := false
+	command.BoolFlag("upx", "Compress final binary with UPX (if installed)", &compress)
+
+	compressFlags := ""
+	command.StringFlag("upxflags", "Flags to pass to upx", &compressFlags)
+
+	defaultPlatform := os.Getenv("GOOS")
+	if defaultPlatform == "" {
+		defaultPlatform = runtime.GOOS
+	}
+	defaultArch := os.Getenv("GOARCH")
+	if defaultArch == "" {
+		if system.IsAppleSilicon {
+			defaultArch = "arm64"
+		} else {
+			defaultArch = runtime.GOARCH
+		}
+	}
+	platform := defaultPlatform + "/" + defaultArch
+
+	command.StringFlag("platform", "Platform to target. Comma separate multiple platforms", &platform)
+
+	// Verbosity
+	verbosity := 1
+	command.IntFlag("v", "Verbosity level (0 - silent, 1 - default, 2 - verbose)", &verbosity)
+
+	// ldflags to pass to `go`
+	ldflags := ""
+	command.StringFlag("ldflags", "optional ldflags", &ldflags)
+
+	// tags to pass to `go`
+	tags := ""
+	command.StringFlag("tags", "tags to pass to Go compiler (quoted and space separated)", &tags)
+
+	outputFilename := ""
+	command.StringFlag("o", "Output filename", &outputFilename)
+
+	// Clean build directory
+	cleanBuildDirectory := false
+	command.BoolFlag("clean", "Clean the build directory before building", &cleanBuildDirectory)
+
+	webview2 := "download"
+	command.StringFlag("webview2", "WebView2 installer strategy: download,embed,browser,error.", &webview2)
+
+	skipFrontend := false
+	command.BoolFlag("s", "Skips building the frontend", &skipFrontend)
+
+	forceBuild := false
+	command.BoolFlag("f", "Force build application", &forceBuild)
+
+	updateGoMod := false
+	command.BoolFlag("u", "Updates go.mod to use the same Wails version as the CLI", &updateGoMod)
+
+	debug := false
+	command.BoolFlag("debug", "Retains debug data in the compiled application", &debug)
+
+	nsis := false
+	command.BoolFlag("nsis", "Generate NSIS installer for Windows", &nsis)
+
+	trimpath := false
+	command.BoolFlag("trimpath", "Remove all file system paths from the resulting executable", &trimpath)
+
+	raceDetector := false
+	command.BoolFlag("race", "Build with Go's race detector", &raceDetector)
+
+	windowsConsole := false
+	command.BoolFlag("windowsconsole", "Keep the console when building for Windows", &windowsConsole)
+
+	dryRun := false
+	command.BoolFlag("dryrun", "Dry run, prints the config for the command that would be executed", &dryRun)
+
+	command.Action(func() error {
+
+		quiet := verbosity == 0
+
+		// Create logger
+		logger := clilogger.New(w)
+		logger.Mute(quiet)
+
+		// Validate output type
+		if !validTargetTypes.Contains(outputType) {
+			return fmt.Errorf("output type '%s' is not valid", outputType)
+		}
+
+		if !quiet {
+			app.PrintBanner()
+		}
+
+		// Lookup compiler path
+		compilerPath, err := exec.LookPath(compilerCommand)
+		if err != nil {
+			return fmt.Errorf("unable to find compiler: %s", compilerCommand)
+		}
+
+		// Tags
+		userTags := []string{}
+		for _, tag := range strings.Split(tags, " ") {
+			thisTag := strings.TrimSpace(tag)
+			if thisTag != "" {
+				userTags = append(userTags, thisTag)
+			}
+		}
+
+		// Webview2 installer strategy (download by default)
+		wv2rtstrategy := ""
+		webview2 = strings.ToLower(webview2)
+		if webview2 != "" {
+			validWV2Runtime := slicer.String([]string{"download", "embed", "browser", "error"})
+			if !validWV2Runtime.Contains(webview2) {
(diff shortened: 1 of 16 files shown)


m29h:
@tmclane I played a bit with your feature but I can't really get it to work. 

My main issue is that an app compiled with `-outputType server`  seems to never call the OnStartup and/or OnDomReady callbacks which takes away opportunity to store the context like I am used to along with all related functionality...

tmclane:
I'll look into that although it leads to a few questions.

- Would you want the OnDomReady to be called each time a new client connects?
- At what point is the `OnStartup` supposed to be called?

tmclane:
I don't expect the server to function exactly the same as the local version since it is multi-user capable by default.  Note that each connection is it's own connection and does not share state automatically either.

m29h:
For my use case of 'server' I need the same functionality of wails such as RPC and events with the only difference that the app runs in the browser and there is no dependency of the app to a desktop window Operating system framework. 

Think an wails app running on a raspberry pi and the user interface operated from any browser in the local network.

I would not have any requirements on shared state between two browser apps. For my use case 99% of time there would be only a single browser connected. I guess the cleanest solution would be if a new instance of every RPC backend app object would be  created on any new browser connection. This would then also start the OnXXXX callback events each time for the new instance.

The probably cleaner way would be if every  wails rpc call would pass a reference to the Caller context on each RPC call. However that seems to be beyond the current way wails works.

tmclane:
You've described my use-case as well.  I'll look into getting the `OnXXX` events to occur.

tmclane:
Thank you for testing!  I'll have an update soon hopefully time permitting.

m29h:
Travis, 
I played around with your branch and committed my current status to [m29/hfeature/add_support_for_server_and_hybrid](https://github.com/m29h/wails/tree/feature/add_support_for_server_and_hybrid)

My main changes are enforcing the null frontend to call the `OnStartup()` callback and I also changed a bit the app_hybrid_server.go because the go.embed asset embedding was apparently broken.

There is a lot of duplicate/redundant code in the app_hybrid_server.go+app_dev.go+app_production.go files that could in principle be restructured given that all of them more-or-less do the same but with slightly different mechanisms.

Functionality wise I would be happy with the version on [m29/hfeature/add_support_for_server_and_hybrid](https://github.com/m29h/wails/tree/feature/add_support_for_server_and_hybrid). 

I even tested some edge-cases like browser/disconnect/reconnect and multiple paralell browser connections, and everything works within my use case.

tmclane:
I completely agree there is a ton of redundancy in the code.  I was trying to go step by step instead of a huge rewrite as the first pass at this.  
I'll be happy to incorporate your change into my branch for now and then will continue to improve it over time.  

I'll kee rebasing until we get it merged to main.

tmclane:
I do feel the null front end will not be needed in the final version.  It was just a workaround to be able to use the existing dev server code.

tmclane:
@m29h Your commit has been included in this branch.

tmclane:
@laoshaw the idea is that the same application would be able to be accessible over the network in a remote client session.  Note that it does not share any state with a local window on the desktop.  Multiple webclients can connect at the same time as well but are not linked together and are basically independent.
I am going to pick this up again in earnest real soon now.

tmclane:
Actually I was redoing this work in a new way.  This may be discarded..

tmclane:
I wasn't planning on adding authentication to the endpoint immediately.   Sounds like that needs to be put on the roadmap as well.

alienator88:
Oh sorry, I might've explained wrong. No auth needed from the wails side for my use case at least. Just the CLI tool we use internally that I mentioned in the example :)

ganeshkbhat:
Did this finally work. I am looking to run a localhost myself. https://github.com/wailsapp/wails/issues/2189

m29h:
I wanted to bump this thread as I quite extensively worked on this topic in my own repository.

Basically I decoupled what is called "devserver" and "desktop" frontends in the master branch.
In my reimplementation all communications including such edge cases like reload in `dev` mode are completely running though the `messageDispatcher` interface. One now can (theoretically) create a slice of multiple `frontend` interfaces in App.frontend that can practically run independently.

The implementation also minimizes code duplication as it uses a single code for `server` and/or `desktop` frontend and dev+hybrid combination mode are now simply a linear combination of the same basic `server`+`desktop` frontends.

Feel free to have a look [m29h/feature/add_support_for_server_and_hybrid](https://github.com/m29h/wails/tree/feature/add_support_for_server_and_hybrid). I would be happy make a pull request if you consider my approach useful.

tmclane:
Howdy!  That is basically the approach I was taking but I haven't finished what I was doing. I'll check out your version.  I bet it would be useful to the project.  Obviously I haven't been very active here in a while although I still am interested in contributing more in the near future.

tmclane:
@leaanthony Is there any benefit to attempting to get this finished up and brought in to v2? 
@m29h 's work looks good to me.  I wanted to also rewrite the devserver itself since it seems a bit messy BUT it does function as it is.

leaanthony:
My only concern with a server build right now will be the inevitable requests for extending it with middleware and all the functionality people have when deploying server apps with a standard webserver. I'm very keen to get a server target, I just think we need to get a few people involved in that conversation and determine how best to do it. For instance, if we were able to generate an app as a Caddy plugin, then you'd get all the middleware and auth etc for free.

tmclane:
I have no issue with using Caddy in the future.  

I think the best suggestion for the moment is to use a  reverse proxy that `provides` the authentication layer such that one wasn't need at the application level.   This would allow this feature to be added with the caveat that there is ZERO authentication present at this time.

m29h:
I agree with @tmclane that a reverse proxy would suffice for all authentication needs. (if authentication would even needed for a particular app)
Practically the benefit of such a wails "server" app lies in the rapid development possibilities due to the wails internal tooling for hot reload + automated bindings generation. Given that the multiple browser connections would all share the same App internal state, doing any sophisticated user management and authentications is probably not the primary use case.

However I see many uses for a remote browser accessible app on a local trusted network where all browser access instances share the same app state. This could for example be:
- a local system app running on a 127.0.0.1 port for all users of the local machine. (persisting between logins)
- some kind of home automation app running on a trusted network
- Human-Machine-Interfaces, etc that benefit from the low level programming possibilities of the Go backend

The code draft I propose a week ago also would not introduce any significant amount of potentially "stale" code that is exclusively for the server use-case as it shares this server code with the current devserver. The idea is that it is more of a generalization of mechanisms that are already there today.

mholt:
Let me know how I can help, if needed!

radoslaw-szpiech:
Is there any progress in this PR? It could be really useful in one of our projects at [VEGA](https://github.com/vegaprotocol/vegawallet-desktop). We are experiencing some issues when testing our wallet app, which we believe my be caused by desktop app being run in the same time we're performing tests in browser.

leaanthony:
No progress. If you find a solution that works for you, it would be great to hear what you did.

suntong:
> I have no issue with using Caddy in the future.

Agree -- if we cannot go there in one step, then make that a send step maybe is a good idea -- Releasing the feature first now, it will draw more people / talent over here and things will surely get easier, especially Caddy author himself had offered help.

> The code draft I propose a week ago also would not introduce any significant amount of potentially "stale" code that is exclusively for the server use-case as it shares this server code with the current devserver. The idea is that it is more of a generalization of mechanisms that are already there today.

This is a really reasonable approach, and mature enough to be merged IMHO, as _"This branch is 300 commits ahead"_. The longer to wait and the more difficult to merge _maybe_. 

So please, wrap up what we have now so that our application can be either be a local desktop GUI program and a cloud-based solution _now_.

Please consider. thanks

tmclane:
300 commits ahead is merely a rebase away to being mostly gone :).

suntong:
Oh... ☺️☺️☺️

tmclane:
I may have more time to resume my work on this soon.  Life has been hectic to say the least for me this summer.

mholt:
This probably isn't helpful as it could be incredibly niche, but just in case it spurs any ideas...

I have recently implemented an HTTP server for my Wails app in such a way that its core application logic (the app bindings, basically -- the functions that can be called by the frontend) can be accessed these ways:

- JS function calls using Wails runtime/bindings
- HTTP JSON API
- CLI

Basically, I still have an `App` struct. And if there's a function with this signature:

```go
func (a *App) DoSomething(input MyStruct) (Result, error) {
   ...
}
```

then it can still be called by the JS frontend via Wails bindings like normal:

```js
import { DoSomething } from "../wailsjs/go/main/App.js";

const result = await DoSomething({
    "foo": "bar",
    "val": 3.14
});
```

Now, my app also has a `Server` struct inside it. The server is configured with a listener, maybe with some authentication, whatever you want. That's up to you. The key is its Handler, an `*http.ServeMux`. When I initialize the app, I add a bunch of routes to its mux, one per app function:

```go
a.server.mux = http.NewServeMux()

// SPA static file server
a.server.staticFiles = http.FileServer(http.FS(a.frontend))
a.server.mux.Handle("/", Endpoint{
	Method:  http.MethodGet,
	Handler: a.server.serveFrontend,
})

// API endpoints
for command, endpoint := range a.commands {
	a.server.mux.Handle("/api/"+command, endpoint)
}
```

What is `a.commands`? It's simply a mapping of function names (aka. API endpoints) to info about that endpoint:

```go
commands map[string]Endpoint
```

The key of this map is the last component of the API URI, e.g.`do-something` for a full URI of `/api/do-something`, and `Endpoint` is:

```go
type Endpoint struct {
	Method      string
	ContentType string
	Payload     any // example of the input
	Handler     http.Handler
	Help        string // help text for CLI
}
```

For example:

```
a.commands = map[string]Endpoint{
	"do-something": {
		Handler:     a.server.handleDoSomething,
		ContentType: "application/json",
		Method:      http.MethodPost,
		Payload:     MyStruct{},
		Help:        "Does something!",
	},
...
```

The `Endpoint` type has a `ServeHTTP()` method so it is itself a handler, which decodes the payload for convenience of the _actual_ handler about to be called (`handleDoSomething` in the example above):

```
func (e Endpoint) ServeHTTP(w http.ResponseWriter, r *http.Request) {
	switch e.ContentType {
	case JSON:
		payload := reflect.New(reflect.TypeOf(e.Payload)).Interface()
		if r.ContentLength > 0 {
			err := json.NewDecoder(r.Body).Decode(&payload)
			// (handle err)
		}
		r = r.WithContext(context.WithValue(r.Context(), "payload", payload))
	}

	e.Handler(w, r)
}
```

Then `handleDoSomething` is this:

```



For context: I actually had to stop using Wails for my app because my UI has pretty intensive stuff going on that Webkit2GTK chokes on: canvas, webgl, lots of media, videos, etc -- Wails itself is great but webkit2gtk keeps crashing and having numerous other issues, video driver problems, etc. So anyway, that's my motivation for working on this independently; I don't _strictly_ need a hybrid model per-se, but my solution is close to one.)

**Comment by suntong** [comment]
Bravo @mholt!

> HTTP, CLI, and GUI/JS access to business logic and stay close/attached to the App itself, and are also congruent/symmetrical. 

This is exactly what I'll be needing, and I'm sure a lot of people would find it handy too. Thanks a lot for sharing!!

**Comment by FrenchGithubUser** [comment]
Yes this will breat a great addition !

**Comment by flarco** [comment]
hi @leaanthony any eta on this?

**Comment by tmclane** [comment]
This isn't planned for inclusion at this time in v2. 

This is left in the repo so that industrious users can leverage it and expand on it but doesn't really have official support.  That being said I can rebase it sometime and push it up again and then you could update your reference to use it.

**Comment by domingguss** [comment]
so, I need this functionality - how can i switch my current wails setup to this hybrid branch? should i just fork it?

**Comment by leaanthony** [comment]
For v2, that's probably your best option at this point.

**Comment by tmclane** [comment]
Hit me up on the discord server for Wails.
I'll rebase and force push again soon and then drop instructions here.

**Comment by askie** [comment]
why this code not merged in v2.9 ?

**Comment by domingguss** [comment]
So, I have forked the branch, and used `replace github.com/wailsapp/wails/v2 => /Users/dominggus/wails-custom/v2` in the `go.mod` file.
Then I tried to build a hybrid using `wails build -tags exp,hybrid`
but it returns the following error:

```sh
  • Generating bindings:   ERROR   
          # github.com/wailsapp/wails/v2/internal/frontend/devserver
          ../../../../wails-custom/v2/internal/frontend/devserver/devserver.go:100:34: undefined: assetserver.NewDevAssetServer
          
          exit status 1

  ERROR   
          # github.com/wailsapp/wails/v2/internal/frontend/devserver
          ../../../../wails-custom/v2/internal/frontend/devserver/devserver.go:100:34: undefined: assetserver.NewDevAssetServer
          
          exit status 1
```
So how am i supposed to build a hybrid?

yarcowang:
I'd say this is a good feature. 

* Wails may be used as a standalone desktop app (so, it doesn't need an embed web server.)
* But in some other case, let's say, I'd like to type `<some_wails_app> -d`, and the `<some_wails_app>` could act as an online service, then I can visit it through browser. (this web server should be very customizable and reusable, so that if I want to add another port to listen and handle different business logic, it could also work -- I don't need to add another http server lib)

tmclane:
The intent of this branch was to make the local web application running on the host to be accessible in the same exact manner over a web browser. It was not intended to expose any arbitrary API that is easily done if you wish it to be done, however, because you can stand up your own a web server .

The branch is so old that I do not think it is easily upgraded to the latest V2. If I have some time, I may look into it again.

leaanthony:
Hey @tmclane - are you ok if we close this? I think we can revisit this in the context of v3?

leaanthony:
This PR is from July 2022 (nearly 4 years old) and the codebase has changed significantly since then. The  and  build types would need to be re-implemented against the current codebase. Please open a new issue if you'd like to discuss this feature for the current version of Wails.

PR #1662, opened 2022-07-26
Windows dialogs - Handle scenario when window is not visible yet.

### LINKED ISSUE #1660: [Windows] runtime.MessageDialog panics when called from "OnStartup" callback
### Description

When trying to show a message dialog in the wails app `OnStartup` callback on *Windows*, the app panics while showing the message. 

### To Reproduce

1. Create new wails project on Windows (vanilla, but actually any frontend style will do here)
2. Modify `App.startup` to open a message dialog: see example below
3. Build and run the app (debug or release or live dev mode doesn't matter)
4. The message shortly is visible, then the app panics in `runtime.MessageDialog` and is gone

Example OnStartup callback:
```
func (a *App) startup(ctx context.Context) {
	a.ctx = ctx
	messageOptions := runtime.MessageDialogOptions{
		Type:    "info",
		Title:   "Some Title",
		Message: "Some Message",
		Buttons: []string{"Okay"},
	}
	_, err := runtime.MessageDialog(a.ctx, messageOptions)
	if err != nil {
		fmt.Fprintf(os.Stderr, "Failed to show error message: %s\n", err.Error())
	}
}
``` 

Please Note: This  happens on *Windows only*. On macOS and Linux the message is showing up just fine.
Also showing the message in `OnDomReady` works fine on Windows, which I'm currently using as a workaround. 

The crash seems to happen in `func (f *Frontend) MessageDialog(options frontend.MessageDialogOptions) (string, error)` at `internal\frontend\desktop\windows\dialog.go`. 

Only `MessageDialog` seems affected. Save/Open File/Directory Dialogs are working fine.

### Expected behaviour

It should be possible to show a system message dialog on Windows in OnStartup.

### Screenshots

_No response_

### Attempted Fixes

_No response_

### System Details

```shell
Wails CLI v2.0.0-beta.42

System
------
OS:             Windows 10 Pro
Version:        2009 (Build: 19044)
ID:             21H2
Go Version:     go1.18.3
Platform:       windows
Architecture:   amd64

Wails
------
Version:        v2.0.0-beta.42

WebView2        N/A             Installed       103.0.1264.71
npm             N/A             Installed       6.14.15
*upx            N/A             Available
*nsis           N/A             Available
```

### Additional context

_No response_

leaanthony:
On Windows, if the window has not yet been activated, calls to dialog methods will crash due to some weirdness deep inside a windows syscall. This PR will use the desktop (nil HWND) if the window is not visible. Fixes #1660

### DIFF
--- a/v2/internal/frontend/desktop/windows/dialog.go
+++ b/v2/internal/frontend/desktop/windows/dialog.go
@@ -135,7 +135,13 @@ func (f *Frontend) MessageDialog(options frontend.MessageDialogOptions) (string,
 		flags = windows.MB_OK | windows.MB_ICONWARNING
 	}
 
-	button, _ := windows.MessageBox(windows.HWND(f.mainWindow.Handle()), message, title, flags|windows.MB_SYSTEMMODAL)
+	// Check if window is visible or not
+	handle := f.mainWindow.Handle()
+	if !f.mainWindow.IsVisible() {
+		handle = 0
+	}
+
+	button, _ := windows.MessageBox(windows.HWND(handle), message, title, flags|windows.MB_SYSTEMMODAL)
 	// This maps MessageBox return values to strings
 	responses := []string{"", "Ok", "Cancel", "Abort", "Retry", "Ignore", "Yes", "No", "", "", "Try Again", "Continue"}
 	result := "Error"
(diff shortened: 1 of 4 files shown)


stffabi (approved):
LGTM 👍

PR #1666, opened 2022-07-27
Add `frontend:dev:install` configuration

### LINKED ISSUE #1663: The `frontend:dev` configuration is used to both install dev dependencies and build the frontend
### Description

PR #1615 introduces the `GetDevInstallerCommand()` method to get the installation command of dev dependencies:
https://github.com/wailsapp/wails/blob/b9882eabe226c9f2cce0acbe910d3e7ebf705378/v2/internal/project/project.go#L87-L92
This method returns the `p.DevCommand` when the `p.DevCommand` is not empty, the corresponding configuration item of the `DevCommand` in wails.json is `frontend:dev`, in the [documentation](https://wails.io/docs/next/reference/project-config) this item is used to build assets in development mode, not for specify the dev mode dependencies installation command.

And in v2/pkg/commands/build/base.go, it use the same `DevCommand` to build frontend:
https://github.com/wailsapp/wails/blob/b9882eabe226c9f2cce0acbe910d3e7ebf705378/v2/pkg/commands/build/base.go#L535-L539

Perhaps we need to add a new configuration item in wails.json called 'frontend:dev:install' to specify the command to install the dev dependencies?

### To Reproduce

Nil

### Expected behaviour

Nil

### Screenshots

_No response_

### Attempted Fixes

_No response_

### System Details

```shell
Nil
```

### Additional context

_No response_

LGiki:
Add a `frontend:dev:install` configuration in wails.json to specify the dev dependencies installation command and update the docs.
Fixes #1663

### DIFF
--- a/v2/internal/project/project.go
+++ b/v2/internal/project/project.go
@@ -22,6 +22,7 @@ type Project struct {
 
 	// Commands used in `wails dev`
 	DevCommand        string `json:"frontend:dev"`
+	DevInstallCommand string `json:"frontend:dev:install"`
 	DevWatcherCommand string `json:"frontend:dev:watcher"`
 	// The url of the external wails dev server. If this is set, this server is used for the frontend. Default ""
 	FrontendDevServerURL string `json:"frontend:dev:serverUrl"`
@@ -85,8 +86,8 @@ type Project struct {
 }
 
 func (p *Project) GetDevInstallerCommand() string {
-	if p.DevCommand != "" {
-		return p.DevCommand
+	if p.DevInstallCommand != "" {
+		return p.DevInstallCommand
 	}
 	return p.InstallCommand
 }
--- a/website/docs/reference/project-config.mdx
+++ b/website/docs/reference/project-config.mdx
@@ -14,6 +14,7 @@ The project config resides in the `wails.json` file in the project directory. Th
 	"frontend:install": "[The command to install node dependencies, run in the frontend directory - often `npm install`]",
 	"frontend:build": "[The command to build the assets, run in the frontend directory - often `npm run build`]",
 	"frontend:dev": "[This command is the dev equivalent of frontend:build. If not specified falls back to frontend:build]",
+	"frontend:dev:install": "[This command is the dev equivalent of frontend:install. If not specified falls back to frontend:install]",
 	"frontend:dev:watcher": "[This command is run in a separate process on `wails dev`. Useful for 3rd party watchers or starting 3d party dev servers]",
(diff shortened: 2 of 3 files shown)


leaanthony:
This looks good to me! If @stffabi is happy, let's go!

stffabi:
Thanks @LGiki for this. LGTM, there's just one thing which we might also need to change.

During the build process we also execute an install command before the frontend is getting build. Currently we use the `frontend:install` even when the build is started for a `wails dev` build. This feels somehow inconsistent when `wails dev` first would use `frontend:dev:install` (if it has been defined) but then the build process for the frontend would execute `frontend:install`.
I would propose to also use the same logic here:
https://github.com/wailsapp/wails/blob/d5574ba9d78740374f44691077fd9381b7113745/v2/pkg/commands/build/base.go#L516-L519

What do you think @leaanthony ?

leaanthony:
Yeah, good call. @LGiki this is what we do for the build command: https://github.com/wailsapp/wails/blob/d5574ba9d78740374f44691077fd9381b7113745/v2/pkg/commands/build/base.go#L533-L547

leaanthony:
I'm thinking about how we can also change `frontend:dev` to `frontend:dev:build` for consistency. Maybe support both for a bit?

LGiki:
Thanks, I've pushed a new commit that uses the same logic to handle the install command when building the frontend if it's called for a `dev` build. And it might be better if I create a new PR to change `frontend:dev` to `frontend:dev:build`?

v2/pkg/commands/build/base.go
```diff
@@ -513,17 +513,27 @@ func (b *BaseBuilder) BuildFrontend(outputLogger *clilogger.CLILogger) error {
 	frontendDir := filepath.Join(b.projectData.Path, "frontend")
 
 	// Check there is an 'InstallCommand' provided in wails.json
-	if b.projectData.InstallCommand == "" {
+	var installCommand string
```
leaanthony, line 516: I think this can be simplified with the helper method you created! 🚀

leaanthony (changes requested):
Looking good! Just one comment 👍👍👍

leaanthony:
> Thanks, I've pushed a new commit that uses the same logic to handle the install command when building the frontend if it's called for a `dev` build. And it might be better if I create a new PR to change `frontend:dev` to `frontend:dev:build`?

That'd be great if you could! I would do something like we've done already and check `frontend:dev:build` but drop back to `frontend:dev` for compatibility reasons.

v2/pkg/commands/build/base.go
LGiki: 🚀 Thanks, I've simplified the logic of determining the installation command.

stffabi (approved):
LGTM 👍 , thanks so much for the time you have put into this.

PR #1673, opened 2022-07-28
[v2, darwin] Use kern.osversion for os id in doctor

stffabi:
It seems like kern.osrevision is always 199506 since several years
and kern.osversion maps to the official documented build id.

Example: MacOS 12.5 with ID 21G72


leaanthony (approved):
LGTM

PR #1691, opened 2022-08-03
Change `frontend:dev` to `frontend:dev:build`

LGiki:
Change `frontend:dev` to `frontend:dev:build` and update the docs.
For more information, see https://github.com/wailsapp/wails/pull/1666#issuecomment-1200995636.


v2/internal/project/project.go
```diff
@@ -85,6 +86,16 @@ type Project struct {
 	NSISType string `json:"nsisType"`
 }
 
+func (p *Project) GetDevBuildCommand() string {
+	if p.DevBuildCommand != "" {
+		return p.DevBuildCommand
+	}
+	if p.DevCommand != "" {
+		return p.DevCommand
```
leaanthony, line 94: Do you think it might be worth doing a `println("Warning: frontend:dev has been deprecated, please use frontend:dev:build")`? I think the templates will need updating too.

leaanthony (approved):

v2/internal/project/project.go
leaanthony: We can do that at a later date 👍

PR #1704, opened 2022-08-06
Attempted fix for both drag and doubleclick

### LINKED ISSUE #1703: [v2] delay when dragging a window with `data-wails-drag` attribute
### Description

After [fix](https://github.com/wailsapp/wails/issues/1313) ([commit](https://github.com/wailsapp/wails/commit/e994a6f8c4c170900e520e58b820ecee28035650)), when [dragging a window using 3 fingers](https://support.apple.com/en-gb/HT204609), the window is not dragged immediately, but after a delay.
When I paste the old code into the console (after 12 seconds in the video), there are no such problems.

### To Reproduce

Dragging a window using 3 fingers on MacOS trackpad.

### Expected behaviour

Move window immediately.

### Screenshots

https://user-images.githubusercontent.com/12181586/183136483-057d3fb2-767b-4515-921a-c2d505732e2d.mov

### Attempted Fixes

Use old code.

### System Details

```shell
$ wails doctor
Wails CLI v2.0.0-beta.42

System
------
OS:		MacOS
Version: 	12.4
ID:		199506
Go Version:	go1.18.5
Platform:	darwin
Architecture:	arm64

Wails
------
Version: 	v2.0.0-beta.42

xcode command line tools 	N/A		Installed	2395
npm 				N/A		Installed	8.12.1
*upx 				N/A		Available
*nsis 				N/A		Available

* - Optional Dependency

Diagnosis
---------
```

### Additional context

_No response_

leaanthony:
Fixes #1703 

Working on:
- [x] MacOS
- [x] Windows
- [x] Linux


stffabi:
I've just tested MacOS and Windows.

MacOS: Everything works fine, dragging works and double click events on a `data-wails-drag` also works as expected.
Windows: Dragging works as expected without any delay, but double click events on a `data-wails-drag` don't get triggered anymore.

leaanthony:
Ah thanks for trying 🙏 I wonder why webview 2 would be so different around this? I'll test Linux tonight and think about a windows fix.

stffabi:
I suspect it's because of the Win32 ReleaseCapture call in startDrag.

leaanthony:
Can confirm it is the drag that is preventing the double click events from propagating.

leaanthony:
I've just pushed a shameful hack that does appear to fix it, though I suspect it wont work on all machines in all scenarios 😱

ghostiam:
@leaanthony I checked on Windows, there is the same problem as it was on MacOS. Without this hack (but with doubleclick not working), there is no delay when moving the window.

leaanthony:
Without rewriting the whole of the move window code (and I did try tonight) I'm not sure what we can do right now

ghostiam:
On Windows, with the hack I sometimes can't right click to open the inspector and check which events are running. This highlights the elements, as if I'm double-clicking, but I'm only pressing the right mouse button.

https://user-images.githubusercontent.com/12181586/183246991-6ef6c7a0-c905-4512-b969-b8134e362aad.mov

ghostiam:
I don't think it's worth using move window anywhere. In practice, this is rarely used. Perhaps it should be documented that there is such a behavior as not passing events to child elements.

ghostiam:
> On Windows, with the hack I sometimes can't right click to open the inspector and check which events are running. This highlights the elements, as if I'm double-clicking, but I'm only pressing the right mouse button.
> 
>  2022-08-06.20.27.03.mov

I found the reason for the quick closing of the context menu. When pressing with two fingers on the touchpad, the events of pressing and releasing occur very quickly, if I make a click, then enough time elapses that this bug does not exist.

leaanthony:
Thanks for your testing. I think we'll just change how windows implements drag. Hacks are too flaky 😁

leaanthony:
I've tried all different approaches: JS only solution that only sends a drag message if the button is down AND you're moving, and also a Go-side solution that basically does the same. It appears the same issue occurs: There's no way to detect the mouse button up event and therefore reset the drag flag. This leads to some odd behaviour where sometimes it drags, sometimes it's delayed, sometimes it works just fine. If webview2 forwarded mouse events to the host then we'd be ok but it doesn't. I've posted a couple of messages to the WebView2 feedback around this so lets see what comes back.

leaanthony:
Just pushed a potential fix. Seems to work well for Windows. Not tried the others but it shouldn't affect it.

leaanthony:
@ghostiam how does this work for you?

ghostiam:
@leaanthony checked, everything works well, events are working on windows, window dragging also works well.
On MacOS, everything is fine too.

stffabi:
I can also confirm that it works as expected on MacOS and Windows within our application. Thanks @leaanthony for looking into this 🙏

PR #1727, opened 2022-08-10
Fix registration of exposed fields

### LINKED ISSUE #1721: Shouldn't we register exported field during binding ?
### Description

I encounter a problem while generating the binding for response object with nested structure.

```go
type As struct {
  B Bs `json:"b"`
}

type Bs struct {
  Name string `json:"name"`
}
```

It appears the struct `Bs` is not added to the struct to generate for typescript, as we end up with an `any` type in the TS models.

I think the problem comes from these lines:
- https://github.com/wailsapp/wails/blob/4385fea49b571b53b03c9cf57983f2113af383bd/v2/internal/binding/binding.go#L159-L161
- https://github.com/wailsapp/wails/blob/4385fea49b571b53b03c9cf57983f2113af383bd/v2/internal/binding/binding.go#L171-L173

In this case, validating the condition `field.PkgPath == ""` means we found an exported field. So, we are skipping the exported fields, which, I think, is not what we want.

We could use `!field.IsExported()` instead. That would be less cryptic. It has been added recently, AFAIR.

### To Reproduce

Not sure if this is accurate, but _I think_ the example above could reproduce the bug.

```go
type As struct {
  B Bs `json:"b"`
}

type Bs struct {
  Name string `json:"name"`
}
```

### Expected behaviour

The binding process should register nested struct in the TS models, not an `any`.

### Screenshots

Screenshot 2022-08-09 at 22 51 03

### Attempted Fixes

The following change worked for me. But I don't have the full picture so, it might not be the complete solution.

```diff
- if field.PkgPath == "" {
+ if field.PkgPath != "" {
    continue
}
```

### System Details

```shell
Wails CLI v2.0.0-beta.42

System
------
OS:		MacOS
Version: 	12.5
ID:		199506
Go Version:	go1.18.4
Platform:	darwin
Architecture:	arm64

Wails
------
Version: 	v2.0.0-beta.42

xcode command line tools 	N/A		Installed	2395
npm 				N/A		Installed	8.1.2
*upx 				N/A		Available	
*nsis 				N/A		Available	

* - Optional Dependency

Diagnosis
---------
```
```


### Additional context

_No response_

ValentinTrinque:
Close #1721

### DIFF
--- a/v2/internal/binding/binding.go
+++ b/v2/internal/binding/binding.go
@@ -156,7 +156,7 @@ func (b *Bindings) AddStructToGenerateTS(packageName string, structName string, 
 		}
 		kind := field.Type.Kind()
 		if kind == reflect.Struct {
-			if field.PkgPath == "" {
+			if !field.IsExported() {
 				continue
 			}
 			fqname := field.Type.String()
@@ -168,7 +168,7 @@ func (b *Bindings) AddStructToGenerateTS(packageName string, structName string, 
 				b.AddStructToGenerateTS(pName, sName, s)
 			}
 		} else if kind == reflect.Ptr && field.Type.Elem().Kind() == reflect.Struct {
-			if field.PkgPath == "" {
+			if !field.IsExported() {
 				continue
 			}
 			fqname := field.Type.String()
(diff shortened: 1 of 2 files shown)


leaanthony:
LGTM! Thanks for raising this PR 👍

PR #1732, opened 2022-08-11
Simplify drag window logic

brys0:
Simplifies and removes alot of drag logic to speed up data-wails-drag ten-fold. 

Removes:
enableResize
defaultCursor


leaanthony (commented):
Hi! Appreciate what you're trying to do here. We can't drop support for resizing frameless windows but I think we can do better at only executing code we need to based on the situation. Will this code allow drag if the parent element has the drag data attribute?

brys0:
Hi Anthony, yes it should. I will work more on making it a drop-in solution, for now I just use it personally.

leaanthony:
Closing in favour of #1750

PR #1750, opened 2022-08-15
Experimental support for CSS Drag detection

leaanthony:
This PR adds experimental support for CSS defined drag regions using the `--wails-drag` CSS variable. To test:

- Add style property to draggable element

```html
<body>
    <div id="root" style="--wails-draggable: drag"></div>
    <script src="./src/main.jsx" type="module"></script>
</body>
```

- Compile app in debug mode
```sh
    wails build -debug
```
- Open dev console and run `wails.useCSSDrag()` 

image

Drag type can be toggled so you can test both original and new drag testers:

```html
<body>
    <div id="root" data-wails-drag style="--wails-draggable: drag"></div>
    <script src="./src/main.jsx" type="module"></script>
</body>
```
To specify a non-draggable region within a draggable region, use any value other than `drag` for the css variable.


brys0:
Hey! Sorry to get back to you so late. That *does* seem to work

leaanthony:
Thanks for testing @brys0 !  I've added a small update that will enable this feature via an experimental option:

```go
// Create application with options
	err := wails.Run(&options.App{
		Title:            "reactsprojec",
		Width:            1024,
		Height:           768,
		Assets:           assets,
		BackgroundColour: &options.RGBA{R: 27, G: 38, B: 54, A: 1},
		OnStartup:        app.startup,
		Bind: []interface{}{
			app,
		},
		Experimental: &options.Experimental{
			UseCSSDrag: true,
		},
	})
```
For now it still outputs a message to the console indicating which drag method it's using. If this works, I'll update docs and merge 👍

leaanthony:
Merging this experimental flag as a means for people to test css drag before deprecating the current way

brys0:
Hey! So, there's an actually with the current CSS draggable implementation. 

In vue if you try to make the style dynamic i.e :style="`--wails-draggable: drag;`" it doesnt work. It also doesnt work with *another* css style applied. It however, does work with just no dynamic templating i.e style="--wails-draggable: drag;"

brys0:
I suspect this is also issues on other dynamically generated template. Like Svelte, or React as well.

leaanthony:
I think that's fine. The requirement is to have that style there in whatever way you want. If it isn't there, the CSS selector won't work.

brys0:
So, it's the intended effect not to work with dynamic styling?

leaanthony:
That's not the intent but it is a side effect. Applying it directly works though. We could change it to a CSS property but there isn't a standard one and in my testing, the ones commonly used weren't always picked up by the CSS selector. Happy to also add a CSS property as a trigger if you can find one that works cross platform and doesn't override any intended behaviour of the attribute. Could also make it user configurable 🤔 that sounds awesome 😃

PR #1764, opened 2022-08-18
[docs] Link to general webview2 runtime download page and not to a specific language

stffabi:


leaanthony:
Never saw the ru-ru before 😬

PR #1765, opened 2022-08-18
Text outside style's brackets

DragoSpiro98:
I deleted some text that was written outside the img tag

image


leaanthony:
Ah, thanks for this. Crowdin destroyed this style all over the docs. The fix is to delete the style and use `className='screenshot'`. If you have time to update then cool otherwise I'll fix on master

PR #1768, opened 2022-08-20
doc: added minecraft updater as showcase

Gurkengewuerz:
I don't want to deny the community our Minecraft Updater. Here you can get inspiration for React, React-Router or even cross-platform compiling.


leaanthony:
Nice! Thanks!

PR #1783, opened 2022-08-23
Add README.ja.md

shinshin86:
Hello!

I have recently discovered and touched Wails. And I am experiencing that Wails are very cool!

I created a Japanese version of the README to let people in also Japanese know more about it.

P.S.
I am creating this PR while listening to Resistance Is Futile by Manic Street Preachers as described in the README.
This music is also very good.


leaanthony (commented):
Thanks so much for taking the time to do this 🙏

PR #1793, opened 2022-08-26
Add garble support

### LINKED ISSUE #1634: Support integers for method call IDs

garble scrambles strings and this breaks the callbacks to the backend

Use integers to represent the call IDs back to Go

leaanthony:
This PR adds support for the garble obfuscation tool. It generates different bindings in the `wailsjs` directory to support this.

Fixes #1634 

CC @osbornnnnn @aivaki


v2/cmd/wails/internal/commands/build/build.go
```diff
@@ -111,9 +111,18 @@ func AddBuildSubcommand(app *clir.Cli, w io.Writer) {
 	windowsConsole := false
 	command.BoolFlag("windowsconsole", "Keep the console when building for Windows", &windowsConsole)
 
+	obfuscate := false
+	command.BoolFlag("obfuscate", "Obfuscate bound methods", &obfuscate)
```
Snider, line 115: Hiya, maybe this is easier to read for non-native English speakers?

"Code obfuscation of bound Wails methods"

v2/cmd/wails/internal/commands/dev/dev.go
```diff
@@ -119,6 +120,7 @@ func AddSubcommand(app *clir.Cli, w io.Writer) error {
 	command.BoolFlag("save", "Save given flags as defaults", &flags.saveConfig)
 	command.BoolFlag("race", "Build with Go's race detector", &flags.raceDetector)
 	command.BoolFlag("s", "Skips building the frontend", &flags.skipFrontend)
+	command.BoolFlag("obfuscate", "Obfuscate bound methods", &flags.obfuscate)
```
Snider, line 123: Same as the last comment for the text?

v2/cmd/wails/internal/commands/dev/dev.go
```diff
@@ -167,10 +169,17 @@ func AddSubcommand(app *clir.Cli, w io.Writer) error {
 
 		if !flags.noGen {
 			self := os.Args[0]
+			var env []string
+			if flags.obfuscate {
```
Snider, line 173: This needs to check for the exact type vs a truth value.

v2/internal/app/app_dev.go
```diff
@@ -73,12 +73,18 @@ func CreateApp(appoptions *options.App) (*App, error) {
...
 	var loglevelFlag *string
+	var obfuscateFlag *bool
 
 	assetdir := os.Getenv("assetdir")
 	if assetdir == "" {
 		assetdirFlag = devFlags.String("assetdir", "", "Directory to serve assets")
 	}
 
+	obfuscate := os.Getenv("WAILS_OBFUSCATE")
+	if obfuscate == "" {
```
Snider, line 84: Could this be a boolean? feels better.

v2/cmd/wails/internal/commands/dev/dev.go
leaanthony: That's a boolean so we're good 👍

v2/internal/app/app_dev.go
leaanthony: Can't. It's an environment variable = string

v2/internal/frontend/desktop/windows/frontend.go
```diff
@@ -91,12 +91,18 @@ func NewFrontend(ctx context.Context, appoptions *options.App, myLogger *logger.
 		return result
 	}
 
-	bindingsJSON, err := appBindings.ToJSON()
-	if err != nil {
-		log.Fatal(err)
+	var bindings string
```
Snider, line 94: I'm not up with Go var handling; what happens if `bindings` never get set a value, just a type, then line 105 runs?

v2/internal/frontend/dispatcher/dispatcher.go
```diff
@@ -38,6 +38,8 @@ func (d *Dispatcher) ProcessMessage(message string, sender frontend.Frontend) (s
 		return d.processEventMessage(message, sender)
 	case 'C':
 		return d.processCallMessage(message, sender)
+	case 'c':
```
Snider, line 41: maybe a future change, but could "S" become "D" for display (head out the gutter) letting "S" mean "Secure"

I can see you having future issues with support over "s" and "S"

v2/pkg/commands/build/base.go
```diff
@@ -165,8 +165,22 @@ func (b *BaseBuilder) CompileProject(options *Options) error {
 		}
 	}
 
+	commands := slicer.String()
+
+	compiler := options.Compiler
+	if options.Obfuscate {
+		if !shell.CommandExists("garble") {
+			options.Logger.Print("garble not found. Please install it with `go install mvdan.cc/garble@latest`. Using Go compiler instead.")
```
Snider, line 173: If in dev mode, could you run that cmd on the host? 

"garble not found, Dev Mode installation, please wait.."
"Installing: `go install mvdan.cc/garble@latest`"

It will save you + developers time :)

Snider (approved):
Lookin goooood

v2/cmd/wails/internal/commands/dev/dev.go
Snider: Nice, nice

v2/internal/app/app_dev.go
Snider: Can it be "1" or "0" as the string values? Or am I showing my GoLang noob status ^.^

Snider (approved):
Nice :)

Snider (approved):

v2/internal/app/app_dev.go
leaanthony: It can be anything you like. We just check it's set.

v2/pkg/commands/build/base.go
leaanthony: We have a policy of not executing installation code - makes people (rightly) nervous. Also can get tricky with permissions errors leading to a ton of tickets. We just give pointers :-)

v2/internal/frontend/desktop/windows/frontend.go
leaanthony: yes. But that's fine

Aivaki:
> CC @Osbornnnnn @Aivaki Looking for your feedback 👍

Sorry I didn't answer for a long time. It's all cool, but I have no idea how to test it right now (how do I get a CLI?)

leaanthony:
Please try the above PR to see if this fixes your problem. If you are unsure how to do this, please follow [this guide](https://wails.io/docs/guides/bleeding-edge#testing-a-pr). Let us know how you get on 👍

Aivaki:
> Please try the above PR to see if this fixes your problem. If you are unsure how to do this, please follow [this guide](https://wails.io/docs/guides/bleeding-edge#testing-a-pr). Let us know how you get on 👍

I have installed the CLI, but...
```
wails build -race  -trimpath -obfuscate -m -tags="desktop,production" -compiler="garble" -garbleargs="-tiny -literals -seed=random" -ldflags="-w -s -H windowsgui" 
Wails CLI v2.0.0-beta.44.2

App Type:               desktop
Platforms:              windows/amd64
Compiler:               C:\Users\makse\go\bin\garble.exe
Build Mode:             Production
Obfuscate:              true
Skip Frontend:          false
Compress:               false
Package:                true
Clean Build Dir:        false
LDFlags:                "-w -s -H windowsgui"
Tags:                   [desktop,production]
Race Detector:          true

Building target: windows/amd64
------------------------------
  - Generating obfuscated bindings:

ERROR:
# github.com/wailsapp/wails/v2/internal/frontend/runtime
vendor\github.com\wailsapp\wails\v2\internal\frontend\runtime\runtime_prod_desktop.go:9:5: RuntimeDesktopJS redeclared in this block
        vendor\github.com\wailsapp\wails\v2\internal\frontend\runtime\runtime_dev_desktop.go:9:5: other declaration of RuntimeDesktopJS
# github.com/go-vgo/robotgo
vendor\github.com\go-vgo\robotgo\hook.go:98:17: invalid operation: geve == 0 (mismatched types bool and untyped int)

exit status 2

If Wails is useful to you or your company, please consider sponsoring the project:


exit status 1
```

Aivaki:
The application interface breaks anyway, the controls don't work (when you click on the button, it doesn't do anything)

leaanthony:
Did you follow the instructions?
https://feature-garble-support.wails.pages.dev/docs/next/guides/obfuscated

Aivaki:
> Did you follow the instructions? https://feature-garble-support.wails.pages.dev/docs/next/guides/obfuscated

For some reason, it just doesn't compile together with -obfuscate

Osbornnnnn:
I was tested too. But have the same error.
image

leaanthony:
That error suggests you are using the 44.3 release library in your projects. Update `go.mod` to use your local checkout. The instructions are here but not explicit in the sub-sections. I'll update the docs 👍 https://wails.io/docs/guides/bleeding-edge/

Aivaki:
> That error suggests you are using the 44.3 release library in your projects. Update `go.mod` to use your local checkout. The instructions are here but not explicit in the sub-sections. I'll update the docs 👍https://wails.io/docs/next/guides/bleeding-edge

I have updated `go.mod` and other, but the error has not disappeared. I'll try again

leaanthony:
@Snider - did you have any issues with this?

Snider:
> @Snider - did you have any issues with this?

I trust your judgment; I have not checked out the PR and done a build test; would you like me to do this?

I'm happy with Wails' securing the user's application code; it is a good move for the users of Wails.io; excited to see the many ways we can put it to use.

I'll wait for confirmation, but I'm happy if the team is happy. +1

leaanthony:
Yeah, if you could try the PR that would be awesome! I cannot reproduce the errors above.

Snider:
> Yeah, if you could try the PR that would be awesome! I cannot reproduce the errors above.

Ok, I'll drop what I'm doing rn and do this. I'll let you know.

Snider:
> Yeah, if you could try the PR that would be awesome! I cannot reproduce the errors above.

Works for me :)

image

image

leaanthony:
What's the error?

Snider:
> What's the error?

Overall i don't see this as a release blocker, as the documented defaults/functionality intended by you is working for both 

`wails build -obfuscate` and `wails build -obfuscate -garbleargs "-literals -tiny -seed=myrandomseed"`



But when i run the user-reported adjustments, it fails on macOS arm64 : 
`wails build -race -trimpath -obfuscate -m -tags="desktop,production" -compiler="garble" -garbleargs="-tiny -literals -seed=random" -ldflags="-w -s -H windowsgui"`

it fails
image

Aivaki:
I found that I have no problems compiling unless the following parameter is specified `-tags="desktop,production"`

leaanthony:
For manual builds, also specify the `obfuscated` build tag

Aivaki:
> For manual builds, also specify the `obfuscated` build tag. I think there's a build tag missing in the Dev file.

Why is it that `-tags="desktop,production,obfuscated"` an error occurs, but so `-tags="desktop,production"` everything is fine?

Aivaki:
image
image
image

leaanthony:
If you use `wails build -v 2` you'll see what you need to run if you really want to run it manually

AlbinoDrought:
Hello, (unrelated to the above tag issues)

In calls.js, a payload like `{ name: 0, args: [123], callbackID: "1-2345" }` is sent: https://github.com/wailsapp/wails/blob/b57c66a0bddd2c76421d1bbfbe0ddb860bca071d/v2/internal/frontend/runtime/desktop/calls.js#L137

The `secureCallMessage` struct has "id" instead of "name", causing secureCallMessage.ID to always be 0: https://github.com/wailsapp/wails/blob/b57c66a0bddd2c76421d1bbfbe0ddb860bca071d/v2/internal/frontend/dispatcher/securecalls.go#L10

This patch fixes it for me:

```diff
From 9069fd96a4e9172231abb491d06f6730771bb168 Mon Sep 17 00:00:00 2001
From: AlbinoDrought <snip>
Date: Tue, 6 Sep 2022 10:46:15 -0700
Subject: [PATCH] fix: secureCallMessage expects JSON .name instead of .id

---
 v2/internal/frontend/dispatcher/securecalls.go | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/v2/internal/frontend/dispatcher/securecalls.go b/v2/internal/frontend/dispatcher/securecalls.go
index 40accaa2..5583cdab 100644
--- a/v2/internal/frontend/dispatcher/securecalls.go
+++ b/v2/internal/frontend/dispatcher/securecalls.go
@@ -7,7 +7,7 @@ import (
 )
 
 type secureCallMessage struct {
-	ID         int               `json:"id"`
+	ID         int               `json:"name"`
 	Args       []json.RawMessage `json:"args"`
 	CallbackID string            `json:"callbackID"`
 }
-- 
2.35.1

```

Cheers!

leaanthony:
Ok a few things are going on here:

- There was an issue with the message being passed back to the backend (Thanks for that @AlbinoDrought!!). FIXED.
- The generation of bindings had an issue regarding the build tags. FIXED.
- The tags flag accept *SPACE* delimited tags, not comma. However, considering that was a terrible decision to begin with, I've added comma as a valid delimiter for tags - you just can't use both! SORTED.

Thanks for all the testing and getting me to a place of finding these bugs 🙏 Let me know how you get on 👍

v2/cmd/wails/internal/commands/common/usertags.go
```diff
@@ -0,0 +1,33 @@
...
+	var userTags []string
+	separator := ""
+	if strings.Contains(tags, ",") {
+		separator = ","
+	}
+	if strings.Contains(tags, " ") {
+		if separator != "" {
+			return nil, errors.New("cannot use both space and comma separated values with `-tags` flag")
+		}
+		separator = ","
```
stffabi, line 24: Shouldn't this be a space?
```suggestion
		separator = " "
```

v2/cmd/wails/internal/commands/dev/dev.go
```diff
@@ -159,28 +155,39 @@ func AddSubcommand(app *clir.Cli, w io.Writer) error {
...
+				env = append(env, "WAILS_OBFUSCATE=true")
+			}
+
+			userTags, err := common.ParseUserTags(flags.tags)
+			if err != nil {
+				return err
+			}
+			if len(userTags) > 0 {
+				buildOptions.UserTags = userTags
+				genModuleTags := lo.Without(userTags, "desktop", "dev", "debug")
```
stffabi, line 180: This seems to be mostly redundant with the code of the module command itself in `module.go` line 44. Should we remove this here and also exclude `dev` in `module.go`, so we are doing this just in one place?

v2/cmd/wails/internal/commands/dev/dev.go
```diff
@@ -333,6 +340,22 @@ func runCommand(dir string, exitOnError bool, command string, args ...string) er
...
 
+func runCommandWithEnv(dir string, exitOnError bool, env []string, command string, args ...string) error {
+	cmd := exec.Command(command, args...)
+	cmd.Dir = dir
+	cmd.Env = append(os.Environ(), env...)
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		println(string(output))
+		println(err.Error())
+		if exitOnError {
```
stffabi, line 351: Should we remove the `exitOnError` and just let the caller determine what we wants to do?

v2/internal/app/app_dev.go
```diff
@@ -73,12 +73,18 @@ func CreateApp(appoptions *options.App) (*App, error) {
...
+		obfuscateFlag = devFlags.Bool("obfuscate", false, "Directory to serve assets")
```
stffabi, line 85: The description seems to be wrong:
```suggestion
		obfuscateFlag = devFlags.Bool("obfuscate", false, "Code obfuscation of bound Wails methods")
```

v2/internal/app/app_dev.go
```diff
@@ -110,6 +116,9 @@ func CreateApp(appoptions *options.App) (*App, error) {
 		if loglevelFlag != nil {
 			loglevel = *loglevelFlag
 		}
+		if obfuscateFlag != nil && *obfuscateFlag {
+			obfuscate = fmt.Sprintf("%s", *obfuscateFlag)
```
stffabi, line 120: Shouldn't we format here with `%t`, since it's a bool?
```suggestion
			obfuscate = fmt.Sprintf("%t", *obfuscateFlag)
```

v2/internal/app/app_production_obfuscated.go
```diff
@@ -0,0 +1,120 @@
...
+
+	// Set up logger
+	myLogger := logger.New(appoptions.Logger)
+	if IsDebug() {
+		myLogger.SetLogLevel(appoptions.LogLevel)
+	} else {
+		myLogger.SetLogLevel(appoptions.LogLevelProduction)
+	}
+	ctx = context.WithValue(ctx, "logger", myLogger)
+	ctx = context.WithValue(ctx, "obfuscated", true)
```
stffabi, line 67: This file seems to be mostly redundant with `app_production.go`. Should we maybe just add a `IsObfuscated()` method whose value depends on the `obfuscated` build tag? We do the same with `IsDebug()`.

v2/pkg/commands/build/base.go
stffabi, line 173: Should be promote this to an error?
Let's assume someone has ci/cd integration for building their project and they use the obfuscate option in `wails.json`, but forgot to install garble in their build agent environment. In such a case I would rather expect the build to fail instead of having a successful build which has no obfuscation active, and might be released undetected.

v2/pkg/commands/build/build.go
```diff
@@ -151,6 +165,51 @@ func Build(options *Options) (string, error) {
 	return compileBinary, nil
 }
 
+func generateBindings(options *Options) error {
+	options.Logger.Print("  - Generating obfuscated bindings: ")
+	defer func() {
+		options.Logger.Println("Done.")
+	}()
+	// Run go mod tidy to ensure we're up-to-date
+	stdout, stderr, err := shell.RunCommand(".", "go", "mod", "tidy")
```
stffabi, line 174: We normally use `compat=1.17` for mod tidy, should we do this here too?
```suggestion
	stdout, stderr, err := shell.RunCommand(".", "go", "mod", "tidy", "-compat=1.17")
```

v2/pkg/commands/build/build.go
```diff
@@ -151,6 +165,51 @@ func Build(options *Options) (string, error) {
...
+	}
+
+	self := os.Args[0]
+	var env []string
+	if options.Obfuscate {
+		env = append(env, "WAILS_OBFUSCATE=true")
+	}
+
+	if len(options.UserTags) > 0 {
+		genModuleTags := lo.Without(options.UserTags, "desktop", "production", "debug")
```
stffabi, line 188: See also comment in the dev use case, about redundancy with the module command itself.

v2/pkg/commands/build/build.go
```diff
@@ -151,6 +165,51 @@ func Build(options *Options) (string, error) {
...
+
+func runCommandWithEnv(dir string, exitOnError bool, env []string, command string, args ...string) error {
+	cmd := exec.Command(command, args...)
+	cmd.Dir = dir
+	cmd.Env = append(os.Environ(), env...)
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		println(string(output))
+		println(err.Error())
+		if exitOnError {
```
stffabi, line 205: Should we let the caller decide if he wants to exit or not depending on the error return value?

stffabi (commented):
Awesome enhancement 🚀 

I just added a few minor comments.

v2/cmd/wails/internal/commands/dev/dev.go
leaanthony: I was using this as a kind of `MustRunCommandWithEnv`. To be honest, I think there are too many similar "run command" functions. Let's do a proper refactor at some point.

v2/pkg/commands/build/base.go
leaanthony: 💯 Really great point 👍

leaanthony:
Pushed fixes as suggested. Works locally. Thanks for taking the time to review this @stffabi 🙏

Snider:
> Pushed fixes as suggested. Works locally. Thanks for taking the time to review this @stffabi 🙏

Tested, and I get a working build.

v2/cmd/wails/internal/commands/dev/dev.go
stffabi: Yeah, that makes sense. 👍

stffabi (commented):
LGTM, just added another comment which came to my mind.

stffabi:
> Pushed fixes as suggested. Works locally. Thanks for taking the time to review this @stffabi 🙏

You're welcome.

v2/cmd/wails/internal/commands/dev/dev.go
```diff
@@ -159,28 +154,37 @@ func AddSubcommand(app *clir.Cli, w io.Writer) error {
...
+			if flags.obfuscate {
+				env = append(env, "WAILS_OBFUSCATE=true")
+			}
+
+			userTags, err := common.ParseUserTags(flags.tags)
+			if err != nil {
+				return err
+			}
+			if len(userTags) > 0 {
+				buildOptions.UserTags = userTags
```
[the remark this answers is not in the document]
leaanthony: Oh nice spot!

stffabi (approved):

stffabi:
I think we are good to go. 👍

leaanthony:
I was doing some testing and unless my local install is borked, it looks like the generation of the bindings isn't working properly (sometimes leaves the unobfuscated versions and doesn't generate the new ones). I'm not going to have time to look at it for a short while so if anyone would like to test/debug, that would be awesome. Thank you 🙏

AlbinoDrought:
I can't reproduce regular bindings not being replaced with obfuscated bindings, but I can reproduce the opposite:

- `wails build -obfuscate`: `frontend/wailsjs/go/main/App.js` replaced with version using `ObfuscatedCall`
- `wails build`: `frontend/wailsjs/go/main/App.js` not changed, still using `ObfuscatedCall`

<details>
  <summary>Reproducible ObfuscatedCall bindings remaining</summary>
  
```sh
mkdir wails-obfs-stay
cd wails-obfs-stay
# install this PR
git clone https://github.com/wailsapp/wails.git
pushd wails/v2/cmd/wails
git checkout feature/garble-support
go install
popd
# create fresh project using this PR
wails init -n obfs-remain-test
cd obfs-remain-test
echo "replace github.com/wailsapp/wails/v2 => ../wails/v2" >> go.mod
# build obfuscated, works
wails build -obfuscate
cat frontend/wailsjs/go/main/App.js | grep ObfuscatedCall || (echo "Not obfuscated" && exit 1)
# build regular, it's still obfuscated
wails build
cat frontend/wailsjs/go/main/App.js | (! grep ObfuscatedCall) || (echo "Still obfuscated" && exit 1)
```

> src/components/HelloWorld.vue(3,21): error TS2307: Cannot find module '../../wailsjs/go/main/App' or its corresponding type declarations.
> src/components/HelloWorld.vue(11,25): error TS7006: Parameter 'result' implicitly has an 'any' type.

</details>

I was able to work around this using a patch to also regenerate bindings if we're building the frontend:

```path
From deeb5e0ecd8c536f0328b0f942bcb7d078ca9623 Mon Sep 17 00:00:00 2001
From: AlbinoDrought <snip>
Date: Sat, 10 Sep 2022 12:34:05 -0700
Subject: [PATCH] fix: also perform early generation of plain bindings if we'll
 be building the frontend

---
 v2/pkg/commands/build/build.go | 11 +++++++++--
 1 file changed, 9 insertions(+), 2 deletions(-)

diff --git a/v2/pkg/commands/build/build.go b/v2/pkg/commands/build/build.go
index 2c8a44fa..a74b0cf3 100644
--- a/v2/pkg/commands/build/build.go
+++ b/v2/pkg/commands/build/build.go
@@ -130,11 +130,14 @@ func Build(options *Options) (string, error) {
 		}
 	}
 
-	if options.Obfuscate {
+	if options.Obfuscate || !options.IgnoreFrontend {
 		err := generateBindings(options)
 		if err != nil {
 			return "", err
 		}
+	}
+
+	if options.Obfuscate {
 		if !lo.Contains(options.UserTags, "obfuscated") {
 			options.UserTags = append(options.UserTags, "obfuscated")
 		}
@@ -166,7 +169,11 @@ func Build(options *Options) (string, error) {
 }
 
 func generateBindings(options *Options) error {
-	options.Logger.Print("  - Generating obfuscated bindings: ")
+	if options.Obfuscate {
+		options.Logger.Print("  - Generating obfuscated bindings: ")
+	} else {
+		options.Logger.Print("  - Generating bindings: ")
+	}
 	defer func() {
 		options.Logger.Println("Done.")
 	}()
-- 
2.37.1

```

but I don't fully understand the binding generation process (when are normal bindings usually generated?)

As a bonus, my Typescript project doesn't build on latest if I `rm frontend/wailsjs -rf` because of the missing definition files. It succeeds with the above patch.

<details>
  <summary>Reproducible failing TS build on latest</summary>
  
```sh
go install github.com/wailsapp/wails/v2/cmd/wails@latest
wails init -n ts-rm-test -t vue-ts
cd ts-rm-test
wails build # success
rm frontend/wailsjs -rf
wails build # failure
```

> src/components/HelloWorld.vue(3,21): error TS2307: Cannot find module '../../wailsjs/go/main/App' or its corresponding type declarations.
> src/components/HelloWorld.vue(11,25): error TS7006: Parameter 'result' implicitly has an 'any' type.

</details>

leaanthony:
Wow, thanks so much @AlbinoDrought for taking the time to test and write up. What you're seeing is similar to what I noticed and it looks like some funkiness with build tags. What you've noted will be a huge help. I hope to get it fixed soon.

Snider:
> Wow, thanks so much @AlbinoDrought for taking the time to test and write up. What you're seeing is similar to what I noticed and it looks like some funkiness with build tags. What you've noted will be a huge help. I hope to get it fixed soon.

Have you looked into testing frameworks for Go, this is an ideal issue to create a test case; if we want Wails to get used more; probably a good idea.

leaanthony:
It's a fair point. The library part of Wails is really hard to test as there's not much in the way of desktop UI testing outside of JVMs. The CLI part is, however, a fine candidate for it. There's already a number of tests in areas that get quite hairy and I think this has been promoted to "quite hairy".

Snider:
> It's a fair point. The library part of Wails is really hard to test as there's not much in the way of desktop UI testing outside of JVMs. The CLI part is, however, a fine candidate for it. There's already a number of tests in areas that get quite hairy and I think this has been promoted to "quite hairy".

Yes, this is a feature that can not fail, ever, for any reason; any release that has a break with any security feature would be a massive reputation hit for wails.io; I'm sure it won't take too much time to confirm the base functionality either.

If you build out a demo app for the UI, I'm sure it is possible to get https://www.selenium.dev via the JS driver to handle GUI testing in GitHub Actions.

I'm busy for a while, but one day, this is a type of thing I could help with, personal finances allowing.

leaanthony:
As far as I'm aware, selenium and the other are built for browser automation and not webviews. Would be interesting to understand how similar projects tackle it.

Snider:
> As far as I'm aware, selenium and the other are built for browser automation and not webviews. Would be interesting to understand how similar projects tackle it.

It might need some double-sided sticky tape and blue tac, but I don't see why it wouldn't work; since browsers, these days are just webview wrappers... some original definition hacking would be required probably, but should work fine afaik.

https://www.npmjs.com/package/selenium-webdriver

AlbinoDrought:
Selenium would at least work great out of the box for the server started by `wails dev`. [Cypress](https://www.cypress.io/) is a similar project, but has a much better developer experience in my opinion. The downside is that Cypress has some proprietary elements (the libraries you use are MIT licensed, but some features are locked behind their optional SaaS dashboard)

Here's a sample test with Cypress:

Here's what it looks like while running in dev mode:




I'm not sure of any great solutions for testing the compiled versions. Depending on how much interactivity is needed, a robotgo + OpenCV wrapper like this could be used: https://github.com/AlbinoDrought/sample-wails-robotgo-test/blob/master/main_test.go

```go
func TestGreeting(t *testing.T) {
	root := "test/TestGreeting"

	// Click the input
	input := mustFindInDesktop(root, "001_input.png")
	clickGCV(input)

	// Type our name
	robotgo.TypeStr("AlbinoDrought")

	// Click the greeting button
	button := mustFindInDesktop(root, "002_button.png")
	clickGCV(button)

	// Must see our greetings text
	time.Sleep(1 * time.Second)
	mustFindInDesktop(root, "003_greetings.png")
}
```

[RunningTest.c.webm](https://user-images.githubusercontent.com/852873/189543975-49b04a8d-d6bf-4d27-9f56-29dca23941cb.webm)

But running compiled-UI tests like this in headless CI is likely another can of worms.

Snider:
You can run chromium headless for CI testing; that's what I used to do in 2009~ to make PDFs server-side with HTML_Unit; it won't be a standard procedure so it will need dev time probably. 

But I don't see why it couldn't work, and selenium doesn't hide the essential features behind a paywall service, if not selenium, then any system that is 100% open source should be OK.

Selenium IDE might not look nice, but it works and is free... someone could donate some time to their project and make it more excellent... 

https://www.selenium.dev/selenium-ide/ 
https://github.com/SeleniumHQ/selenium-ide/releases

leaanthony:
Good discussion. I feel there's a distinction to be made here about what gets tested. For me, there's 3 things:

- The Wails CLI
- The Wails library
- The final application

The first 2 should be covered by Go's standard tooling plus maybe a testing library. The last one could be covered by Selenium if your concern is what the application does and there might be a use case for using it to check the default templates work correctly, but it's potentially limited benefit as it only checks that it's working in dev mode. I looked fairly heavily into desktop GUI automation and like I said previously, there's a lot around the JVM but not much else. My conclusion was pretty much what @AlbinoDrought came to: robotgo.

leaanthony:
Ok, just pushed a few changes:

- `-obfuscate` flag is now `-obfuscated` (It was getting confusing!)
- `wails build` now generates bindings by default unless `-skipbindings` is given. This was the reason it wasn't generating non-obfuscated bindings previously.
- `wails dev` now does *not* have the option for `-obfuscated`. I don't believe it's necessary.
- Refactored bindings generation (I really want to rename the command `wails generate bindings`!)

So this seems to work pretty well for me. Let me know how you get on 👍

AlbinoDrought:
(Switched to 822baa5dddf2d2f4addb4558d1f3a9a9df216ab1 on my end, running great :+1: )

stffabi (approved):
LGTM 🚀

leaanthony:
Thanks everyone for all the effort. We got there! 💥 🚀 💥

Snider:
Well done, you lot, some slick improvements and teamwork :)

PR #1794, opened 2022-08-26
Update templates.mdx

KiddoV:
Adding new community template with pure JavaScript.


leaanthony:
I'd recommend adding this to `docs/community/templates` or else it will only exist in `v44` and be lost after a couple of versions.

KiddoV:
My bad! I'll fix that.

PR #1811, opened 2022-08-31
Only set GDK_BACKEND to "x11" if GDK_BACKEND is unset and XDG_SESSION_TYPE is not "wayland"

### LINKED ISSUE #1420: [v2] [Linux] Feat - Wayland Support

Fedora, Ubuntu, and other distros are moving to Wayland and moving away from X11. Currently wails will use the XWayland compatibility layer. It would be good to eventually support Wayland. 

I first found out this was an issue because my app was listed on Flathub and I had used one of the template flatpak xml files that listed Wayland as the default socket. Users who were actually running Wayland would not be able to use the app due to GTK errors. 

I figured out that this is due to the `--socket=wayland` finish arg. Here's the older [flatpak xml](https://github.com/flathub/app.riftshare.RiftShare/blob/f1d20f0e5faafd7f9029947b14a00a9ee4bf430f/app.riftshare.RiftShare.yaml) I was using for reference. 

This is the error that was coming up: 
` Gtk-WARNING **: 15:42:30.500: cannot open display: `

The fix was to force the X11 session -- see the updated [flatpak xml](https://github.com/flathub/app.riftshare.RiftShare/blob/master/app.riftshare.RiftShare.yaml)

I am not a Linux expert by any means so I'm not sure what the longer term implications are, but thought it would be prudent to raise the request. 

Support for Wayland Sessions

Fallback to X11 / XWayland

### Additional context

See the detailed description above

prurigro:
The existing logic always sets the `GDK_BACKEND` variable to `x11`, which prevents the application from using the wayland backend (or running at all in a wayland compositor if xwayland is unavailable). The reason this logic is present is because windows would fail to centre without it, so we need to keep the behaviour in x11 sessions.

The `GDK_BACKEND` variable can and typically should remain unset in a wayland session, and the `XDG_SESSION_TYPE` variable is usually set and can have a value of `unspecified`, `tty`, `x11`, `wayland` or `mir`.

The logic in the patch is as follows; if the system has already set the `GDK_BACKEND` variable, we should assume that it knows what it's doing and keep the existing value. If the `GDK_BACKEND` variable is empty however, we check to see if the `XDG_SESSION_TYPE` variable is set to `wayland`, and if it is we shouldn't touch the `GDK_BACKEND` variable. If the `XDG_SESSION_TYPE` variable is not set to `wayland` though, we can assume we're most likely in an x11 environment in those cases. If we're in an x11 environment, we set `GDK_BACKEND` to `x11` and windows can be centred :)

I tested this in x11 and wayland with xwayland disabled using all combinations of `XDG_SESSION_TYPE` and `GDK_BACKEND` being set appropriately, inappropriately and being unset, and got the expected behaviour in all cases.

This resolves #1420

### DIFF
--- a/v2/internal/frontend/desktop/linux/frontend.go
+++ b/v2/internal/frontend/desktop/linux/frontend.go
@@ -58,8 +58,8 @@ func init() {
 
 func NewFrontend(ctx context.Context, appoptions *options.App, myLogger *logger.Logger, appBindings *binding.Bindings, dispatcher frontend.Dispatcher) *Frontend {
 
-	// Set GDK_BACKEND=x11 if currently unset and XDG_SESSION_TYPE isn't wayland to prevent warnings
-	if os.Getenv("GDK_BACKEND") == "" && os.Getenv("XDG_SESSION_TYPE") != "wayland" {
+	// Set GDK_BACKEND=x11 if currently unset and XDG_SESSION_TYPE is unset, unspecified or x11 to prevent warnings
+	if os.Getenv("GDK_BACKEND") == "" && (os.Getenv("XDG_SESSION_TYPE") == "" || os.Getenv("XDG_SESSION_TYPE") == "unspecified" || os.Getenv("XDG_SESSION_TYPE") == "x11") {
 		_ = os.Setenv("GDK_BACKEND", "x11")
 	}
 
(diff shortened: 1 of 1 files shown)


prurigro:
I just pushed a second commit to this PR after deciding that `GDK_BACKEND` probably also shouldn't be overridden if `XDG_SESSION_TYPE` is `mir`, `tty` or some future session type. Rather than ensure it isn't `wayland`, it now ensures it is either unset, `unspecified` or `x11`.

I also ran another set of tests to make sure everything is still working as expected.

leaanthony (approved):
Thanks so much for this! I think this is a more than acceptable way of dealing with the different backends 👍

PR #1821, opened 2022-08-31
Use Promise<void> when Go routine does not output

### LINKED ISSUE #1818: Go routines that do not return a value have incorrect JS generated types
### Description

A Go routine that does not return a value should have a type that returns `Promise<void>` instead of `void`.

### To Reproduce

1) add to app.go:

```
func (a *App) Noop() {}
```

### Expected behaviour

The generated type should be

```js
export function Noop():Promise<void>;
```

### Screenshots

_No response_

### Attempted Fixes

_No response_

### System Details

```shell
n/a
```

### Additional context

_No response_

SheetJSDev:
Fixes #1818

### DIFF
--- a/v2/internal/binding/generate.go
+++ b/v2/internal/binding/generate.go
@@ -80,7 +80,7 @@ func (b *Bindings) GenerateGoBindings(baseDir string) error {
 					}
 					returnType += ">"
 				} else {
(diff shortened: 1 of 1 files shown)


leaanthony:
Thanks for this! 🙏

PR #1827, opened 2022-09-03
Better watcher

leaanthony:
MacOS has really low limits on number of open files and fsnotifier smashes it, leading to `too many open files` errors. This PR greatly reduces the number of open files be using `.gitignore` as a basis for things that should not be watched.


v2/cmd/wails/internal/commands/dev/watcher.go
```diff
@@ -0,0 +1,67 @@
...
+	"path/filepath"
+
+	"github.com/wailsapp/wails/v2/internal/fs"
+
+	"github.com/fsnotify/fsnotify"
+	gitignore "github.com/sabhiram/go-gitignore"
+	"github.com/samber/lo"
+)
+
+type Watcher interface {
```
stffabi, line 15: Seems like we don't need the Watcher interface at the moment. Should we remove it?

stffabi (approved):
LGTM, that's a great enhancement 👍

v2/cmd/wails/internal/commands/dev/watcher.go
leaanthony: I'll keep it in for now as I really do want to use FSEvents for Mac

PR #1828, opened 2022-09-04
Make draggable CSS property customisable

leaanthony:
Adds 2 new applications allowing the developer to choose which CSS property and value should be used to indicate drag regions.


website/docs/reference/options.mdx
```diff
@@ -40,6 +40,8 @@ func main() {
         OnShutdown:         app.shutdown,
         OnBeforeClose:      app.beforeClose,
         WindowStartState:   options.Maximised,
+        CSSDragProperty:   "widows",
```
stffabi, line 43: Should we use the default value here?
```suggestion
        CSSDragProperty:   "--wails-draggable",
```

website/docs/reference/options.mdx
```diff
@@ -40,6 +40,8 @@ func main() {
...
+        CSSDragValue:      "1",
```
stffabi, line 44: Should we use the default value here?

```suggestion
        CSSDragValue:      "drag",
```

stffabi (approved):
LGTM, just two minor comments about the docs.

leaanthony:
Thanks @stffabi 🙏

PR #1853, opened 2022-09-14
[cli] Add check for minimum required go version

stffabi:
- Update out of sync go.mod with minimum go version
- Check for minimum go version during build with build constraint
  “internal/goversion/build_constraint.go:9:2: MinGoVersionRequired (constant "You need Go 1.18 or newer to compile this program" of type string) is not used”

### DIFF
--- a/v2/cmd/wails/internal/commands/build/build.go
+++ b/v2/cmd/wails/internal/commands/build/build.go
@@ -6,7 +6,6 @@ import (
 	"io"
 	"os"
 	"os/exec"
-	"path/filepath"
 	"runtime"
 	"strings"
 	"text/tabwriter"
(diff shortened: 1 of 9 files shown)


stffabi:
According to docs, we currently require 1.17 as minimum version that's why I used 1.17 as minimum version. If we require 1.18 we should update the docs for the v2 release.

>  I think moving forward, we'll always support current-1 as a minimum. Thoughts?

Yeah, I also think that's fine, it corresponds to the Go Release Policy: `Each major Go release is supported until there are two newer major releases.`

leaanthony:
The docs and templates will be updated 👍

stffabi:
> The docs and templates will be updated 👍

I've just pushed PR #1854 for that.

leaanthony:
LGTM!

stffabi:
Thanks for merging, didn't expect this to still land in v2. 🙏

leaanthony:
With the need to use 1.18 I think it's a good addition 👍

PR #1858, opened 2022-09-15
[build] Fix buildtags parsing if only one tag is specified

stffabi:


leaanthony (approved):

leaanthony:
Nice catch! I guess this is not the normal scenario so it wasn't picked up in testing. I might just push a test up to go with it.

stffabi:
Yeah adding a test makes absolutely sense. 👍

leaanthony:
Copilot is weirdly accurate with the intent of tests......almost spooky.... 👻

PR #1862, opened 2022-09-16
[templates] Use embed all to include all files

stffabi:
Otherwise files prefixed with `.` and `_` won't get embedded.


leaanthony:
💯 great addition. Thanks 🙏

PR #1884, opened 2022-09-23
mention Wails is for deskotp apps

tomByrer:
By reading the README, I thought this program was a backend service for web apps, not desktop apps.

TBH, your Intro at the website is more clear; perhaps they should be the same? https://wails.io/docs/introduction

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,9 +61,10 @@
 
 ## Introduction
 
-The traditional method of providing web interfaces to Go programs is via a built-in web server. Wails offers a different
-approach: it provides the ability to wrap both Go code and a web frontend into a single binary. Tools are provided to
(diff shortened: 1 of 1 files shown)


tomByrer:
BTW, project looks interesting; if I ever learn GO, I'll likely use this over Electron.

misitebao:
I think the relationship between go and wails is different from that between node and electron. The biggest difference is that wails does not embed browsers.

leaanthony:
Appreciate you opening this, however the wording is deliberate: it's not just for desktop apps. There is already a PR for server builds and a PoC for mobile available. If anything, we need to update the website to reflect that.

Thanks for your time 🙏

PR #1887, opened 2022-09-23
fix: Synchronise README.ja.md using GitHub actions

### LINKED ISSUE #1876: Sync README.ja.md
Synchronize the contents of the "README.ja.md" file.

shubham-1806:
### What?
It synchronizes contents of the English README into Japanese and creates README.ja.md 

### Solution
- Created a workflow using Translate Readme Action 

### Related Issue
Fixes #1876


misitebao:
@leaanthony 
What do you think about this way?

leaanthony:
@misitebao - I think this is a good way to get *something* there that's better than a copy of the English docs. We could always update it if the translation isn't great? I'm more concerned it will destroy formatting 😂

misitebao:
@leaanthony 
I think it is optional to use this tool without special maintenance, but does the trigger method of action need to be changed to manual click trigger?

PS: For READMEs, I personally recommend manual maintenance.😁

leaanthony:
I think it's a good idea, however I have a number of concerns:

- The linked action appears to call the google API for every single text node in the document which will be a lot. I fear it might fail on us and we wouldn't know
- It appears that the generated readme removes spaces around links which looks bad. This will require manual updates. Which brings me to
- Any manual updates will be lost next time the action is run.

I don't think we should use the action.

misitebao:
@shubham-1806 
For "README" files, using this action has side effects, and we still recommend maintaining it manually. If you are interested, you are welcome to manually sync it and submit a PR.

Issue #1885, opened 2022-09-23
run wails dev bug


NULLB8:
### Description

Description
when run wails dev
![image](https://user-images.githubusercontent.com/21289548/192026156-4e446b9f-c7e9-47d3-81b6-19d8fdd58176.png)
When I run wails run, it shows that build will use this command to compile.
`go build -gcflags all=-N -l -tags dev -o C:\Users\test\GolandProjects\NextExploit\build\bin\NextExploit-dev.exe`
![image](https://user-images.githubusercontent.com/21289548/192026768-eb63c727-17e0-40eb-b1ad-d5184f16e9ba.png)
But when I try to compile the application with this statement, golang shows no -l parameter
I tried removing the -l and it compiled successfully.
I think it is possible that the new version of golang has removed the -l parameter
this is my config:
windows 10 x64
go1.19.1 amd64

### To Reproduce

test

### Expected behaviour

test

### System Details

```shell
Wails CLI v2.0.0

Scanning system - Please wait (this may take a long time)...Done.

System
------
OS:             Windows Server 2019 Datacenter
Version:        1809 (Build: 17763)
ID:
Go Version:     go1.19.1
Platform:       windows
Architecture:   amd64

Wails
------
Version:        v2.0.0

Dependency      Package Name    Status          Version
----------      ------------    ------          -------
WebView2        N/A             Installed       105.0.1343.42
npm             N/A             Installed       8.15.0
*upx            N/A             Available
*nsis           N/A             Available

* - Optional Dependency

Diagnosis
---------
Your system is ready for Wails development!
Optional package(s) installation details:
  - upx : Available at https://upx.github.io/
  - nsis : Available at https://nsis.sourceforge.io/Download



If Wails is useful to you or your company, please consider sponsoring the project:
https://github.com/sponsors/leaanthony
```

leaanthony:
The output appears to be stripping the quotes. It should be `-gcflags "all=-N -l"`

leaanthony:
Please try the above PR to see if this fixes your problem. If you are unsure how to do this, please follow [this guide](https://wails.io/docs/guides/bleeding-edge#testing-a-pr). Let us know how you get on 👍

NULLB8:
I tested this command manually and it worked, but I still got the same `Access is denied` error as before
![image](https://user-images.githubusercontent.com/21289548/192134454-606297d3-1aac-44f3-9141-817536688a09.png)

stffabi:
[...]

PR #1892, opened 2022-09-23
Quote command arguments if they have a space

leaanthony:
Generates better output when using `-v 2`.

Before:
` Build command: go build -tags desktop,production -ldflags -w -s -H windowsgui -o default_no.exe`

After:
` Build command: go build -tags desktop,production -ldflags "-w -s -H windowsgui" -o default_no.exe`

Fixes: #1885


stffabi (approved):
LGTM, 👍

PR #1895, opened 2022-09-24
obfuscated instead of obfuscate in the docs

arifali123:
the option in the docs is the wrong flag you have to add a d at the end


leaanthony:
Nice catch! Thank you 🙏

Issue #1901, opened 2022-09-25
Generated models.ts contains invalid TypeScript in a few edge cases


JulioDRF:
### Description

There's a few cases where `models.ts` contains invalid Typescript, which breaks TS Intellisense in most IDEs:

- Go structs that contain imported structs will sometimes not namespace the imported types correctly
- Go structs that contain Maps with non-string keys fail to convert the key type to JS  (like `uint32` -> `number`)
- Go structs with non-standard JS variable names are used as-is instead of being escaped with double quotes

### To Reproduce

These are the cases I was able to identify.

1. Slices containing imported structs:
```
type Bar struct {
    Foo []imported.Foo `json:"foo"`
}
```

2. Maps containing imported structs:
```
type Bar struct {
    Foo map[string]imported.Foo `json:"foo"` 
}
```

3. Maps containing non-string key types:

```
type Bar struct {
    Foo map[uint64]any `json:"foo"`;
}
```

4. Fields with JSON bindings that don't make valid JS variable names

```
type Bar struct {
    Foobar any `json:"foo.bar"`
}
```

### Expected behaviour

1. Slices containing imported structs:
  For this Go struct:
```
type Bar struct {
    Foo []imported.Foo `json:"foo"`
}
```
 I expected the following TS binding:
```
export class Bar {
    foo: imported.Foo[];
    ...
}
```
but got:
```
export class Bar {
   foo: Foo[];
    ...
}
```

2. Maps containing imported structs:
  For this Go struct:
```
type Bar struct {
    Foo map[string]imported.Foo `json:"foo"` 
}
```
 I expected the following TS binding:
```
export class Bar {
   foo: {[key: string]: imported.Foo};
    ...
}
```
but got:
```
export class Bar {
   foo: {[key: string]: Foo};
    ...
}
```

3. Maps containing non-string key types:

  For this Go struct:
```
type Bar struct {
    Foo map[uint64]any `json:"foo"`;
}
```
 I expected the following TS binding:
```
export class Bar {
  foo: {[key: number]: any};
    ...
}
```
but got:
```
export class Bar {
   foo: {[key: uint64]: any};
    ...
}
```

4. Fields with JSON bindings that don't make valid JS variable names
  For this Go struct:
```
type Bar struct {
    Foobar any `json:"foo.bar"`
}
```
 I expected the following TS binding:
```
export class Bar {
  "foo.bar": any;
    ...
}
```
but got:
```
export class Bar {
   foo.bar: any
    ...
}
[...]

Issue #1772, opened 2022-08-20
Unable to infer types of sub backend packages and some golang data types in wailsjs models.d.ts generation


Purple-CSGO:
## Problem 1
My Directory structure is like this:
```
backend/
  config/
  core/
  app.go
frontend/
main.go
```
So there are sub/subusb/... packages. In `models.d.ts`, types from different sub/... packages are divided to different namespaces, but wails does not refer to the namespaces, resulting in typescripts errors. Part of my `models.d.ts` file is as follows, where `Item[]` should be `tool.Item[]`

```typescript
export namespace tool {
  export class Item {
    name: string;
    value: string;
    ...
  }
}

export namespace config {
  export class CFG {
    ...
    videoSetting: Item[];
    hwinfo: Item[];
  }
}
```

## Problem 2

some go types are incorrectly dealed with.

``` typescript
members: {[key: uint64]: CommonDemoSinglePlayerInfo};
```

Console output after wails dev/build
```
wailsjs/go/models.ts(242,22): error TS2304: Cannot find name 'uint64'.
```

### To Reproduce

`wails dev` or `wails build`

### Expected behaviour

```bash
    wailsjs/go/models.ts(58,17): error TS1268: An index signature parameter type must be 'string', 'number', 'symbol', or a template literal type.
    wailsjs/go/models.ts(58,22): error TS2552: Cannot find name 'uint64'. Did you mean 'GLuint64'?
    wailsjs/go/models.ts(186,20): error TS2304: Cannot find name 'Item'.
    wailsjs/go/models.ts(187,14): error TS2304: Cannot find name 'Item'.
    wailsjs/go/models.ts(199,16): error TS2304: Cannot find name 'User'.
    wailsjs/go/models.ts(218,73): error TS2304: Cannot find name 'Item'.
    wailsjs/go/models.ts(219,61): error TS2304: Cannot find name 'Item'.
    wailsjs/go/models.ts(231,65): error TS2304: Cannot find name 'User'.
     ELIFECYCLE  Command failed with exit code 2.
```

### System Details

```shell
Wails CLI v2.0.0-beta.44.2

Scanning system - Please wait (this may take a long time)...Done.

System
------
OS:             Windows 10 Pro
Version:        2009 (Build: 22000)
ID:             21H2
Go Version:     go1.18.3
Platform:       windows
Architecture:   amd64

Wails
------
Version:        v2.0.0-beta.44.2

Dependency      Package Name    Status          Version
[...]

Issue #1476, opened 2022-06-21
binding / typescriptify: nested structs are not "known"


devnode:
### Description

https://github.com/wailsapp/wails/blob/43f8a1f773f3a3ce03898da65335abca62f898a5/v2/internal/binding/reflect.go#L101-L108

https://github.com/wailsapp/wails/blob/43f8a1f773f3a3ce03898da65335abca62f898a5/v2/internal/typescriptify/typescriptify.go#L631-L634

since there's no recursion when a binding a struct, a nested struct will not be added to `typescriptify.KnownStructs`, which leads to an invalid `modules.ts` file being generated and eventually a console error of: `Uncaught TypeError: n is not a constructor`.

```go
// one line workaround
func (a *App) StructSettings() Settings { return Settings{} }
```

### To Reproduce

`app.go`

```go
type Config struct {
    Target string `json:"target"
    Settings Settings `json:"settings"`
}

type Settings struct {
    Enabled bool `json:"enabled"`
}

...

func (a *App) ConfigState() Config {
    ...
    return a.cfg
}
```

`frontend/wailsjs/go/models.ts`
```ts
export namespace main {
    export class Config {
            // Go type: Settings
            settings: any;  
            ...
	    constructor(source: any = {}) {
	        if ('string' === typeof source) source = JSON.parse(source);
	        this.target = source["target"];
	        this.settings = this.convertValues(source["settings"], null);
	    }
            ...
    }

    // this generated fine
    export class Settings {
    ...
    }
}
```

### Expected behaviour

`frontend/wailsjs/go/models.ts`
```ts
export namespace main {
    export class Config {
            settings: Settings;  
            ...
	    constructor(source: any = {}) {
	        if ('string' === typeof source) source = JSON.parse(source);
	        this.target = source["target"];
	        this.settings = this.convertValues(source["settings"], Settings);
	    }
            ...
    }
    ...
}
```

### Attempted Fixes

`main.go`

```go
	err := wails.Run(&options.App{
                ...
		Bind: []interface{}{
			app,
			&Config{},
			&Settings{},
		},
	})
```

### System Details

```shell
System
------
OS:		Fedora Linux
Version: 	36
ID:		fedora
Go Version:	go1.18.3
Platform:	linux
Architecture:	amd64

Wails
------
Version: 		v2.0.0-beta.37
Package Manager: 	dnf

Dependency	Package Name		Status		Version
----------	------------		------		-------
*docker		Unknown			Not Found	
[...]

Issue #1444, opened 2022-06-08
TS model from mixed package struct don't generated


avengerweb:
### Description

I have struct that defined in "main" package, this struct has fields (struct) defined in "internal" package, in such case model will be generated only for "main" struct and for "internal" it will be missing. If I move all fields to same package it works fine.

### To Reproduce
`github.com/example/example/internal/config.go`
```golang
package internal

type Authenticate struct {
	Token string `json:"token"`
}

type Cache struct {
	Limit     int    `json:"limit"`
	Enabled   bool   `json:"enabled"`
	Directory string `json:"directory"`
}
```

`github.com/example/example/main.go`
```golang
package main

import "github.com/example/example/internal"

type GlobalConfiguration struct {
	Authenticate *internal.Authenticate `json:"authenticate"`
	Cache        *internal.Cache        `json:"cache"`
}
```

### Expected behaviour

cross-package structs should be generated correctly

### System Details

```shell
$ wails doctor
Wails CLI v2.0.0-beta.37


Scanning system - Please wait (this may take a long time)...Done.

System
------
OS:             Windows 10 Enterprise
Version:        2009 (Build: 19042)
ID:             20H2
Go Version:     go1.18.1
Platform:       windows
  - upx : Available at https://upx.github.io/
  - nsis : Available at https://nsis.sourceforge.io/Download
```

leaanthony:
Thanks for raising this.

leaanthony:
Please try the above PR to see if this fixes your problem. If you are unsure how to do this, please follow [this guide](https://wails.io/docs/guides/bleeding-edge#testing-a-pr). Let us know how you get on 👍

leaanthony:
The PR has now been merged. If this issue is still valid, please reopen. Thanks 👍

PR #1902, opened 2022-09-26
Fix binding generation special cases

JulioDRF:
Fixes https://github.com/wailsapp/wails/issues/1901

These changes prevent `binding.GenerateModels()` from yielding broken TypeScript for the following special cases:

- [x] Struct fields of imported types that also use imported types don't create a namespace for the nested type (it's messy to explain, see the `binding_nestedimport_test`)
- [x] Struct fields of type `[]T` or `map[any]T` where `T` is an imported struct don't reference the imported namespace in the generated code (e.g. we get `foo: Foo[]` instead of  `foo: imported.Foo[]`)
- [x] Struct fields with non-standard JS variable names (e.g. `` Foobar string `json:"foo-bar"`)
- [x] Struct fields of type `map[T]any` (where `T` is anything but `string`)

And finally:
- [x] Added tests for all of the above

Extra notes:

I necromanced the binding tests from https://github.com/wailsapp/wails/tree/feature/ts-generation-tests and expanded upon them. I see that the branch has since been deleted, and the directory structure for the tests in this PR feels inelegant, so let me know if there's an obvious way to improve it and I will revise it.

As far as I can tell, the new tests cover these issues and they are resolved
- https://github.com/wailsapp/wails/issues/1444
- https://github.com/wailsapp/wails/issues/1476
- https://github.com/wailsapp/wails/issues/1772


leaanthony:
This is amazing! Thank you so much for taking the time and effort to look into this important area. It would be good to get feedback from those who opened the linked issues and see what the feedback is. Other than that, it looks great!

leaanthony:
I think we can go with this. Thank you so much 🙏

leaanthony (approved):

PR #1907, opened 2022-09-26
docs: sync documents

misitebao:
Sync Chinese and Japanese documents.


leaanthony (approved):
Sorry for the delay!!

PR #1908, opened 2022-09-27
docs: Add reference to built-in templates

lalyos:


leaanthony:
Hey there! Thanks for taking the time to open this. The page referenced is specifically for community templates and the built-in templates are already documented. Will close this for now 👍

PR #1926, opened 2022-10-01
Feature: Add window transparency for linux

### LINKED ISSUE #1923: [v2 Linux] Transparent window on linux

Currently in v2 we can use a transparent window on Windows and Mac, but Linux is missing this feature.

I would like to be able to use transparent windows on Linux as well.

I would like to add the transparent window functionality to `internal/frontend/desktop/linux/window.go` and a new option to `linux.Options` to enable it.

I was able achieve window transparency without any blur/frosting effect. (I think blur could be added in GTK4)

If it is ok to add before v2.1.0 (it is not in the road map) I'd be happy to make a PR for it.

Did not consider anything else.

### Additional context

Screenshot from 2022-09-30 20-10-17

Lyimmi:
This PR resolves #1923 by implementing window transparency on linux.

- Added transparency in `v2/internal/frontend/desktop/linux/window.go`
- Created `WindowIsTranslucent` option in `v2/pkg/options/linux/linux.go`
- Updated docs in `website/docs/reference/options.mdx`

Tested on:

- Linux/amd64, PopOs 22.04 with X11 (In this case its the same as Ubuntu 22.04)
    - Works as intended.
- Linux/arm, Raspbian GNU/Linux 11 (bullseye)
    - It does not work on raspbian, because the window manager doesn't supports compositing, but the case is handled, so the window opens as opaque.


leaanthony:
Nice one!

leaanthony (approved):

PR #1927, opened 2022-10-01
Add WindowExecJS method

leaanthony:
Exposes the existing `ExecJS` method as `WindowExecJS`


v2/pkg/runtime/window.go
```diff
@@ -163,6 +163,12 @@ func WindowIsNormal(ctx context.Context) bool {
 	return appFrontend.WindowIsNormal()
 }
 
+// WindowExecJS executes the given Js in the window
+func WindowExecJS(ctx context.Context, js string) {
```
stffabi, line 167: We are missing the implementation in the DevServer frontend. We could maybe embed the base Frontend into the DevServer so we don't need to always add a wrapper method there.

I have commit for that ready, if you want I can add it to the PR?

website/docs/reference/runtime/window.mdx
```diff
@@ -41,6 +41,12 @@ Centers the window on the monitor the window is currently on.
 Go: `WindowCenter(ctx context.Context)`<br/>
 JS: `WindowCenter()`
 
+### WindowExecJS
+
+Executes arbitrary JS code in the window.
```
stffabi, line 46: Should we explicitly mention this is fire and forget, so incorrect/failing script snippets won't be logged out somewhere and won't panic?

v2/pkg/runtime/window.go
leaanthony: That would be great 👍 Thank you 👍

website/docs/reference/runtime/window.mdx
leaanthony: Yes, great point. I'll update the docs.

v2/pkg/runtime/window.go
stffabi: Just pushed it.

website/docs/reference/runtime/window.mdx
leaanthony: @stffabi - docs updated. Looks like there's a merge conflict caused by #1931 (clashes with the embed commits). Happy to look at it later but if it's obvious to you and a really quick fix then that would be cool 🙏 I think we're almost ready for v2.0.1 🎉

stffabi: Just pushed the fix for the conflicts.

stffabi: Docs looking great 👍

stffabi (approved):
LGTM 👍

Issue #1928, opened 2022-10-02
Init the project will leave the git commit record of the template project


misitebao:
### Description

After initializing the project with the remote template, the `.git` directory in the template project is not automatically deleted, which will cause the newly created project to still retain the commit record of the template project. The git commit record will not be reinitialized even if the `-g` flag is used.

### To Reproduce

1. Run `wails init -n Wails-Vue -t https://github.com/misitebao/wails-template-vue`
2. Run `cd Wails-Vue`
3. Run `git log`
4. See errors

### Expected behaviour

When initializing the project, with or without the `-g` flag, the `.git` directory in the template project should be removed.

### System Details

```shell
$ wails doctor
Wails CLI v2.0.0

Scanning system - Please wait (this may take a long time)...Done.

System
------
OS:             Windows 10 Enterprise LTSC 2019
Version:        1809 (Build: 17763)
ID:
Go Version:     go1.19
Platform:       windows
Architecture:   amd64

Wails
------
Version:        v2.0.0
Revision:       6a3b89d212fb23e47b24e10ea21aa95988d04b90
Modified:       true

Dependency      Package Name    Status          Version
----------      ------------    ------          -------
WebView2        N/A             Installed       105.0.1343.53
npm             N/A             Installed       8.19.2
*upx            N/A             Available
*nsis           N/A             Available

* - Optional Dependency

Diagnosis
---------
Your system is ready for Wails development!
Optional package(s) installation details:
  - upx : Available at https://upx.github.io/
  - nsis : Available at https://nsis.sourceforge.io/Download
```

### Additional context

I am willing to fix this bug.

PR #1929, opened 2022-10-02
fix: remove the `.git` directory in the template

misitebao:
fixed: https://github.com/wailsapp/wails/issues/1928


leaanthony (approved):
Nice one!

Issue #1919, opened 2022-09-29
[v2 linux] wails dev command fails on latest


lyimmi:
### Description

After the commit b2069c871ddb9dc5a530e927e7559e90731cb152 `wails dev` fails with the following message:

```
Building application for development...
  - Generating bindings: Done.
  - Compiling application: # github.com/wailsapp/wails/v2/internal/frontend/devserver
../mywails/v2/internal/frontend/devserver/devserver.go:86:56: not enough arguments in call to assetserver.NewAssetHandler
	have (context.Context, *options.App)
	want (context.Context, fs.FS, http.Handler)
../mywails/v2/internal/frontend/devserver/devserver.go:338:60: cannot use d (variable of type *DevWebServer) as type frontend.Frontend in argument to d.dispatcher.ProcessMessage:
	*DevWebServer does not implement frontend.Frontend (missing RunMainLoop method)

```

v2/internal/frontend/devserver/devserver.go:86
Looks like the definition of `NewAssetHandler()` changed but the usage didn't.

### To Reproduce

1. create a new project with the latest master
2. run wails dev
3. see the error message

### Expected behaviour

I expect that wails runs in dev mode.

### Screenshots

![Screenshot from 2022-09-29 17-33-58](https://user-images.githubusercontent.com/8627125/193076462-ebc952fe-4665-4320-b40e-bae81b2649e3.png)

### Attempted Fixes

I have not tried to fix this.

### System Details

```shell
System
------
OS:		Ubuntu
Version: 	22.04
ID:		ubuntu
Go Version:	go1.19.1
Platform:	linux
Architecture:	amd64

Wails
------
Version: 		v2.0.0
Package Manager: 	apt

Dependency	Package Name		Status		Version
----------	------------		------		-------
*docker 	docker.io 		Installed 	20.10.18
gcc 		build-essential 	Installed 	12.9ubuntu3
libgtk-3 	libgtk-3-dev 		Installed 	3.24.33-1ubuntu2
libwebkit 	libwebkit2gtk-4.0-dev 	Installed 	2.36.8-0ubuntu0.22.04.1
npm 		npm 			Installed 	6.14.13
*nsis 		nsis 			Available 	3.08-2
pkg-config 	pkg-config 		Installed 	0.29.2-1ubuntu3

* - Optional Dependency

Diagnosis
---------
Your system is ready for Wails development!
Optional package(s) installation details: 
  - nsis: sudo apt install nsis
```

leaanthony:
[...]

PR #1909, opened 2022-09-27
Refactor app

leaanthony:
* Refactor the way the main application works as a starting point for #1480 and #1521 
* Clean up `app` package
* Remove need for application options in AssetServer
* Add experimental programmatic API that uses the refactored app

PR #1931, opened 2022-10-02
Fix wails dev

JulioDRF:
Fixes https://github.com/wailsapp/wails/issues/1919

I made the minimum necessary changes to get the dev server working again.
I understand there was a recent refactor in #1909, so please let me know if these changes get in the way of something else.

### DIFF
--- a/v2/internal/frontend/devserver/devserver.go
+++ b/v2/internal/frontend/devserver/devserver.go
@@ -122,28 +122,28 @@ func (d *DevWebServer) Run(ctx context.Context) error {
 	})
 
 	if devServerAddr := d.devServerAddr; devServerAddr != "" {
-		wg := sync.WaitGroup{}
+		wg := &sync.WaitGroup{}
 		wg.Add(1)
 		// Start server
-		go func(server *echo.Echo, log *logger.Logger) {
+		go func(server *echo.Echo, log *logger.Logger, wg *sync.WaitGroup) {
 			defer wg.Done()
 			err := server.Start(devServerAddr)
 			if err != nil {
 				log.Error(err.Error())
 			}
-		}(d.server, d.logger)
+		}(d.server, d.logger, wg)
 
 		d.LogDebug("Serving DevServer at http://%s", devServerAddr)
 
-		go func() {
+		go func(server *echo.Echo, log *logger.Logger, wg *sync.WaitGroup) {
 			wg.Wait()
 			d.LogDebug("Starting shutdown")
-			err := d.server.Shutdown(context.Background())
+			err := server.Shutdown(context.Background())
 			if err != nil {
-				d.logger.Error(err.Error())
+				log.Error(err.Error())
 			}
 			d.LogDebug("Shutdown completed")
-		}()
+		}(d.server, d.logger, wg)
 	}
 
 	// Launch desktop app
--- a/v2/internal/frontend/devserver/devserver.go
+++ b/v2/internal/frontend/devserver/devserver.go
@@ -122,30 +122,38 @@ func (d *DevWebServer) Run(ctx context.Context) error {
 	})
 
 	if devServerAddr := d.devServerAddr; devServerAddr != "" {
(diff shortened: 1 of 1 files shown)


v2/internal/frontend/devserver/devserver.go
```diff
@@ -122,32 +122,40 @@ func (d *DevWebServer) Run(ctx context.Context) error {
...
+		wg.Add(1)
 		// Start server
 		go func(server *echo.Echo, log *logger.Logger) {
+			defer wg.Done()
 			err := server.Start(devServerAddr)
 			if err != nil {
 				log.Error(err.Error())
 			}
-			d.LogDebug("Shutdown completed")
 		}(d.server, d.logger)
```
leaanthony, line 134: I'd be tempted to pass in the address of the waitgroup here

v2/internal/frontend/devserver/devserver.go
```diff
@@ -122,32 +122,40 @@ func (d *DevWebServer) Run(ctx context.Context) error {
...
-		defer func() {
+		go func() {
+			wg.Wait()
+			d.LogDebug("Starting shutdown")
 			err := d.server.Shutdown(context.Background())
 			if err != nil {
 				d.logger.Error(err.Error())
 			}
+			d.LogDebug("Shutdown completed")
 		}()
```
leaanthony, line 146: For consistency, perhaps we should pass everything in, like we do for startup?

leaanthony (commented):
Just a couple of comments. Thoughts?

v2/internal/frontend/devserver/devserver.go
JulioDRF: Makes sense to me

JulioDRF: I added the other parameters but I left the `d.LogDebug` statements the way they are.

I tried to come up with ways to avoid it, but they all felt contrived and/or painful to maintain. Let me you what you think!

leaanthony: Nice one! I'm now wondering if these have to be 2 goroutines. I'm assuming the `server.Start` call is blocking otherwise the shutdown process would begin straight away. What do you think about removing the second goroutine and just put everything in the first? Then we could remove the waitgroup altogether?

stffabi: Maybe I'm missing something but aren't we now blocking the call to `server.Shutdown`? `sever.Start` blocks until someone calls a `server.Shutdown`, the WaitGroup onto which the `server.Shutdown` waits for is only decremented when `server.Start` finishes. So it seems like both are blocking each other.

leaanthony: I think you're right - I missed the `defer` in the original whilst looking at github mobile 😬 I take it the dev server can be shut down with out of band actions such as `ctrl-c`?

JulioDRF: @stffabi Great point! I believe you're right. The problem I ran into was that the deferred `server.Shutdown()` was running immediately because `frontend.Frontend.Run()` no longer blocks.

It didn't click for me that by the time the waitgroup unblocks, the server has already shut down. And yes, `Ctrl-C` shuts down the server as expected with the current commit, I've tested it.

It may not make sense to explicitly call `server.Shutdown` anymore; though I could phase out the waitgroup and make the goroutine wait for a context cancellation instead. I'm not sure what's better, so please let me know your thoughts.

stffabi: Personally I'm fine with completely removing the shutdown for the time being. For the devserver an ungraceful shutdown is not much of a problem.
We could revisit this for the server/hybrid mode where a graceful shutdown might be more relevant.
@leaanthony what do you think?

leaanthony: Yeah happy with that.

stffabi (approved):
LGTM 👍, thanks for this fix 🙏

PR #1946, opened 2022-10-08
Fix for considering new `go` files in dev filesystem watcher

scottopell:
## What is this
Adds support for watching for creation of new files while running `wails dev`.

## Motivation
I encountered a bug where updates to my `go` files were not resulting in a refresh of the wails app.

I added some logging to see what events were coming out of `fsnotify` and I found that when I wrote updates to my code, fsnotify was picking up a sequence of `REMOVE` `CREATE` `CHMOD` events instead of a `WRITE` as this code was expecting.

Tested and working on macOS 12.4.

```
FSNotify Event Item:  "/Users/scottopell/dev/heaprof/app.go": REMOVE
FSNotify Event Item:  "/Users/scottopell/dev/heaprof/app.go": CREATE
FSNotify Event Item:  "/Users/scottopell/dev/heaprof/app.go": CHMOD
```

### DIFF
--- a/v2/cmd/wails/internal/commands/dev/dev.go
+++ b/v2/cmd/wails/internal/commands/dev/dev.go
@@ -574,7 +574,19 @@ func doWatcherLoop(buildOptions *build.Options, debugBinaryProcess *process.Proc
 		case err := <-watcher.Errors:
 			LogDarkYellow(err.Error())
 		case item := <-watcher.Events:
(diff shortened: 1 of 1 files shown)


leaanthony (approved):

leaanthony:
Nice catch @scottopell! PR looks good too. Thanks heaps! 👍

PR #1947, opened 2022-10-08
feature/turn off color terminal output of dev command

scottopell:
Sometimes you don't want color in your terminal (or maybe you're piping the output out somewhere and you don't want the color control chars)


leaanthony:
Thanks for doing this and I agree it would be worthwhile. Wails has historically used the International English version of "colour". Could we make this PR consistent with that please? "Col" is also fine 👍

leaanthony:
Thanks heaps for doing this 👍

leaanthony (approved):

Issue #1083, opened 2022-01-17
[v2] Windows: random access violations during DispatchMessage


stffabi:
**Description**
We have seen several reports from users with AccessViolations during `DispatchMessage/GetMessage` in `RunMainLoop`

```
Exception 0xc0000005 0x0 0x7ff811db0998 0x7ff911b3d98c
PC=0x7ff911b3d98c
runtime.cgocall(0x86060, 0x9888a0)
    /usr/local/go/src/runtime/cgocall.go:156 +0x4a fp=0xc000115af8 sp=0xc000115ac0 pc=0x2404a
syscall.Syscall(0x7ff967eb0bf0, 0x1, 0xc00037c0f0, 0x0, 0x0)
    /usr/local/go/src/runtime/syscall_windows.go:483 +0xf4 fp=0xc000115b30 sp=0xc000115af8 pc=0x80c74
syscall.Syscall(0x0, 0x100, 0x10, 0x9, 0xc000115be8)
    <autogenerated>:1 +0x2b fp=0xc000115b80 sp=0xc000115b30 pc=0x86f4b
syscall.(*Proc).Call(0x2d507, {0xc00038c020, 0x488c00, 0x82401})
    /usr/local/go/src/syscall/dll_windows.go:183 +0xc6 fp=0xc000115c70 sp=0xc000115b80 pc=0x9e9e6
syscall.(*LazyProc).Call(0xc000188f00, {0xc00038c020, 0x1, 0x1})
    /usr/local/go/src/syscall/dll_windows.go:340 +0x50 fp=0xc000115ca0 sp=0xc000115c70 pc=0x9fd50
github.com/leaanthony/winc/w32.DispatchMessage(...)
    /builds/app/vendor/github.com/leaanthony/winc/w32/user32.go:326
github.com/leaanthony/winc.RunMainLoop()
...
```

**To Reproduce**
Steps to reproduce the behaviour: This is very hard, because it only manifests randomly. Fortunatelay we had one user on whose machine it happened during every startup of the Application. Apart from this user, we only had panics which occurred randomly. 

**Additional context**
After some investigation I've found several Go projects, which also run Windows MessageLoops, to have reports of similar AccessViolations during `DispatchMessage/GetMessage`, e.g. `lxn/walk` or `go-sciter`. Many thanks to @zx2c4 who created a PR for lxn/walk#493 which fixed the issue there. It turned out it also fixed our issue for that user.

[...]

PR #1959, opened 2022-10-12
[RFC] Support tray menus (Windows)

leaanthony:
# PR Goals
This PR has been opened to solicit feedback and discussion on system tray support. 
The goals of this PR are:
* Determine if feature set is complete enough for an initial release
* Determine what likely gaps there are with the other platforms, e.g. Template icons for Mac, HiDPI alternative icons.
* Get feedback on the experimental programmatic API and where it could be improved
* Find bugs 

# What does this PR provide?
This PR provides the ability to add an icon to the Windows system tray area, with the option of adding a menu. The system tray has callbacks for both left and right buttons. By default, both these buttons will show the associated menu. Currently, this capability is provided by an experimental API that builds on what is currently available but provides a more hierarchical model for Wails applications. 

An example application is provided in `v2/examples/systray`. It starts with the main window hidden. This can be made visible by left clicking the tray icon. Right clicking the tray icon shows the menu. The menu is a "kitchen sink" example of what is possible. There are plenty of comments in the example which detail more about what is possible with the system tray.

# How can I provide feedback?

Please comment on this PR with any feedback. Thank you! 🙏


v2/pkg/options/systemtray.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/wailsapp/wails/v2/pkg/menu"
+)
+
+// SystemTray contains options for the system tray
+type SystemTray struct {
+	LightModeIcon *SystemTrayIcon
+	DarkModeIcon  *SystemTrayIcon
+	Title         string
+	Tooltip       string
+	StartHidden   bool
```
teddywing, line 13: What does it mean to `StartHidden` on the `SystemTray`?

v2/internal/platform/systray/windows.go
```diff
@@ -0,0 +1,385 @@
...
+	err := p.Stop()
+	if err != nil {
+		println(err.Error())
+	}
+}
+
+func (p *Systray) Update() error {
+	// Delete old menu
+	if p.menu != nil {
+		p.menu.Destroy()
```
teddywing, line 56: Not sure if I’ve understood this correctly, but is the menu destroyed on each `Update()`?

I’m not familiar with Windows APIs, but it seems a tad wasteful to completely remove the old menu and recreate it on update. Is it possible to update the menu item objects and have them redraw when needed?

teddywing (commented):
This looks cool!

I tried out and read over the example code in `v2/examples/systray/main.go`. The API looks straightforward.

There’s one UI behaviour that I was confused by. If I:

1. Open the menu by right-clicking the tray icon
2. Left-click the tray icon
3. The menu disappears and the window opens

Should the window be opening? I would have thought that the menu should close, and the window would only open if the menu wasn’t already open.

You asked about gaps with other platforms. In addition to the template icons you mentioned, `NSStatusItem`s have a `behavior` property (https://developer.apple.com/documentation/appkit/nsstatusitembehavior?language=objc). It might be nice to add behaviours to provide additional interaction options for removing the `NSStatusItem`.

teddywing:
One more idea: Is there any way to animate the tray icon in order to indicate, for instance, that the application is making a network request?

leaanthony:
Thanks @teddywing for taking the time to test this out and provide feedback 👍 

> There’s one UI behaviour that I was confused by. If I:
> 
> Open the menu by right-clicking the tray icon
> Left-click the tray icon
> The menu disappears and the window opens
> Should the window be opening? I would have thought that the menu should close, and the window would only open if the menu wasn’t already open.

I'm not sure we can be 💯 sure this is the desired behaviour for all system trays. We do have the left and right click callbacks. Do you think this would be sufficient for a developer to produce the outcome you suggested?
 
> You asked about gaps with other platforms. In addition to the template icons you mentioned, NSStatusItems have a behavior property (https://developer.apple.com/documentation/appkit/nsstatusitembehavior?language=objc). It might be nice to add behaviours to provide additional interaction options for removing the NSStatusItem.

👍

v2/pkg/options/systemtray.go
leaanthony: It means it won't show it automatically on start up, but will rely on the developer calling `Show` when ready e.g. there may be the need to do some background tasks at application startup and you want to populate the menu based on the result of that.

v2/internal/platform/systray/windows.go
leaanthony: I've debated this but so far I'm unconvinced that it's an optimisation worth the effort. How often do people do this? How much time or memory do you really save? A lot of APIs make the developer remember where menu items are based on array indexes and use that as a mechanism for updating the underlying item only. I'm unconvinced of that approach too.

v2/pkg/options/systemtray.go
teddywing: Understood, thanks for clearing that up.

v2/internal/platform/systray/windows.go
teddywing: Thanks for the explanation. I agree that the current approach makes it simpler to update the menus.

teddywing:
> I'm not sure we can be 💯 sure this is the desired behaviour for all system trays. We do have the left and right click callbacks. Do you think this would be sufficient for a developer to produce the outcome you suggested?

Yes you’re right, it makes more sense for these behaviours to be defined by the app developer. Are there APIs to know when the menu is visible and to hide & show it?

> You should be able to call `SetIcon()` from a Goroutine with a timer (I haven't tried it but it totally should work).

Cool, that sounds good.

leaanthony:
> Yes you’re right, it makes more sense for these behaviours to be defined by the app developer. Are there APIs to know when the menu is visible and to hide & show it?

There are now 👍

leaanthony:
Considering that this PR has now been open for 2 weeks and that this feature will be a WIP for a couple of releases, I propose that we merge this as is and move onto implementing the same functionality on mac. Rince and repeat for Linux. Otherwise we'll end up with very stale branches and merge problems. The documentation can be updated when finalised and ready for public consumption.

leaanthony:
Considering it's been 2 weeks since this PR was opened, I propose we lock this API down for now and move on to Mac support.

v2/internal/platform/systray/windows.go
```diff
@@ -0,0 +1,430 @@
...
+	default:
+		//msg := int(wparam & 0xffff)
+		//println(win32.WMMessageToString(uintptr(msg)))
+	}
+
+	result, _, _ := DefWindowProc.Call(uintptr(hwnd), uintptr(msg), wparam, lparam)
+	return result
+}
+
+func (p *Systray) Run() error {
```
stffabi, line 356: Should we really have several MessageLoops running in the same application? Windows and e.g. .NET almost always runs only one MessageLoop on the main thread. I thank that would make it a lot easier to reason about on which thread one can access GUIs and not.

v2/internal/platform/systray/windows.go
```diff
@@ -0,0 +1,430 @@
...
+	var msg win32.MSG
```
stffabi, line 357: We might suffer here from the same problem as we had with our main message loop here: #1083

v2/internal/platform/systray/windows.go
```diff
@@ -0,0 +1,430 @@
...
+	for {
```
stffabi, line 358: Don't we need to lock down the OS thread in here? There's no guarantee that if run in a Go routine, all calls will be executed on the same OS thread. And we might then pull messages from the wrong OS thread or miss some events if it's not run on the message loop that created the systray.

v2/internal/platform/win32/structs.go
```diff
@@ -0,0 +1,54 @@
...
+	DwStateMask      uint32
+	SzInfo           [256]uint16
+	UVersion         uint32
+	SzInfoTitle      [64]uint16
+	DwInfoFlags      uint32
+	GuidItem         GUID
+	HBalloonIcon     HICON
+}
+
+type GUID struct {
```
stffabi, line 21: IIRC this is a duplicate definition in another package, maybe also using `windows.GUID`?

v2/internal/platform/win32/structs.go
```diff
@@ -0,0 +1,54 @@
+package win32
+
+type NOTIFYICONDATA struct {
```
stffabi, line 3: IIRC this is a duplicate definition with a struct in another package, should we unify this?

v2/internal/platform/win32/winres.go
```diff
@@ -0,0 +1,7 @@
+package win32
+
+import "unsafe"
+
+func MakeIntResource(id uint16) *uint16 {
```
stffabi, line 5: I think using this one will make problems with the race detector as seen in: #1554

v2/pkg/application/systray.go
```diff
@@ -0,0 +1,151 @@
...
+		onLeftClick:        options.OnLeftClick,
+		onRightClick:       options.OnRightClick,
+		onLeftDoubleClick:  options.OnLeftDoubleClick,
+		onRightDoubleClick: options.OnRightDoubleClick,
+		onMenuOpen:         options.OnMenuOpen,
+		onMenuClose:        options.OnMenuClose,
+	}
+}
+
+func (t *SystemTray) run() {
```
stffabi, line 46: Don't we have to lock down the OS thread here? Since Windows will send the messages of the systray to the message loop of the OS thread that created the window. So we would have to make sure all of the creation of the SysTray will be done on the same OS thread as t.impl.Run() is run.

This is OS specific for windows, so we might need to put that into platform specific code behind a build tag.

stffabi (commented):
This looks awesome to me, awesome stuff 🚀.

I think the API is fine as it is, so from my side we can lock that down. 

Furthermore I added some comments regarding Windows message loop specific things. It seems like we are getting a bunch of duplicate code between winc and the new platform/win32 package. Maybe we should someday take the opportunity to clean that up, maybe the best time when we do multi-window support.

v2/internal/frontend/desktop/windows/win32/consts.go
```diff
@@ -41,3 +51,217 @@ func IsWindowsVersionAtLeast(major, minor, buildNumber int) bool {
 		windowsVersion.Minor >= minor &&
 		windowsVersion.Build >= buildNumber
 }
+
+// http://msdn.microsoft.com/en-us/library/windows/desktop/aa373931.aspx
+type GUID struct {
```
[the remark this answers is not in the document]
leaanthony: Are they 1:1?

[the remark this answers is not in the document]
stffabi: Yeah, it seems: https://cs.opensource.google/go/x/sys/+/refs/tags/v0.1.0:windows/types_windows.go;l=1698

v2/internal/platform/systray/windows.go
leaanthony: I did ask a win32 guy about this and he said that a message loop per thread is fine. What would moving this to a main loop look like? What if this is a tray only application with no window?

v2/internal/frontend/desktop/windows/win32/consts.go
```diff
@@ -13,16 +13,26 @@ type HANDLE uintptr
...
-	procSystemParametersInfo = moduser32.NewProc("SystemParametersInfoW")
-	procGetWindowLong        = moduser32.NewProc("GetWindowLongW")
-	procSetClassLong         = moduser32.NewProc("SetClassLongW")
-	procSetClassLongPtr      = moduser32.NewProc("SetClassLongPtrW")
-	procShowWindow           = moduser32.NewProc("ShowWindow")
-	procIsWindowVisible      = moduser32.NewProc("IsWindowVisible")
-	procGetWindowRect        = moduser32.NewProc("GetWindowRect")
-	procGetMonitorInfo       = moduser32.NewProc("GetMonitorInfoW")
-	procMonitorFromWindow    = moduser32.NewProc("MonitorFromWindow")
+	moduser32                       = syscall.NewLazyDLL("user32.dll")
```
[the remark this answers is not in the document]
stffabi: Yeah, will do that.

v2/internal/platform/systray/windows.go
leaanthony: Oooph that is a monumental catch 👍

v2/internal/platform/systray/windows.go
stffabi: Yes, it is not a problem per se. I'm just used to having only one MessageLoop from my old .NET days 😄. A drawback would be to have several dedicated OS threads that are being run for every MessageLoop.

AFAIK there's no problem of running a message loop if there's no window, it should run without any window. Let's consider the following case, which I think would also be easier to reason about with just one message loop. One has an application with just a tray, for one click another window is created (in the multi-window case). 

So the callback of the click is done in the dedicated systray os thread. If we now create a new window, we would need to span a go routine, create the window and start another message loop for that. Otherwise we would implicitly use the message loop of the systray. 

Let's consider the same case with just one message loop on the main-thread which is dedicated to the UI. In that case the systray is run in that message loop, a click callback is also run in the main-thread, one can simply create a new window without needing to span a go routine. Show the window and the window is also run on the same main-thread, so do all the callbacks. That would probably also make it easier to interact with WebView2 if there are several windows, since all the COM initialisation would only have to be done once.

stffabi:
> The winc/win32 thing I should have made clearer: everything in `platform` is the target from after we merge this PR. Eventually, as we migrate from `frontend` we move everything we need into `platform/win32` and drop both `windows/win32` and the `winc` directory. I'm looking at `frontend/win32` exposing a slightly higher interface than just a 1:1 api onto win32 so I wanted to "port" that over as needed. I hate the duplication but in my head they are different things 😅

Ah great, thanks for clarifying 🙏

teddywing:
> > Yes you’re right, it makes more sense for these behaviours to be defined by the app developer. Are there APIs to know when the menu is visible and to hide & show it?
> 
> There are now 👍

Nice, thanks!

leaanthony:
Closing in favour of #2181 - Thanks for the feedback 👍

Issue #1504, opened 2022-06-29
[Mac] Can not show the UI when exec 'wails dev'


cxt90730:
### Description

Can not show the UI when exec 'wails dev' (just an empty window),  but web ui works. Also, the app created by 'wails build' works too.

### To Reproduce

wails dev on my Mac(10.15 catalina)

### Expected behaviour

the UI when exec 'wails dev' works

### Screenshots

![image](https://user-images.githubusercontent.com/11382804/176369117-7d423da7-b920-47db-9c65-2ae1c2532b9b.png)

### System Details

```shell
❯ wails doctor
Wails CLI v2.0.0-beta.38


Scanning system - Please wait (this may take a long time)...Done.

System
------
OS:		MacOS
Version: 	10.15.7
ID:		199506
Go Version:	go1.17.6
Platform:	darwin
Architecture:	amd64

Wails
------
Version: 	v2.0.0-beta.38

Dependency			Package Name	Status		Version
----------			------------	------		-------
xcode command line tools 	N/A		Installed	2373
npm 				N/A		Installed	8.0.0
*upx 				N/A		Available
*nsis 				N/A		Available

* - Optional Dependency

Diagnosis
---------
Your system is ready for Wails development!
Optional package(s) installation details:
  - upx : Available at https://upx.github.io/
  - nsis : Available at https://nsis.sourceforge.io/Download
```

stffabi:
Thanks for reporting this issue and using Wails.

Just a couple of questions to hopefully narrow down the problem.
- Does it work with a clean Wails template? Like plain `wails init -t plain -d test -n test` or vanilla `wails init -t vanilla -d test -n test`
- Do you see anything of interest in the "DevTools", could be opened by right clicking into the empty window and clicking "Inspect Element"?
- Does the console output of `wails dev` show anything that is failing?

stffabi:
Thanks for your answers. Let's stick with the vanilla template case.

Can you please create again a new vanilla template in a new empty folder, just to make sure there's no mixed content. Also please make sure there's no other `wails dev` or npm running with your myproject UI. 

When running `wails dev` in this new folder, could you please provide or try the following?
- A screenshot of the "Network" tab of the dev tools from the Wails window.
- A copy of the complete `wails dev` output.
- What do you see when accessing `http://localhost:3000` with your normal browser?
- What do you see when accessing `http://localhost:34115` with your normal browser?

leaanthony:
Sounds like a vite error if plain works

[...]

PR #1960, opened 2022-10-12
Add support for `Info.dev.plist`

### LINKED ISSUE #1943: Support `Info.dev.plist` on macOS
There are a couple of scenarios where there is a need to add certain permissions in `Info.plist` for `wails dev` to work correctly. The current solution is to add those settings to `Info.plist` but this runs the risk of making it to a production build.

This ticket proposes to add an extra step during a Dev build:
- if `Info.dev.plist` doesn't exist, it creates it as a copy of `Info.plist`
- when packaging the app for Dev mode, it uses `Info.dev.plist` 

This will allow the developer to customise and experiment with the settings for Dev builds.

Previously discussed in #1504.

leaanthony:
Creates or uses the `Info.dev.plist` file as `Info.plist` when using `wails dev`.
Closes #1943 - Discussed in #1504


stffabi (approved):
LGTM 👍

Issue #1962, opened 2022-10-13
Fine-grained event listener API


joshbuddy:
### Is your feature request related to a problem? Please describe.

The existing event api allows for the deregistration of all event listeners of a certain type, but does not allow the deregistration of a single event listener. This feature is essential for react work where single components are responsible for registering and deregistering for events and should not interfere with each other.

### Describe the solution you'd like

The simplest solution would be to provide a function that allows the deregistration of a single listener. Perhaps something like:

```
/**
 * ListenerOff unregisters a listener previously registered with On
 *
 * @param {EventListener} listener
 */
 export function ListenerOff(listener) {
```

### Describe alternatives you've considered

An alternate version of this would be to move towards a more "active" pattern for event listeners, where it would work more like this.

```
const listener = CreateListener<EventType>("eventName", (event) => { ... })
listener.start();
listener.stop();
```

PR #1969, opened 2022-10-13
Add single listener deregistration

joshbuddy:
This provides an API to address this issue.

https://github.com/wailsapp/wails/issues/1962

My one question with this change is how do you test these sorts of changes?


v2/internal/frontend/runtime/desktop/events.js
```diff
@@ -50,11 +52,13 @@ export const eventListeners = {};
  * @param {string} eventName
  * @param {function} callback
  * @param {number} maxCallbacks
+ * @returns {Listener}
  */
 export function EventsOnMultiple(eventName, callback, maxCallbacks) {
```
avengerweb, line 57: You are missing EventsOn and EventsOnce

v2/internal/frontend/runtime/desktop/events.js
```diff
@@ -50,11 +52,13 @@ export const eventListeners = {};
...
     eventListeners[eventName] = eventListeners[eventName] || [];
-    const thisListener = new Listener(callback, maxCallbacks);
+    const thisListener = new Listener(eventName, callback, maxCallbacks);
     eventListeners[eventName].push(thisListener);
+    return thisListener;
```
avengerweb, line 61: probably better to return "cancel" callback instead, to avoid exposing internal Listener API

avengerweb (commented):
I also worked on similar PR

v2/internal/frontend/runtime/desktop/events.js
joshbuddy: Good catch, thanks

v2/internal/frontend/runtime/desktop/events.js
joshbuddy: This is the question I brought up in my issue. I can see advantages in having a more active style api, and returning the listener itself would facilitate that. Otoh, simply returning the cancel function is the 80% usecase, so I dig that simplicity.

leaanthony: Returning a cancel callback was the first thing I thought of too. How easy is it to manage those callbacks though? Are the 2 approaches mutually exclusive?

leaanthony:
This change would need an update to `runtime.d.ts` too. Have you considered how the Go API should be affected by this?

v2/internal/frontend/runtime/desktop/events.js
joshbuddy: So, I think it would be better to have a design discussion on the issue instead of the PR, but happy to do it here if you're really into that.

In short, if this was my design, I would have a low level Listener interface, and then some higher level functions on top that correspond to the existing api, but with a few tweaks. I would more closely align it with nodejs's eventemitter api with a goal of it being more idiomatic js.

joshbuddy:
I think I would not touch the go api for now, merely ship this, and then revisit the go api with an eye towards making it more go-like.

v2/internal/frontend/runtime/desktop/events.js
joshbuddy: And, I'm happy to make the changes to the api to more what I'm describing, just didn't want to actually do that unless people were into it.

avengerweb: > Returning a cancel callback was the first thing I thought of too. How easy is it to manage those callbacks though? Are the 2 approaches mutually exclusive?

I think main problem that approach with returning listener object will expose it when it is not intended to be exposed (it seems private interface with limited documentation).

v2/internal/frontend/runtime/desktop/events.js
```diff
@@ -74,6 +79,7 @@ export function EventsOn(eventName, callback) {
  * @export
  * @param {string} eventName
  * @param {function} callback
+ * @returns {function}
  */
 export function EventsOnce(eventName, callback) {
     EventsOnMultiple(eventName, callback, 1);
```
avengerweb, line 85: return is missing here.

leaanthony:
I think it's important to keep the API as consistent as possible between Go and JS. I think returning a function there is also ok?

joshbuddy:
@leaanthony I've duplicated the interface into go. I'm curious about your motivation for keeping things consistent vs idiomatic. For instance, with this go interface, there are potential data race issues, but perhaps a more go-centered interface could avoid those issues? Or we introduce a mutex around the listeners map.

v2/internal/frontend/runtime/desktop/events.js
```diff
@@ -173,4 +179,21 @@ export function EventsOff(eventName, ...additionalEventNames) {
...
+}
+
+/**
+ * ListenerOff unregisters a listener previously registered with EventsOn
+ *
+ * @param {Listener} listener
+ */
+ function ListenerOff(listener) {
+    // Remove local listener
+    eventListeners[eventName] = eventListeners[eventName].filter(l => l !== listener);
```
leaanthony, line 191: Should this be `eventListeners[listener.eventName]` ?

joshbuddy: Yup, it should be. Btw, how do you test these changes?

leaanthony:
Added a sample test for Go. Feel free to update with your own tests 👍

joshbuddy:
@leaanthony so, I haven't seen any place for js tests to go here. If there is I'll happily add them. If there isn't I'm going to hold off on adding tests to this pr, better to do it in a separate branch

leaanthony:
I'm happy for you to add a testing framework for JS to this PR. There isn't too much to test. I think my preference would be vtest

joshbuddy:
@leaanthony sure, sounds good! Any thoughts on integration tests themselves? I couldn't see any work in that direction

leaanthony:
Open to ideas. If you mean between Go and JS, then yeah... Nobody is sure yet. I've been meaning to check out https://nutjs.dev/ to see if that might help. 

I think we may need a "test" target that builds against a headless frontend that we can query. It's pretty tricky.

avengerweb:
> Open to ideas. If you mean between Go and JS, then yeah... Nobody is sure yet. I've been meaning to check out https://nutjs.dev/ to see if that might help.
> 
> I think we may need a "test" target that builds against a headless frontend that we can query. It's pretty tricky.

Probably is best idea to start with unit tests for core functionality, then integration tests using browser and after that time will come to e2e tests. I did small research some time ago for my own apps; however, there is no good solution for e2e tests of webview that cover all 3 targets (Windows, MacOS, Linux). It easy for Windows and Linux as they use Chromium but MacOS will be left aside.

joshbuddy:
Sounds like writing good integrations tests is a big unknown but high value if we can make them robust. Anyhoo, lets finish off this work before figuring out stuff like that.

@leaanthony thanks for the vitest reco. Very pleasant test framework to use!

v2/internal/frontend/runtime/desktop/events.js
```diff
@@ -107,7 +113,11 @@ function notifyListeners(eventData) {
         }
 
         // Update callbacks with new list of listeners
-        eventListeners[eventName] = newEventListenerList;
+        if (newEventListenerList.length === 0) {
+            removeListener(eventName);
```
joshbuddy, line 117: @leaanthony hey, note this line, I had to add this so WailsInvoke got called with `EX${eventName}`, which I think was the intent (as I captured here https://github.com/wailsapp/wails/pull/1969/files#diff-eda323563cc3726a10525d8ae664e9cd8ba98fce7d961ed8b8d6c634670506d1R39). Does that look right to you?

leaanthony (commented):
Is this still a work in progress?

v2/internal/frontend/runtime/desktop/events.js
leaanthony: Yeah I think so 👍

joshbuddy:
@leaanthony I think this is ready

leaanthony:
Thanks so much for taking the time to do this! A couple of minor things and I think we should be good to go:

- https://github.com/wailsapp/wails/blob/master/v2/internal/frontend/events.go will need updating with the API changes
- https://github.com/wailsapp/wails/blob/master/website/docs/reference/runtime/events.mdx will need updating with the changes
-

joshbuddy:
@leaanthony how's that look?

leaanthony:
@joshbuddy - yeah cool. I think the first link wasn't updated (same filename - easy to confuse!).

joshbuddy:
@leaanthony Hey, then I'm not clear what you mean by updating https://github.com/wailsapp/wails/blob/master/v2/internal/frontend/events.go. This is the current state of that file:

https://github.com/wailsapp/wails/blob/e4eea15a6b43e4103342c6294bebdef979fdd467/v2/internal/frontend/events.go

leaanthony (approved):

leaanthony:
Oh that's so weird - I must have mistakenly been on a different branch 🤯

leaanthony:
Thanks so much for this @joshbuddy !

PR #1974, opened 2022-10-14
[webview2loader] Add full featured go implementation

stffabi:
The new go loader can be activated with the exp_gowebview2loader build tag.


leaanthony:
This is incredible! Might take a while for me to get through it but I'll aim to do that this weekend!

v2/internal/frontend/desktop/windows/go-webview2/webviewloader/env_create.go
```diff
@@ -0,0 +1,159 @@
...
+
+	return nil
+}
+
+type environmentCreatedHandler struct {
+	originalHandler ICoreWebView2CreateCoreWebView2EnvironmentCompletedHandler
+}
+
+func (r *environmentCreatedHandler) EnvironmentCompleted(errorCode HRESULT, createdEnvironment *ICoreWebView2Environment) HRESULT {
+	// The OpenWebview2Loader has some retry logic and retries once, didn't encounter any case when this would have been
```
leaanthony, line 130: Perhaps @jchv has more insight

v2/internal/frontend/desktop/windows/go-webview2/pkg/combridge/bridge.go
```diff
@@ -0,0 +1,221 @@
...
+	"sync/atomic"
+)
+
+var (
+	comIfcePointersL sync.RWMutex
+	comIfcePointers  = map[uintptr]*comObject{} // Map from ComInterfacePointer to the Go ComObject
+)
+
+// Resolve the GoInterface of the specified ComInterfacePointer
+func Resolve[T IUnknown](ifceP uintptr) T {
```
leaanthony, line 16: Love this approach 🚀

v2/internal/frontend/desktop/windows/go-webview2/pkg/combridge/bridge.go
```diff
@@ -0,0 +1,221 @@
...
+	return o.obj.queryInterface(guidOf[T](), false)
+}
+
+// Close releases the native com object from the go side. It will only be destroyed if the ref counter
+// reaches zero.
+// After closing `Ref()` will panic.
+func (o *ComObject[T]) Close() error {
+	if atomic.CompareAndSwapInt32(&o.closed, 0, 1) {
+		runtime.SetFinalizer(o, nil)
+		o.obj.release()
```
stffabi, line 132: This should be done in a new goroutine because it acquires locks internally, which might block the finalizer goroutine for a longer time.
Will add a new commit with a fix for that.

stffabi:
> This is incredible! Might take a while for me to get through it but I'll aim to do that this weekend!

Thanks so much, was a great experience with all the com specific stuff 😀

leaanthony:
Never thought I'd ever see "com" and "great experience" in the same sentence 😂 Awesome stuff 👍

stffabi:
> Never thought I'd ever see "com" and "great experience" in the same sentence 😂 Awesome stuff 👍 

That was more meant from a technical point of view of interacting with com and playing nicely together with the go runtime 😀

v2/internal/frontend/desktop/windows/go-webview2/pkg/combridge/bridge.go
stffabi: Just pushed the fix for this one.

v2/internal/frontend/desktop/windows/go-webview2/pkg/edge/create_env_native.go
```diff
@@ -0,0 +1,41 @@
...
+	if err != nil {
+		return fmt.Errorf("Error calling UTF16PtrFromString for %s: %v", browserExecutableFolder, err)
+	}
+
+	userPathPtr, err := windows.UTF16PtrFromString(userDataFolder)
+	if err != nil {
+		return fmt.Errorf("Error calling UTF16PtrFromString for %s: %v", userDataFolder, err)
+	}
+
+	hr, err := webviewloader.CreateCoreWebView2EnvironmentWithOptions(
```
leaanthony, line 26: Both the native and go versions of this call have the same package and name but have different signatures. It took me a little time to realise what was going on. Do you think changing the method name of one of the versions would be a bad idea (would that make it inconsistent)?

leaanthony (approved):
Looking really good! I've left a couple of very minor questions that are no biggies. 

What do you think we need to have in the documentation for people to try it out and get feedback? Is this a flag you'd pass with the `-flags` parameter for `wails build`? 

I noticed the comment on supporting all the features so we could do away with the dll. What scenarios would we still need it?

v2/internal/frontend/desktop/windows/go-webview2/pkg/edge/create_env_native.go
stffabi: My idea was to keep the method names of the original loader so people can match this to the original MS documentation to get more detailed information of the methods.

stffabi: I can change the names if you want.

v2/internal/frontend/desktop/windows/go-webview2/pkg/combridge/bridge.go
```diff
@@ -0,0 +1,233 @@
...
+// New returns a new ComObject which implements the specified Com Interface, com calls will be redirected
+// to the specified go interface.
+func New[T IUnknown](obj T) *ComObject[T] {
+	cObj := new(
+		ifceDef[T]{obj},
+	)
+	return newComObject[T](cObj)
+}
+
+// New returns a new ComObject which implements the specified Com Interface, com calls will be redirected
```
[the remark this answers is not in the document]
stffabi: Nice catch, I've enhanced the comment for the multiple inheritance cases.

v2/internal/frontend/desktop/windows/go-webview2/pkg/combridge/bridge.go
```diff
@@ -0,0 +1,233 @@
...
+// to the specified go interface.
+func New2[T IUnknown, T2 IUnknown](obj T, obj2 T2) *ComObject[T] {
+	cObj := new(
+		ifceDef[T]{obj},
+		ifceDef[T2]{obj2},
+	)
+	return newComObject[T](cObj)
+}
+
+// New returns a new ComObject which implements the specified Com Interface, com calls will be redirected
```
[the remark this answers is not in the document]
stffabi: Nice catch, I've enhanced the comment for the multiple inheritance cases.

v2/internal/frontend/desktop/windows/go-webview2/pkg/edge/create_env_native.go
leaanthony: Nah, I got it after a bit and understand that rationale 👍

stffabi:
> Looking really good! I've left a couple of very minor questions that are no biggies.

Thanks for reviewing it ❤️ 

> What do you think we need to have in the documentation for people to try it out and get feedback? Is this a flag you'd pass with the `-flags` parameter for `wails build`?

It's a build tag to be passed as -tags: `wails build --tags exp_webview2loader`. As for documentation I hadn't a good idea where to put it, do you have a good idea?

> I noticed the comment on supporting all the features so we could do away with the dll. What scenarios would we still need it?

The following features are not supported and I think we explicitly don't want or need them:
- Registry Overrides of Parameters: I don't think we want that someone (enduser) can just override the WebView2 options to be used by changing the registry key which maps to the binary. This would e.g. also allow to active the debugger of the webview.
- Env Variable Overrides of Parameters: Same as above, from a security perspective I think even more dangerous. This would also allow to active the debugger by just starting the binary with a env variable. We already disable this for the native loader as well.
- Does not incorporate `GetCurrentPackageInfo` to search for an installed runtime: This would allow override the runtime to be when bundled in an appx. So I think not relevant for Wails.

I think the options to be used should always be set by the Wails application and not by some user, which would allow to inspect internals of the app or change the behaviour of it.

leaanthony:
I think we could add these details to the output of `wails build` to invite people to trial it?

stffabi:
Oh, yeah, that's a great idea. Will add that...

stffabi:
Just the information to `wails build`, what do you think about that?

v2/internal/frontend/desktop/windows/go-webview2/webviewloader/env_create.go
jchv: Unfortunately, I do not. I also never noticed a case where it is needed. If there *is* a good reason for it, one would probably have to peel back more layers to find it.

stffabi: Thanks for your reply. So I think we could leave this as it is for now and see what the feedback will be from the users.

leaanthony:
Yeah, that's great. I'm wondering if we should open a ticket/discussion for capturing the feedback and have that link at the bottom. The issue/discussion should have a bit of detail there: What is the benefit of using it, gotchas, maybe a test matrix so people can report it works with `wails doctor` output. Thoughts?

v2/internal/frontend/desktop/windows/go-webview2/webviewloader/env_create.go
leaanthony: Thanks for taking the time to leave a comment @jchv - appreciate it 🙏

stffabi:
Yeah, that sounds great.

stffabi:
@leaanthony just added the link to the Feedback - Issue. For the content of the issue we could discuss that in slack.

So I think we are good to go with this one, what do you think?

stffabi:
Awesome, thanks for merging.

PR #1977, opened 2022-10-15
chore: update workflows

misitebao:
- Fix check document workflow
- Add workflow to auto push to crowdin after website update

**Work before the merger**: Add an Actions secret called `CROWDIN_PERSONAL_TOKEN` with the value of your Crowdin [Personal Access Tokens](https://crowdin.com/settings#api-key)


leaanthony:
Did you mean to delete `build.yml`?

misitebao:
@leaanthony 
Yes, because the `build_and_test.yml` script has it fully covered.

leaanthony:
I was looking at the fact that we are setting the crowdin token in the GITHUB_TOKEN env so went to look at crowdin and found this: https://developer.crowdin.com/github-crowdin-action/ - This seems like the official way of doing this?

misitebao:
@leaanthony 
Both ways can be achieved, and both require setting environment variables, but environment variables cannot be stored in the repository in plaintext, so you can only set `Actions Secret`, and then use it in the workflow file.

leaanthony:
I couldn't find any docs saying that setting GITHUB_TOKEN to the Crowdin token does anything.

.github/workflows/push-to-crowdin.yml
```diff
@@ -0,0 +1,38 @@
...
+
+      - name: Set node
+        if: steps.verify-changed-files.outputs.files_changed != 'true'
+        uses: actions/setup-node@v2
+        with:
+          node-version: 16.x
+
+      - name: Push files
+        env:
+          GITHUB_TOKEN: ${{ secrets.CROWDIN_PERSONAL_TOKEN }}
```
leaanthony, line 33: I'm trying to understand what this is for. How does the Crowdin cli read this in?

misitebao: Oh sorry, the key should be: `CROWDIN_PERSONAL_TOKEN`, my mistake, I will update this branch later.

leaanthony (approved):

PR #1989, opened 2022-10-18
[dragging] Trim the css property value

stffabi:
Webview2 on Windows returns a potential whitespace when defining the style like this `style="--wails-draggable: drag"`. Whereas Darwin already trims the value and returns "drag" and not " drag".


leaanthony (approved):
Nice catch 👏

stffabi:
Thanks for approving.

PR #2000, opened 2022-10-20
[Docs] Add troubleshooting guide for unable to start service

willdot:
This happened to me when I migrated my work to a new work laptop. Being someone that's not savvy with frontend work (node modules and package lock etc) it never occurred to me that this would be the problem.

Deleting these files and folders and then adding them to my .gitignore has worked for me (I've moved it to another machine to test and it works).


leaanthony:
Just realised (hard to see on mobile) that this is in the versioned_docs directory. Any chance you could add it to the path under `docs`? This is entirely on me and I need to make a contribution guide for the docs. The reason is the docs dir is the source of truth and the others are snapshots that will expire and I don't want to lose this.

Happy to do this myself over the weekend though 👍

willdot:
> Just realised (hard to see on mobile) that this is in the versioned_docs directory. Any chance you could add it to the path under `docs`? This is entirely on me and I need to make a contribution guide for the docs. The reason is the docs dir is the source of truth and the others are snapshots that will expire and I don't want to lose this.
> 
> Happy to do this myself over the weekend though 👍

Done

Issue #1877, opened 2022-09-22
Rename `frontend` / modularize project


carljkempe:
### Is your feature request related to a problem? Please describe.

Well, I was having trouble combining my freshly initialized `wails` app within my very well split back-end (_BE_).  Had to re-approach it, but the other way around and make my _BE_ fit into the folder structure of the `wails` app instead!  Obviously this isn't too big of a concern, but it still does make a valid suggestion? 😅

Two main questions stemmed from this:

### Describe the solution you'd like

1) Are we, or should we be able to rename the `frontend` directory?

2) Is there currently a way, or should there be a way for us developers to integrate `wails` into our per-existing, neat, folder structures?  (I was initially trying to split my `cmd/` dir into `cmd/client` and `cmd/server`, instead of leaving a "mess" in the root dir for `wails` alone.)

### Describe alternatives you've considered

I'm away from it all, so I won't be entirely specific; but I think that this might be appreciated by many!

### Additional context

![2022-09-25-151636_1920x1080_scrot](https://user-images.githubusercontent.com/110561976/192167984-d0f40f7e-fe68-4d41-9fa8-00a6e2cb27ab.png)

olup:
I am interested by 2 as well. I wanted to build my golang library as one cmd and the app on top of it as another one, but it did not work out, and I ended up creating two repos and go getting my library from github to my wails app.

mholt:
Subscribing, as I'm also integrating into an existing app and having to shuffle files around and refactor was not ideal.

Very similar to https://github.com/wailsapp/wails/issues/615

Basically, it'd be nice to:

- Have go.mod in a parent folder
- Rename "frontend"
- Have "frontend" be somewhere else

leaanthony:
@mholt @olup @carljkempe - Please try the above draft PR to see if this fixes your problem. If you are unsure how to do this, please follow [this guide](https://wails.io/docs/guides/bleeding-edge#testing-a-pr). Let us know how you get on 👍

mholt:
Works well for what I need! Thank you!

leaanthony:
Master now has the fixes for this. Please test using [this guide](https://wails.io/docs/guides/bleeding-edge). If it doesn't work as expected, please reopen 👍

mholt:
Awesome! Thank you again!!

olup:
Just refactored https://github.com/olup/lunii-admin to use this, going great ! Thanks a lot 🚀

learncodingforweb:
[...]

PR #2009, opened 2022-10-23
Modularize wails

leaanthony:
This PR aims to make a Wails project more configurable as requested in #1877.
Changes:
  - A new `frontend:dir` entry in `wails.json`. This can be an absolute path or a relative path from the `wails.json` file. The last segment of this path is considered the name of the frontend directory. Not specifying this will result in the surrent default behaviour.
  - If `go.mod` is not found in the same directory as `wails.json`, the cli will look up the parent directories to find it.

Note: The CLI commands `build` and `dev` must still be run in the same directory as `wails.json`.

Example project in `v2/examples/customlayout`


v2/internal/project/project.go
```diff
@@ -36,7 +37,7 @@ type Project struct {
 	/*** Internal Data ***/
 
 	// The path to the project directory
-	Path string
+	Path string `json:"projectdir"`
```
stffabi, line 40: We are now exposing this to the user, what's the meaning of this folder? AFAIK it's the folder where the `main.go` could be found, should we use a better naming for this?
Should this also be relative to the location of `wails.json` like frontendir and wailsjsdir? This would make it possible to put wails.json in the root of a project and put main.go of wails in e.g. `cmd/desktop`.

v2/internal/project/project.go
```diff
@@ -88,6 +89,23 @@ type Project struct {
...
+
+	// Frontend directory
+	FrontendDir string `json:"frontend:dir"`
+}
+
+func (p *Project) GetFrontendDir() string {
+	if filepath.IsAbs(p.FrontendDir) {
+		return p.FrontendDir
+	}
+	return filepath.Join(p.Path, p.FrontendDir)
```
stffabi, line 101: Since we are exposing `p.Path` to the user `p.FrontendDir` is not always relative to where `wails.json` is found. It might be relative to that or relative to `projectdir`.
Would it be more consistent and easier to reason about this folder if it is always relative to `wails.json` and `projectdir` is also relative to `wails.json`?
So every folder defined in `wails.json` would be relative to the location if the file itself.

v2/internal/project/project.go
```diff
@@ -88,6 +89,23 @@ type Project struct {
...
+}
+
+func (p *Project) GetWailsJSDir() string {
+	if filepath.IsAbs(p.WailsJSDir) {
+		return p.WailsJSDir
+	}
+	return filepath.Join(p.Path, p.WailsJSDir)
```
stffabi, line 108: See comment on `GetFrontendDir`

v2/internal/project/project.go
```diff
@@ -119,6 +137,73 @@ func (p *Project) Save() error {
...
+	}
+	if p.Info.Copyright == nil {
+		v := "Copyright........."
+		p.Info.Copyright = &v
+	}
+	if p.Info.Comments == nil {
+		v := "Built using Wails (https://wails.io)"
+		p.Info.Comments = &v
+	}
+	if p.DevServer == "" {
```
stffabi, line 200: Seems like a duplicate of line 170.

v2/internal/project/project.go
```diff
@@ -119,6 +137,73 @@ func (p *Project) Save() error {
...
+		p.WailsJSDir = p.FrontendDir
+	}
+	if p.DebounceMS == 0 {
+		p.DebounceMS = 100
+	}
+	if p.DevServer == "" {
+		p.DevServer = "localhost:34115"
+	}
+	if p.NSISType == "" {
+		p.NSISType = "portable"
```
stffabi, line 181: IIRC portable is not a valid value for NSISType, it should be one of "single"/"multiple" and the default was empty which is mapped to "multiple".

```suggestion
		p.NSISType = "multiple"
```

v2/internal/project/project.go
```diff
@@ -119,6 +137,73 @@ func (p *Project) Save() error {
...
+	}
+	if p.Version == "" {
+		p.Version = "2"
+	}
+	// Create default name if not given
+	if p.Name == "" {
+		p.Name = "wailsapp"
+	}
+	if p.AssetDirectory == "" {
+		p.AssetDirectory = "assets"
```
stffabi, line 160: Won't this break all projects that don't use a frontend devserver and rely on inferring of the AssetDirectory? The default until now was empty, which would be to infer it from the assets embed.FS.

stffabi (commented):
Great enhancement 🚀 , left some minor comments.

v2/pkg/buildassets/buildassets.go
```diff
@@ -83,6 +80,7 @@ func ReadFileWithProjectData(projectData *project.Project, file string) ([]byte,
 // ProjectInfo if necessary.
 // It will also write the resolved final file back to the project build folder.
 func ReadOriginalFileWithProjectDataAndSave(projectData *project.Project, file string) ([]byte, error) {
+	rootFolder := filepath.Base(projectData.GetBuildDir())
 	file = path.Join(rootFolder, file)
 	content, err := iofs.ReadFile(assets, file)
```
stffabi, line 85: Same comment as above, I would say this should be: `content, err := iofs.ReadFile(assets, path.Join("build", file))` where file is the unchanged file argument.

v2/pkg/buildassets/buildassets.go
```diff
@@ -124,15 +122,16 @@ func resolveProjectData(content []byte, projectData *project.Project) ([]byte, e
 }
 
 func writeFileSystemFile(projectData *project.Project, file string, content []byte) error {
-	path := filepath.Clean(filepath.Join(projectData.Path, filepath.FromSlash(file)))
+	baseDir := filepath.Dir(projectData.GetBuildDir())
+	targetPath := filepath.Clean(filepath.Join(baseDir, filepath.FromSlash(file)))
```
stffabi, line 126: With the above changed code, file should always be a slash path relative to BuildDir, in which case the targetPath could just be: `targetPath := filepath.Join(projectData.GetBuildDir(), filepath.FromSlash(file))`

v2/pkg/commands/build/desktop.go
```diff
@@ -20,9 +20,10 @@ func newDesktopBuilder(options *Options) *DesktopBuilder {
 func (d *DesktopBuilder) BuildAssets(options *Options) error {
 
 	// Check assets directory exists
-	if !fs.DirExists(options.ProjectData.BuildDir) {
+	buildDir := options.ProjectData.GetBuildDir()
+	if !fs.DirExists(buildDir) {
 		// Path to default assets
-		err := buildassets.Install(options.ProjectData.Path)
+		err := buildassets.Install(buildDir)
```
stffabi, line 26: Doesn't `buildassets.Install` already add a `build` dir to the path since the `assets embed.FS` from build dir contains that root folder?

v2/internal/project/project.go
leaanthony: Yeah, this looks like it was left over from a previous branch I cherry picked :/ I'll take this out.

stffabi: Great, thanks for clarifying. So then some of my comments will become obsolete 😀

v2/internal/project/project.go
leaanthony: Yes, I agree - every relative path in `wails.json`. Right now `p.Path` *is* where `wails.json` is (after I remove "projectdir") so I think it's good for now. I think eventually, `project.Load()` will be path-less and use cwd by default. It will traverse the filesystem to find it. Once found, it will load and write this location to the project data internally. This method will calculate based off that.

v2/internal/project/project.go
leaanthony: WTF copilot! Oooph. Was this even set before?

v2/pkg/commands/build/desktop.go
leaanthony: It does not, though I can see why it looks like it would. Running `wails build` in `examples/customlayout/cmd/customlayout` works as expected.

leaanthony:
I feel a CLI rewrite coming on 😜

v2/internal/project/project.go
stffabi: Yeah, with removing projectdir that's fine.

v2/internal/project/project.go
stffabi: It was empty, which mapped to multiple.

v2/pkg/commands/build/desktop.go
stffabi: That's interesting, I'm confused on that with diff of the PR. Before we passed e.g. `options.ProjectData.Path = ./` and it wrote it into `./build/`, now we are passing `./build/` and it also writes it into `./build/`.

Does that method `desktopBuilder.BuildAssets` even get called?

stffabi:
Just added a commit which should fix the buildassets comments.

mholt:
Hey, this works for what I need! :partying_face: 

I only tested it briefly, thanks for the great instructions. But so far it's quite nice.

leaanthony:
Thanks for pushing for this. It really has been on the backburner for quite some time so good to get it done finally 🙏

mholt:
Thank you for working on it. Wow, this was not a small change!

leaanthony:
@mholt @stffabi - you think it's worth finishing this off with a `bin:dir` too or are we good with `${build:dir}/bin`?

v2/internal/project/project.go
stffabi: Just saw this conversation was marked as resolved. I think we should still remove this default value to prevent problems for those users.

stffabi:
I think we are good with `${build:dir}/bin` for the time being.

v2/internal/project/project.go
leaanthony: Not sure if it's just threading issues on GitHub mobile but your reply is to a thread that's unresolved.

leaanthony:
Thanks for explaining. Yeah, should use same build dir 👍

mholt:
One thing I just thought of that I am curious your thoughts on, is that if I do change my `frontend:dir` (let's say, I'm testing something real quick) I also need to update my `go:embed` tag on my `embed.FS` value. If these are out of sync things get weird real quick.

It could be nice if Wails could infer the frontend dir from the `go:embed` on the fs.FS set on `options.App.Assets` (soon to be `options.App.AssetServer.Assets`), but since `Assets` takes an `fs.FS` and not an `embed.FS`, I guess there's no guarantee it's a real folder in the first place.

So in summary, I don't know an easy way to infer the frontend dir, and this is not a big deal to me, and definitely not a blocker for this PR or the next release. But I wanted to document this nuance here for the record.

Thanks again for this super handy improvement!

stffabi:
It's pretty hard to infer that, since it is currently needed pretty early in the build process. So one would either need to do static analysis or build the binary in some special mode and execute it to get the inferred directory if the assets are an embed.FS. The problem is also how to find the root directory when inferring.

IIRC the frontend dir is also not exactly the same as the assets dir since the assets dir would point to the output dir of the frontend, e.g. frontend/dist.

I wonder if we could change that to some code generation method. We could add the assetdir again to wails.json which is the only source of truth for it. The build command could then create a 'assets.go' where 'main.go' is and with the correct go embed content.
'
//go:embed all:frontend/dist
var assets embed.FS
'

Would only work if the frontend dir is a child of the project dir. If it's in another dir one could add the embed there and add an import in the assts.go in the main package and referring to the assets in the other package.

Just a quick thought without thinking too much about the details and consequences...

mholt:
@leaanthony Yep, totally understand. No problem.

@stffabi Ah yeah, you're right... I forget that `frontend` and `frontend/dist` ("Assets") are different for most people, as I have a vanilla JS frontend that doesn't use build steps. So to me they are the same.

Anyway, carry on. :smile:

leaanthony:
> @leaanthony The dev build from `wails dev` is appearing in the folder from which I run `wails dev` (in my case, `cmd/timelinize`). It has the binary in it and then after I exit the program it goes away but the `build/bin` folders remain, empty.
> 
> I think they're supposed to be in the root of the project, right? (the build dir is at the root in the template project, at least)

@mholt - I am unable to reproduce this! The `dev` exe is appearing in the same place as the production binary for me with the `-dev` suffix 😖

mholt:
@leaanthony Ok, no worries then! I probably misconfigured something in my setup. I wouldn't worry about it right now or let it be a blocker.

leaanthony:
No worries @mholt. Just let me know if you can reproduce with a clean project and we'll fix it!

olup:
Just wanted to add a useless comment sharing how exciting this is, I can't wait to finally refactor my app to include all sub modules alongside the wails code !

stffabi (approved):
LGTM 🚀

W1M0R:
Just leaving this note here for others that might have the same issue.

The `../` path does not seem be supported by `go:embed`. So your frontend will still have to live next to or below the `wails.json` file:

```go
// THE FOLLOWING WON'T WORK

//go:embed all:../frontend/dist
var assets embed.FS
```

Issue #1613, opened 2022-07-20
Global recovery option


matt9mg:
### Is your feature request related to a problem? Please describe.

There are situations where the app may work running wails dev but once compiled the app may crash for whatever reason maybe someone forgot to embed a file system or something else.

### Describe the solution you'd like

 The ability to add recovery middleware to the run section of  wails.Run perhaps as the &options.App{ that is passed  this would allow additional crash logs to be added or sent to third part systems.

leaanthony:
Thanks for opening a ticket! Great idea. Happy to accept a PR for this 👍

stffabi:
I have used [panicwrap](https://github.com/mitchellh/panicwrap) in combination with Wails in one project.

There are panics which one might not recover from, e.g. if the panic happens in some go routine that get spanned outside of the control e.g. in a 3rd party lib. Or there might be segfaults wich lead to panics that occur in native syscall, which can't always be recovered.

leaanthony:
Closing this due to inactivity. There's a solution here for v2 and we are supporting it natively in v3 👍

PR #2016, opened 2022-10-25
[assetserver] Introduce middleware and extract options

### LINKED ISSUE #2007: More control over frontend

All of my frontend pages (on an existing project) need server-side template processing (I use Go's `text/template` and `html/template` packages), including index.html. Thus I need to be able to render them in my own ServeHTTP, not a built-in handler implementation.

I'm not sure the best solution as I don't know enough about the internals, but maybe a way that my own http.Handler can serve _all_ the assets instead of as a "fallback" would be enough. Then I could use index.html and wouldn't need to rewrite things (but could if I wanted to).

My first attempt was to set `Assets` to nil in the app options and set an `AssetsHandler` to my own HTTP handler, but the built-in handler still magically takes over. Maybe it finds the embedded FS on its own? Anyway, I was thinking I could skirt around the built-in handler if the index.html file 404'ed, thus calling my AssetsHandler, as long as I make sure every link goes to a page that doesn't exist. I just have to rewrite the URL server-side, internally. No problem.

But I'm having a bit of trouble with this, since index.html is a required asset:

```
    Could not resolve entry module (index.html).
    error during build:
    Error: Could not resolve entry module (index.html).
        at error (...project/frontend/node_modules/rollup/dist/shared/rollup.js:198:30)
        at ModuleLoader.loadEntryModule (...project/frontend/node_modules/rollup/dist/shared/rollup.js:22483:20)
        at async Promise.all (index 0)
```

This means I can't rewrite requests to `/` or `/index.html` to, say `_index.html` or `myindex.html` because the program won't even build without a real index.html file. Which means I can't serve it using my own handler, and can't evaluate templates server-side.

### Additional context

Thank you for Wails, it's quite awesome :D

stffabi:
This PR introduces a HTTP middleware for the AssetServer, furthermore it changes the API for defining the AssetServer config.

Fixes #2007

### DIFF
--- a/v2/internal/frontend/assetserver/common.go
+++ b/v2/internal/frontend/assetserver/common.go
@@ -8,8 +8,25 @@ import (
 	"net/http"
 	"strings"
 
+	"github.com/wailsapp/wails/v2/pkg/options"
+	"github.com/wailsapp/wails/v2/pkg/options/assetserver"
 	"golang.org/x/net/html"
 )
+
+func BuildAssetServerConfig(options *options.App) assetserver.Options {
+	if opts := options.AssetServer; opts != nil {
+		if options.Assets != nil || options.AssetsHandler != nil {
+			panic("It's not possible to use the deprecated Assets and AssetsHandler options and the new AssetServer option at the same time. Please migrate all your Assets options to the AssetServer option.")
+		}
+
+		return *opts
+	}
+
+	return assetserver.Options{
+		Assets:  options.Assets,
+		Handler: options.AssetsHandler,
+	}
+}
 
 const (
 	HeaderHost          = "Host"
--- a/v2/pkg/options/assetserver/middleware.go
+++ b/v2/pkg/options/assetserver/middleware.go
@@ -1,0 +1,20 @@
+package assetserver
+
+import (
+	"net/http"
+)
+
+// Middleware defines a HTTP middleware that can be applied to the AssetServer.
+// The handler passed as next is the next handler in the chain. One can decide to call the next handler
+// or implement a specialized handling.
+type Middleware func(next http.Handler) http.Handler
+
+// ChainMiddleware allows chaining multiple middlewares to one middleware.
+func ChainMiddleware(middleware ...Middleware) Middleware {
+	return func(h http.Handler) http.Handler {
+		for i := len(middleware) - 1; i >= 0; i-- {
+			h = middleware[i](h)
+		}
+		return h
+	}
+}
--- a/v2/pkg/options/assetserver/options.go
+++ b/v2/pkg/options/assetserver/options.go
@@ -1,0 +1,35 @@
+package assetserver
+
+import (
+	"io/fs"
+	"net/http"
+)
+
+// Options defines the configuration of the AssetServer.
+type Options struct {
+	// Assets defines the static assets to be used. A GET request is first tried to be served from this Assets. If the Assets returns
+	// `os.ErrNotExist` for that file, the request handling will fallback to the Handler and tries to serve the GET
+	// request from it.
+	//
+	// If set to nil, all GET requests will be forwarded to Handler.
+	Assets fs.FS
+
+	// Handler will be called for every GET request that can't be served from Assets, due to `os.ErrNotExist`. Furthermore all
+	// non GET requests will always be served from this Handler.
+	//
+	// If not defined, the result is the following in cases where the Handler would have been called:
+	//   GET request:   `http.StatusNotFound`
+	//   Other request: `http.StatusMethodNotAllowed`
+	Handler http.Handler
+
+	// Middleware is a HTTP Middleware which allows to hook into the AssetServer request chain. It allows to skip the default
+	// request handler dynamically, e.g. implement specialized Routing etc.
+	// The Middleware is called to build a new `http.Handler` used by the AssetSever and it also receives the default
+	// handler used by the AssetServer as an argument.
+	//
+	// If not defined, the default AssetServer request chain is executed.
+	//
+	// Multiple Middlewares can be chained together with:
+	//   ChainMiddleware(middleware ...Middleware) Middleware
+	Middleware Middleware
+}
--- a/v2/pkg/options/options.go
+++ b/v2/pkg/options/options.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"runtime"
 
+	"github.com/wailsapp/wails/v2/pkg/options/assetserver"
 	"github.com/wailsapp/wails/v2/pkg/options/linux"
 	"github.com/wailsapp/wails/v2/pkg/options/mac"
 	"github.com/wailsapp/wails/v2/pkg/options/windows"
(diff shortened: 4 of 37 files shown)


v2/pkg/options/assetserver/options.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+// Options defines the configuration of the AssetServer.
+type Options struct {
+	// FSys defines the static assets to be used. A GET request is first tried to be served from this FSys. If the FSys returns
+	// `os.ErrNotExist` for that file, the request handling will fallback to the Handler and tries to serve the GET
+	// request from it.
+	//
+	// If set to nil, all GET requests will be forwarded to Handler.
+	FSys fs.FS
```
stffabi, line 15: I think naming this 'Assets' like before would be better.

v2/internal/frontend/assetserver/common.go
```diff
@@ -8,9 +8,31 @@ import (
...
 
+func IsAssetFsDefined(options *options.App) bool {
+	opts := BuildAssetServerConfig(options)
+	return opts.Assets != nil
+}
+
+func BuildAssetServerConfig(options *options.App) *assetserver.Options {
+	if opts := options.AssetServer; opts != nil {
+		if options.Assets != nil || options.AssetsHandler != nil {
+			panic("It's not possible to use the deprecated Assets and AssetsHandler options and the new AssetServer option at the same time. Please migrate all your Assets options to the AssetServer option.")
```
leaanthony, line 24: In the case of windows, this panic may not be visible to the developer. That's not a problem as I'm sure it'll be obvious to work it out but as a general thought, I'm wondering if we should have a global recovery which dumps the error to a `panic.txt` and still panics.

stffabi: Yeah, that's a good point always forget about that on Windows. We have a ticket for the global recovery #1613

stffabi: What about adding a MessageBox for windows in this case?

leaanthony: Yeah, I think it's ok as is - just a related thought 👍

leaanthony:
I'm keen to know how this plays into the "server/hybrid target" that @tmclane is aiming for.

stffabi:
> I'm keen to know how this plays into the "server/hybrid target" that @tmclane is aiming for.

That should also work well there, since the internal `AssetServer/AssetHandler` takes care of that and AFAIK @tmclane is also using that.

v2/internal/frontend/assetserver/common.go
```diff
@@ -8,9 +8,26 @@ import (
...
+		}
+
+		return opts
+	}
+
+	return &assetserver.Options{
```
leaanthony, line 25: Considering this is a small struct and it's not getting called a whole heap of times, I'm wondering if it's worth changing the signatures to accept a struct rather than a struct pointer? At this point in the code, it's gone from being optional to a mandatory struct. Just a thought as it's easier to reason about knowing there won't be nil pointer dereferencing. Thoughts?

v2/pkg/options/options.go
```diff
@@ -47,9 +48,13 @@ type App struct {
...
-	BackgroundColour   *RGBA
-	Assets             fs.FS
-	AssetsHandler      http.Handler
+	BackgroundColour *RGBA
+	// Deprecated: Use AssetServer.Assets instead.
+	Assets fs.FS
+	// Deprecated: Use AssetServer.Handler instead.
+	AssetsHandler http.Handler
+	// AssetServer configures the Assets for the application
+	AssetServer        *assetserver.Options
```
leaanthony, line 57: If we made this a non-pointer, wouldn't the fields be testable for nil anyway? Just wondering if the zero value for this might be a better option.

leaanthony (approved):
A couple of small questions but looking good. I think we should get @mholt to sign off on it 😄

v2/pkg/options/options.go
stffabi: I think that would make it impossible to distinct some cases, e.g. if:

```go
Assets: embedFS,
AssetServer: assetserver.Options{}
```

This would mean `AssetServer.Assets` is nil. Does this now mean the developer wanted to have `nil` assets but just forgot removing it from the deprecated `Assets` or did they want the `Assets` and forgot to add them to the `AssetServer` struct.

For the check of inconsistency it's much easier to have struct-pointer. If it's nil one can be assured the developer didn't want to change anything and was using the deprecated options. If it has been set, the developer wanted to migrate to the `AssetServer` option, but forgot to remove the old assignments.

v2/internal/frontend/assetserver/common.go
stffabi: I think we might change the return value to a struct type instead instead of a struct pointer.

For the input I'm not sure if we can do that, see my reply to your comment on the options struct.

v2/pkg/options/options.go
stffabi: But yeah, if migration is over I would also opt for having a struct instead of a struct pointer there.

v2/internal/frontend/assetserver/common.go
stffabi: I have changed the return value to a struct.

mholt:
Aye, this is working well for me as far as I can tell!

Thank you thank you thank you for this, it allows me to use vanilla JS for my frontend which vastly simplifies things, and I can render "static" HTML pages dynamically using Go templates (which I also use to include HTML components onto my HTML pages).

I'll let you know if I run into any hiccups further on.

v2/pkg/options/assetserver/middleware.go
```diff
@@ -0,0 +1,20 @@
...
+	"net/http"
+)
+
+// Middleware defines a HTTP middleware that can be applied to the AssetServer.
+// The handler passed as next is the next handler in the chain. One can decide to call the next handler
+// or implement a specialized handling.
+type Middleware func(next http.Handler) http.Handler
+
+// ChainMiddleware allows chaining multiple middlewares to one middleware.
+func ChainMiddleware(middleware ...Middleware) Middleware {
```
mholt, line 13: FYI, I have not found myself using this, but I can't say it wouldn't be useful to others (or even myself in the future). Just wanted to let you know that I was able to do what I needed with a single middleware. Since I control all of the middleware handlers, I can just multiplex them from my single middleware function.

stffabi: Thanks for letting me know. I didn't want to have an array of middleware in the API, so I thought this would round it up with this for people that want to chain multiple middlewares.

stffabi:
Just updated the PR with docs and the template updates. @leaanthony would you mind taking a look into those commits? 🙏

stffabi:
@mholt did you see any problems with this PR, which would need a fix?
Otherwise I think we are good to go and could merge it.

mholt:
Nothing sticks out to me as problematic. So far I am using it successfully!

stffabi:
Awesome, you're welcome. So I'm going to merge this PR.

mholt:
This is fabulous, thank you. Has been working great.

PR #2062, opened 2022-11-08
Added quasar-js and quasar-ts templates

sgosiaco:
I added two new community templates based on [quasar](https://quasar.dev/)
Quasar has quite a few configuration options so these templates use relatively "safe/standard" defaults.
There is one template for JavaScript and one for TypeScript.


leaanthony:
Nice!

PR #2073, opened 2022-11-10
docs: sync documents

misitebao:
- Fix document formatting
- Sync Chinese and Japanese documents.


leaanthony:
Nice one @misitebao 🙏

PR #2086, opened 2022-11-13
Rename bleeding edge -> local development

leaanthony:
It's the local development setup guide so let's rename it.


misitebao (commented):
Good change💗.

PR #2091, opened 2022-11-14
[assetserver] Add support for range requests if the fs.FS provides an io.Seeker

stffabi:
This allows streaming audio and videos with range requests if the platform webview supports it.


leaanthony:
Awesome addition! I'm wondering if there's an easy way to demonstrate this that I could make and put in the examples directory.

leaanthony (approved):

stffabi:
Thanks for merging 🙏 

Yeah we could add an example, do we need to enable LFS on the repo to add the media files?

PR #2093, opened 2022-11-15
docs: Add README.ko.md

cybertramp:
Hi! 

I want to add korean version README.md documents. So I requested!

- Add korean version README
- Update 'README.ko.md' linked in README.md

I love the Go language and was happy to come across Wails while looking for an alternative to heavy Electron.

Thanks! Also, I will update the Korean version of the manual through __crowdin__ whenever the time is right.

Bye! 🎈

### DIFF
--- a/README.ja.md
+++ b/README.ja.md
@@ -42,7 +42,7 @@
 <strong>
 <samp>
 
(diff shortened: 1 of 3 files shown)


misitebao (approved):
LGTM.

Issue #2058, opened 2022-11-06
Add the ability to define a prefix/suffix to the generated class models


Goncalo-Marques:
### Is your feature request related to a problem? Please describe.

I am having some problems with my generated models.

Basically, I have a Go method that should return a slice of a structure named `Object`:
```go
type Object struct {}

type App struct {
	ctx context.Context
}

func (a *App) MyMethod() []Object {
	return nil
}
```

Then, after running `wails generate module`, I get the error (`wailsjs/go/models.ts`): `error TS2339: Property 'keys' does not exist on type 'typeof Object'`.

Generated code for context (focus on line 9):
```typescript
convertValues(a: any, classs: any, asMap: boolean = false): any {
	if (!a) {
		return a;
	}
	if (a.slice) {
		return (a as any[]).map(elem => this.convertValues(elem, classs));
	} else if ("object" === typeof a) {
		if (asMap) {
		     for (const key of Object.keys(a)) {
		         a[key] = new classs(a[key]);
		      }
		      return a;
		  }
		  return new classs(a);
	}
	return a;
}
```

This is happening because the JavaScript [Object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object) is being "confused" with my generated class also named `Object`.
It is possible that this could happen to every Go structure that as a similar name to the [built-in objects](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects) in JavaScript.

### Describe the solution you'd like

Instead of restricting the names we give to Go structures, we can solve this problem by:
- Adding two new flags to `wails generate module` 
   - `-prefix` (Prefix added to the generated class models)
   - `-suffix` (Suffix added to the generated class models)
- The zero values of `-prefix` and `-suffix` should work like the zero value of the string in Go
- The Pascal Case should be considered to keep consistency

#### Notes:
- We shouldn't forget to update [docs](https://wails.io/docs/reference/cli#module)
- The flags could be renamed to make them clearer

### Example

Generating wailsjs:
```bash
wails generate module -prefix myPrefixwithmultiplewords
```

"Pseudocode" to generate the class name:
```go
func generateClassName(prefix, suffix string) string {
   structNameGo := "Object"

   if len(prefix) != 0 {
      prefix = strings.Title(prefix)
   }
   if len(suffix) != 0 {
      suffix = strings.Title(suffix)
   }

   return prefix + structNameGo + suffix 
}
```

Result:
```javascript
[...]

PR #2101, opened 2022-11-16
Ability to define prefix / suffix for TS module

OlegGulevskyy:
As discussed in the issue #2058 , this PR adds ability for the app developer to define `prefix` and `suffix` for their return entities from the backend. Those pref/suff will be used in the generated `models.ts` file, to avoid clash with JS reserved keywords.

- Added a full list of JS keywords and detection of the clash
- Added warning message for each type to the console log when the types are being converted. A note here: - I am not exactly happy how this warning implemented (it's a simple log and outputting it into `stderr` but I don't know any other better way, I'd love some guidance here.
- Made sure prefix / suffix are added in all modes - `wails dev` / `wails build` / `wails generate module`. Please let me know if I missed any other.

Some implementation details question:
1. Prefix and Suffix flags are implemented as structs that can be reimported (they have `Flag` and `Description` properties). Is this ok thing to do in Go lang? I thought of treating it as enum, because retyping it in all other places is error prone and if the flag needs to be changed - it can be just changed in one place and affect all the other.
2. What is the better way to implement warning logging for cases when there is a clash with JS reserved keyword?

Please don't hesitate to say what's not idiomatic way to do things, what does not make sense or what does, I am a webdev with JS/TS background and Go is my new tool in the toolbox, so I am just learning and open to any feedback!

Thanks in advance


v2/internal/typescriptify/typescriptify.go
```diff
@@ -901,3 +909,13 @@ func differentNamespaces(namespace string, typeOf reflect.Type) bool {
 	}
 	return false
 }
+
+func typeClashWithReservedKeyword(input string) bool {
+	for _, v := range jsReservedKeywords {
```
leaanthony, line 914: might be worth doing a `strings.TrimSpace(input)` here?

v2/cmd/wails/internal/commands/dev/dev.go
```diff
@@ -33,36 +33,14 @@ import (
 
 	"github.com/fsnotify/fsnotify"
 	"github.com/leaanthony/clir"
+	commonFlags "github.com/wailsapp/wails/v2/cmd/wails/internal/flags"
 	"github.com/wailsapp/wails/v2/internal/fs"
 	"github.com/wailsapp/wails/v2/internal/process"
 	"github.com/wailsapp/wails/v2/pkg/clilogger"
 	"github.com/wailsapp/wails/v2/pkg/commands/build"
+	"github.com/wailsapp/wails/v2/cmd/wails/internal/utils"
```
leaanthony, line 41: In Go, you generally don't have ambiguous package names. In this instance you may want to call it "logutils" or something. It actually doesn't bother me so much, more an FYI 👍

leaanthony (approved):
Just a couple of comments 👍

v2/internal/typescriptify/typescriptify.go
OlegGulevskyy: agree, it won't hurt :)
will add it

v2/cmd/wails/internal/commands/dev/dev.go
OlegGulevskyy: great tip, thank you! will refactor it and keep it in mind!

stffabi:
Awesome work, thanks @OlegGulevskyy  👍 

Just a note which might be worth to be discussed:
I wonder if those settings should be put into `wails.json`. I think for most people that need to solve the ambiguity with JS keywords want this to be setup once for their project. Otherwise they would need to remember to define those options for every  `wails dev` and for every `wails build` call, because we now generate the bindings for every build. Also if multiple people are working on one project they would need to remember those settings. And one needs to make sure to update the build server definition as well, if a company is using build agents. If it was defined in `wails.json` this would only need to be defined once and every one using that project would use the correct settings.

Thoughts?

leaanthony:
That's a great call actually.

OlegGulevskyy:
Agreed, that's a good call!
Looking [at schema here](https://wails.io/schemas/config.v2.json), what would you say the properties could be?
My initial thinking is something like this
image

leaanthony:
Perhaps `ts_generation`?

OlegGulevskyy:
`ts_generation` works for me. What do you think? @stffabi

stffabi:
> `ts_generation` works for me. What do you think? @stffabi 

Yeah that sounds good for me👍

OlegGulevskyy:
Updated PR to use `wails.json` file to configure this one + updated `json` schema
Let me know if it looks good to you and I can update changelog

leaanthony:
LGTM. One thing I'm going to suggest is a test in the `binding_test` directory to ensure it works as intended and we don't get regressions in the future 👍

misitebao:
Does this feature have a "flag" that syncs to the "wails generate module" command?

OlegGulevskyy:
Had busy few days with work, but got your comment, I haven't forgotten and will look into it! :)

OlegGulevskyy:
Added a simple use cases for automated testing, to make sure the entity is generated with prefix and suffix whenever one or the other is supplied.
Let me know if you'd like me to cover some other cases :)

Thanks for your feedback and guidance, I appreciate it!

cc @leaanthony @stffabi

website/static/schemas/config.v2.json
```diff
@@ -226,5 +226,25 @@
...
+				"type": "object",
+				"description": "Bindings configurations",
+				"properties": {
+						"ts_generation": { 
+								"type": "object",
+								"description": "model.ts file generation config",
+								"properties": {
+										"prefix": {
+												"type": "string",
+												"description": "All generated JavaScript entities will be prefixed"
```
leaanthony, line 240:
```suggestion
												"description": "All generated JavaScript entities will be prefixed with this value"
```

website/static/schemas/config.v2.json
```diff
@@ -226,5 +226,25 @@
...
+										},
+										"suffix": {
+												"type": "string",
+												"description": "All generated JavaScript entities will be suffixed"
```
leaanthony, line 244:
```suggestion
												"description": "All generated JavaScript entities will be suffixed with this value"
```

v2/internal/typescriptify/typescriptify.go
```diff
@@ -901,3 +907,20 @@ func differentNamespaces(namespace string, typeOf reflect.Type) bool {
...
+		if in == v {
+			return true
+		}
+	}
+
+	return false
+}
+
+func warnAboutTypesClash(entity string) {
+	l := log.New(os.Stderr, "", 0)
```
leaanthony, line 923:
```suggestion
	// TODO: Refactor logging
	l := log.New(os.Stderr, "", 0)
```
Adding this as a reminder for future self that we need to refactor logging, support better warnings, honour silent flag

leaanthony (approved):
Just a few small suggestions 👍

leaanthony:
Also, please update the CHANGELOG in the website so it isn't missed next release 👍🙏

OlegGulevskyy:
Sure, all done! Let me know if the wording in changelog needs tweaking

leaanthony:
Awesome first issue PR! :D

Issue #2100, opened 2022-11-16
wails dev -race -v 2 create (very deep) nested dir in build/bin


bokunodev:
### Description

wails dev -race -v 2 create (very deep) nested dir in build/bin

### To Reproduce

1. create wails project with `svelte+ts` template
2. run `wails dev -race -v 2`
3. change some go code to trigger wails to recompile the project
4. repeat step 3 few times

### Expected behaviour

only create one binary file in `build/bin`

### Screenshots

![image](https://user-images.githubusercontent.com/54968427/202053566-4182ffdd-638d-4e38-bec8-0e30cab78754.png)

### Attempted Fixes

i believe it have something todo with this line
https://github.com/wailsapp/wails/blob/18b2d315dda8f4abe6681b399479315f5092fc33/v2/pkg/commands/build/base.go#L276

### System Details

```shell
wails doctor output


System
------
OS:		Artix Linux
Version: 	Unknown
ID:		artix
Go Version:	go1.19.3
Platform:	linux
Architecture:	amd64

Wails
------
Version: 		v2.2.0
Package Manager: 	pacman

Dependency	Package Name	Status		Version
----------	------------	------		-------
*docker 	docker 		Available 	1:20.10.21-1
gcc 		gcc 		Installed 	12.2.0-1
libgtk-3 	gtk3 		Installed 	1:3.24.34+r156+g812b3930d0-1
libwebkit 	webkit2gtk 	Installed 	2.38.2-1
npm 		npm 		Installed 	8.19.3
pkg-config 	pkgconf 	Installed 	1.8.0-1
```

OlegGulevskyy:
I had the same issue once but didn't have time to repro it, nice catch!

leaanthony:
I wonder if this is a regression from the recent modularisation changes.

OlegGulevskyy:
I was checking this right now, and the results are weird
@bokunodev  is right, `compiledBinary` has this incorrect path when saving couple of times on `dev` mode.
But what's weird to me is that I don't see why it duplicates itself like this.

Here is me reproducing it and keeping track of those lines below
```go
	outputFile := b.OutputFilename(options)
	options.Logger.Println("APPDIR", appDir)
	options.Logger.Println("OUTPUTFILE", outputFile)

	compiledBinary := filepath.Join(appDir, outputFile)
	options.Logger.Println("COMPILED BINARY", compiledBinary)

```
initial `wails dev` logs
```
[...]

PR #2103, opened 2022-11-16
Fix multiple nested `build/bin` folders in `dev` mode

OlegGulevskyy:
Following discussion in #2100 - this PR removes assignment to path where executable is stored (`/build/bin`). This was adding `/build/bin` to the executable name on each save in `dev` mode, which results in `/build/bin/build/bin...` duplicated nested path.

Have tested on MacOS only, can test on Windows 11 a bit later to make sure no regression


leaanthony (approved):

stffabi:
LGTM, would it be possible for you to also add your changes to the changelog here `website/src/pages/changelog.mdx`? That would be awesome 🙏

OlegGulevskyy:
Of course, done!
Will keep in mind `changelogs` from now on :)

leaanthony:
Yeah, that's how we're tracking changes because GitHub's commit message bundle isn't good enough 😀

leaanthony:
I tested on Windows 11:
* Generate new project
* `wails dev`, then close
* `wails dev`, then close
* Checked FS, all good
* `wails build` - all good 

Thanks heaps @OlegGulevskyy 🙏

Issue #2057, opened 2022-11-06
`wails dev` downgrades wails in go.mod


mholt:
### Description

`wails dev` is downgrading wails when running `go mod tidy` (or `go get`? but it's during the output that says running `go mod tidy`), causing compilation to fail.

I think this happens if go.mod is using a newer version than the latest tag (`@latest`), even when the CLI is also built using the same commit that is newer than the latest tag.

### To Reproduce

I am using the latest commit on master because of some recent changes that were merged in that my program relies on.

I cloned wails locally, changed to `v2/cmd/wails` and ran `go install`. I confirmed that `wails doctor` says that the latest commit was built:

```
Wails
------
Version:                v2.1.0
Revision:               928bc75602bb1302ed211d5b08b902c58236192a
Modified:               false
...
```

and then I updated my program's go.mod file to use the same  commit (`go get github.com/wailsapp/wails/v2@928...` (or `@master` at the time).

But curiously, I noticed that my go.mod file was being modified to revert the version of wails to v2.1.0. This causes a compile error because my code requires the latest changes.

I suspect this function (or something around here) is culprit:

https://github.com/wailsapp/wails/blob/55f76a291bc5ee5c92bdedbc6c05c4f4400b85b3/v2/internal/gomod/gomod.go#L86

### Expected behaviour

`wails dev` can run `go mod tidy` and `go get`, but it shouldn't change go.mod if the CLI and code base match, even if they're not on a tagged release.

### Attempted Fixes

I can work around this temporarily by adding a `replace` to my go.mod that points to my locally-cloned copy of the repo.

### System Details

```shell
(see above)
```

### Additional context

Thanks again for Wails!!

OlegGulevskyy:
Thanks for detailed explanation, I am gonna try to repro on my end and see if I can fix it :)
If I manage to fix it, I will send a PR through

OlegGulevskyy:
Sorry for the delay guys, got busy with some stuff
I had some time to check this today, and I am surprised to say that suggested versions by @leaanthony  is actually not being picked up by this.
I tested with this version here:
https://github.com/wailsapp/wails/commit/3474d4e1e1ecc9cc3051a2fd8cb7c8836276d07f

Whenever commit is set - it is not overwritten
Whenever `X.X.X-beta.` is set, it is overwritten to the latest - `2.2.0`
[...]

PR #2107, opened 2022-11-18
Do not overwrite specific wails versions

OlegGulevskyy:
This PR fixes the issue raised in #2057 , where comparing versions of `wails` package installed via a specific commit `go get github.com/wailsapp/wails/v2@928...` was detected as **out of date** version and was overwritten all the time.

Adding very simple logic to check if a version contains something else apart from the correct semver, it means the developer knows what they are doing, so we should let things as they are.

Not sure if this is the best way to detect this though, I quickly looked over `semver` package to see if they have a method detecting stuff apart from the actual version, but didn't find any.

Happy to get better ideas from you awesome people!


mholt:
Oh nice. I'll leave it up to the maintainers to decide the best approach (of course) but from my own perspective I think that's logical: if I'm not using a tagged release, it's probably intentional.

Either that, or the logic needs to be smarter to "go get" the specific version the binary is running. But then it changes my go.mod which I might not want.

So, personally, I think this approach is good. :man_shrugging: 

Thanks! :D

leaanthony:
Thanks @OlegGulevskyy ! Historically we've used `vX.X.X-beta.YY` which may get picked up by this. We're some way off v3 beta so I'm not that concerned about it right now but might be worth thinking about.

OlegGulevskyy:
Sure, that's a good point! If that's a potential in future again, then we should probably figure out a smarter handling for this.
Maybe like first check for words beta, alpha etc ? Feels a bit Jacky though. I wonder if semver does not offer this kind of checks for beta alpha versions etc.

Thanks for feedback guys!

leaanthony:
I was sure semver library allowed you to extract the metadata part of the version

OlegGulevskyy:
Got it, I am not familiar with it, so I will follow up on this

Thanks!

mholt:
You probably want https://pkg.go.dev/github.com/Masterminds/semver/v3#Version.Prerelease and possibly https://pkg.go.dev/github.com/Masterminds/semver/v3#Version.Metadata

leaanthony:
@OlegGulevskyy - shall we convert this to draft?

OlegGulevskyy:
My bad, I completely forgot about this PR!
Thanks for bumping it up. I will look into it by the end of the week max, so with that in mind if you feel like it should be in Draft stage then let's convert it

OlegGulevskyy:
My bad, I completely forgot about this PR!
Thanks for bumping it up. I will look into it by the end of the week max, so with that in mind if you feel like it should be in Draft stage then let's convert it

leaanthony:
Yeah if @mholt is happy with this behaviour I think it's good to go. Good job!

mholt:
I will test it today :)

mholt:
Hmm, when I try this patch, I still get the error from when I tried other non-released versions after 2.2:

```
ERROR: 
flag provided but not defined: -tsprefix
Usage of /tmp/wailsbindings:
  -config string
        path to the configuration file (default "/home/matt/.config/timelinize/timelinize.json")
  -gui
        run GUI (default true)
  -server
        run HTTP server
```

I think it's unrelated to the go.mod stuff, but it means I can't test this patch until I figure out why this is happening.

I don't know enough about Wails' internals yet to know when/why this is happening, but it looks like Wails is compiling a binary called `/tmp/wailsbindings` that happens to be my program's `main.go`; because my `main.go` is what defines these flags. Apparently Wails is trying to run my binary but is passing `-tsprefix`, which I believe is a new flag on the Wails binary.

So it seems like Wails is trying to run a binary that should know what `-tsprefix` is, but it's running mine instead.

Is that intended? :thinking:

OlegGulevskyy:
Thanks for testing it @mholt 
What was the issue with -tsprefix in the end?

mholt:
@OlegGulevskyy It's unrelated to this, but basically, if your app calls `flag.Parse()` then it breaks the latest yet-unreleased versions of Wails, which adds flags to your program. I think @stffabi said he'd make a PR that changes it to env vars instead of messing with flags, which sounds much better!

mholt:
Anyway, to clarify, this patch LGTM and could be taken out of "draft mode" IMO.

mholt:
Cool :) Thank you!

PR #2123, opened 2022-11-22
Refactored build command

leaanthony:
This PR refactors the `build` command. Why do this? A number of the commands share similar flags and we have a lot of redundant code. We also have inconsistent style. By taking advantage of the latest [clir](https://github.com/leaanthony/clir) release, we can tidy a lot of this code up. If the approach is agreed on, I'll move to the others.


v2/cmd/wails/build.go
```diff
@@ -0,0 +1,248 @@
...
+		if buildOptions.Platform == "windows" {
+			desiredFilename += ".exe"
+		}
+		buildOptions.OutputFile = desiredFilename
+
+		if f.OutputFilename != "" {
+			buildOptions.OutputFile = f.OutputFilename
+		}
+
+		if f.Upx && f.SkipBindings {
```
stffabi, line 198: Just a minor fix.
```suggestion
		if f.Obfuscated && f.SkipBindings {
```

stffabi (approved):

PR #2128, opened 2022-11-23
Supplement the missing WindowSetAlwaysOnTop in runtime.d.ts and runti…

gotid:
WindowSetAlwaysOnTop is not exported in runtime.d.ts and runtime.js, which makes this function unavailable.

The function has been completed in all runtime.d.ts and runtime.js.

After testing, it is ready for use.


gotid (commented):
After testing, it is ready for use.

leaanthony (approved):

Issue #2033, opened 2022-11-01
Models to Typescript with omit empty and with special characters adds a question mark to much


Implocell:
### Description

If you have a struct in Go with omitempty tag and the name contains special characters it will translate to TS with a questionmark inside the string.

### To Reproduce

compile the following struct
```go
type MovingAverage struct {
	Two00DayMovingAverage      string `json:"200DayMovingAverage,omitempty"`
}
```

In the models.ts file it gets converted to:
 "200DayMovingAverage?"?: string;

Seems like it just parses the omitempty and puts it inside the string then outside.

### Expected behaviour

expects it to handle it in the same way as without special characters

### System Details

```shell
System
------
OS:		Ubuntu
Version: 	20.04
ID:		ubuntu
Go Version:	go1.19.2
Platform:	linux
Architecture:	amd64

Wails
------
Version: 		v2.1.0
Package Manager: 	apt

Dependency	Package Name		Status		Version
----------	------------		------		-------
*docker 	docker.io 		Installed 	20.10.20
gcc 		build-essential 	Installed 	12.8ubuntu1.1
libgtk-3 	libgtk-3-dev 		Installed 	3.24.20-0ubuntu1.1
libwebkit 	libwebkit2gtk-4.0-dev 	Installed 	2.36.8-0ubuntu0.20.04.1
npm 		npm 			Installed 	6.14.4+ds-1ubuntu2
*nsis 		nsis 			Available 	3.05-2
pkg-config 	pkg-config 		Installed 	0.29.1
```

### Additional context

If someone can point me to the conversion between the go structs and models.ts I can add a fix for this

leaanthony:
Hi there! 👋 Thanks for raising this. A good place to start is to add a test [here](https://github.com/wailsapp/wails/tree/master/v2/internal/binding/binding_test) that shows the bug. Submit that as a PR and we should be able to see it fail in the pipelines. Then update the PR afterwards with a fix and we should be good to go 👍

PR #2130, opened 2022-11-24
BugFix: 2033

Implocell:
# Description
If you have a struct in Go with omitempty tag and the name starts with a special characters it will translate to TS with a questionmark inside the string.

Creating a pull request with a test inside as mentioned in the issue I made #2033


tmclane:
Is this dead?

Implocell:
To be totally honest I have totally forgotten this. Not sure what I did to fix this on my end, or if it even is an issue today? I can try to have a look at it and see if it still is an issue, if it is not, I can close the issue and remove this.

Implocell:
Seems the tests are passing so I will close this, sorry should have checked this earlier

PR #2151, opened 2022-11-30
[linux] Add support for WebKit2GTK 2.36+ features

### LINKED ISSUE #1498: [linux] Add support for WebKitGTK 2.36+

WebKitGTK 2.36+ brings support for several new features when processing requests from the frontend, like response status codes, headers, http methods.

This would be especially interesting to bring the `AssetsHandler` at almost the same feature level like the other platforms and allows it to be used for a new plugin concept.

Add support for WebKitGTK 2.36+ and implement available features on the `AssetsHandler`.

_No response_

### Additional context

It should be discussed in which form we want to add the new minimal dependency.
- Just require WebKITGTK 2.36+ for any build.
- Make it configurable to also allow building with an older version at the cost of loosing the new features.

2.36.0 has been released on March 21st, 2022 so it might be that not all distributions have already updated to it. On the other hand it seems like SecurityUpdates are only provided for the last major minor release. But still distributions might track their own branch with backports.

stffabi:
Fixes #1498

### DIFF
--- a/v2/internal/frontend/desktop/linux/webkit2_36.go
+++ b/v2/internal/frontend/desktop/linux/webkit2_36.go
@@ -1,0 +1,73 @@
+//go:build linux && webkit2_36
+
+package linux
+
+/*
+#cgo linux pkg-config: gtk+-3.0 webkit2gtk-4.0 libsoup-2.4
+
+#include "gtk/gtk.h"
+#include "webkit2/webkit2.h"
+#include "libsoup/soup.h"
+*/
+import "C"
+
+import (
+	"net/http"
+	"strings"
+	"unsafe"
+
+	"github.com/wailsapp/wails/v2/internal/frontend/assetserver"
+)
+
+const webkit2MinMinorVersion = 36
+
+func webkit_uri_scheme_request_get_http_method(req *C.WebKitURISchemeRequest) string {
+	method := C.GoString(C.webkit_uri_scheme_request_get_http_method(req))
+	return strings.ToUpper(method)
+}
+
+func webkit_uri_scheme_request_get_http_headers(req *C.WebKitURISchemeRequest) http.Header {
+	hdrs := C.webkit_uri_scheme_request_get_http_headers(req)
+
+	var iter C.SoupMessageHeadersIter
+	C.soup_message_headers_iter_init(&iter, hdrs)
+
+	var name *C.char
+	var value *C.char
+
+	h := http.Header{}
+	for C.soup_message_headers_iter_next(&iter, &name, &value) != 0 {
+		h.Add(C.GoString(name), C.GoString(value))
+	}
+
+	return h
+}
+
+func webkit_uri_scheme_request_finish(req *C.WebKitURISchemeRequest, code int, header http.Header, stream *C.GInputStream, streamLength int64) error {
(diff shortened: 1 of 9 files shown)


website/docs/reference/options.mdx
```diff
@@ -249,24 +249,24 @@ dynamically with an `http.Handler` or hook into the request chain with an `asset
...
-| Response Body           | ✅  | ✅  | ✅  |
-| Response Body Streaming | ❌  | ❌  | ✅  |
-| WebSockets              | ❌  | ❌  | ❌  |
-
-NOTE: Linux is currently very limited due to targeting a WebKit2GTK Version < 2.36.0. In the future some features will be
-supported by the introduction of WebKit2GTK 2.36.0+ support.
+| Feature                 | Win | Mac | Lin    |
+| ----------------------- | --- | --- | ------ |
+| GET                     | ✅  | ✅  | ✅      |
+| POST                    | ✅  | ✅  | ✅ [^1] |
```
stffabi, line 255: @leaanthony , @misitebao does anybody of you know if that footnotes do work with our docs?

v2/internal/frontend/desktop/linux/webkit2_36.go
```diff
@@ -0,0 +1,73 @@
+//go:build linux && webkit2_36
+
+package linux
+
+/*
+#cgo linux pkg-config: gtk+-3.0 webkit2gtk-4.0 libsoup-2.4
```
leaanthony, line 6: Is `libsoup` a new dependency that we need to add to our website and package manager code?

leaanthony:
Confirmed working on `Ubuntu 22.04`. Nice job!

v2/internal/frontend/desktop/linux/webkit2_36.go
stffabi: WebKit2GTK uses `libsoup` internally, so I think it should not be possible to have WebKit2GTK installed without this dependency. But maybe it's better to be on the safe side and add it.

I'll add that as soon as we have merged the CLI refactoring so it doesn't interfere with that. Are you fine with that?

website/docs/reference/options.mdx
stffabi: Ah yeah, always forget about that 😄

v2/internal/frontend/desktop/linux/webkit2_36.go
leaanthony: Yeah sure, though I suspect you are right: it's probably installed by webkit2gtk anyway so no need to specify it 👍

leaanthony:
#2152 merged, you should be good to go with this.

stffabi:
Awesome, thanks for approving 🙏

website/docs/reference/options.mdx
leaanthony: Oh my bad! I wondered why it looked like nothing had happened. The superscript 1 is super hard to click and it jumps to the bottom of the page. How about a `*` next to is with some text right under it or maybe another column with `Linux (webkit 2.36+)` in it?

stffabi: Sorry already merged 😔, but will open another PR to improve that.

PR #2154, opened 2022-11-30
chore: update-workflows

misitebao:
- [x] Merge push, test, PR workflows
- [x] Simplify sponsors image generation workflow
- [x] Add auto sync translation document workflow


.github/workflows/build-and-test.yml
```diff
@@ -23,24 +23,64 @@ jobs:
         if: matrix.os == 'ubuntu-latest'
         run: sudo apt-get update -y && sudo apt-get install -y libgtk-3-dev libwebkit2gtk-4.0-dev build-essential pkg-config
 
-      - name: Set up Go 1.18
+      - name: Setup Go
```
leaanthony, line 26: When I looked at this, the tests did not seem to go into the `v2` directory. I tried using `working-directory: ./v2/` but that generated some build constraint errors that I haven't investigated yet.

.github/workflows/pr.yml
```diff
@@ -1,30 +1,51 @@
 name: PR Checks
+
 on:
   pull_request:
   pull_request_review:
-    types: [ submitted ]
+    types: [submitted]
 
 jobs:
   check_branch_name:
```
leaanthony, line 9: I think I want to get rid of this check. It's artificial and generates noise. I guess we don't care about source branch name in the end.

.github/workflows/sync-translated-documents.yml
```diff
@@ -0,0 +1,31 @@
...
+    if: github.repository == 'wailsapp/wails'
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Set node
+        uses: actions/setup-node@v2
+        with:
+          node-version: 16.x
+
+      - name: Update sponsors
```
leaanthony, line 18: Is this the right now?

leaanthony (commented):
Just some comments... 👍

.github/workflows/pr.yml
misitebao: Meaning there is no need to check the branch name of the PR, right?

.github/workflows/sync-translated-documents.yml
misitebao: Oh, this is my mistake. 😭

.github/workflows/build-and-test.yml
misitebao: `/wails/v2/cmd/wails/internal/commands/initialise/templates/templates/vue/main.tmpl.go` is a template file, but `//go:embed` in the file will cause an error in the `go test -v ./...` command: 
```
$ go test -v ./...
cmd\wails\internal\commands\initialise\templates\templates\vue\main.tmpl.go:11:12: pattern all:frontend/dist: no matching files found
```

leaanthony: We can call it `main.go.tmpl` and it will still work 👍

leaanthony (approved):

.github/workflows/build-and-test.yml
misitebao: Will be fixed in this PR https://github.com/wailsapp/wails/pull/2169

PR #2169, opened 2022-12-01
fix: fix `go test` errors

misitebao:
- [x] Fix errors in tests
- [x] Fix failed test cases

Most errors are caused by missing `//go:build windows` build tags.

Fatal errors during testing have been fixed, but there are still some use cases that don't pass the test.


v2/internal/project/project.go
```diff
@@ -38,7 +38,7 @@ type Project struct {
 	/*** Internal Data ***/
 
 	// The path to the project directory
-	Path string
+	Path string `json:"projectdir"`
```
misitebao, line 41: The modification here is so that the test case can run normally, but I don't know if it will affect other code operations.

leaanthony (approved):

leaanthony:
I've pushed a couple of changes to see if it would fix the linker errors and it does appear to be fixed. Thoughts?

misitebao:
> I've pushed a couple of changes to see if it would fix the linker errors and it does appear to be fixed. Thoughts?

I think it's ok to merge.

leaanthony:
Good job!!

PR #2173, opened 2022-12-02
Add wails-naive-template

tk103331:


misitebao:
The recommended naming format is `wails-template-xxx`.

PR #2176, opened 2022-12-02
examples/customlayout: feature: add event listener for enter key

Holmqvist1990:
## Description
Commonly expected UX on inputs is to handle submit on enter key.

## Changes
Attached an event listener on the document to call `window.greet()` on `keydown` for enter key.


leaanthony:
Nice one thanks!

PR #2181, opened 2022-12-05
Windows tray menus

leaanthony:
A leaner PR based on #1959


stffabi (commented):
LGTM, apart from the open things regarding the message loops discussed in #1959

leaanthony:
Thanks for reviewing. I imagine this will evolve over time...

PR #2182, opened 2022-12-05
go vet cleanup

tmclane:
Updated to remove only one line of `unreachable` code.

### DIFF
--- a/v2/pkg/menu/styledlabel.go
+++ b/v2/pkg/menu/styledlabel.go
@@ -251,6 +251,4 @@ func ParseANSI(input string) ([]*StyledText, error) {
 			}
 		}
 	}
-
-	return result, nil
 }


tmclane:
Ah I forgot we weren't supporting older Go versions any longer.
The `goimports` adds it automatically I think for backwards-compatibility.
I can take that out.

leaanthony (commented):
That reduced it down to 1 line! Do you know why the return was removed?

tmclane:
Because of the `default` case in the `switch` statement the final line in the function will never fire.

tmclane:
I am thinking that perhaps leaving the final return and updating the `default` case to not do a `return` might be more appropriate.

leaanthony:
It's hard to review on mobile 😅

leaanthony:
I think it's ok as is 👍

leaanthony (approved):

Issue #2038, opened 2022-11-02
Crash during typescript generation if struct contains anonymous sub-struct


avengerweb:
### Description

Crash during typescript generation if struct contains anonymous sub-struct

```
Build error - 
panic: runtime error: index out of range [1] with length 1

goroutine 1 [running]:
github.com/wailsapp/wails/v2/internal/binding.(*Bindings).AddStructToGenerateTS(0x14000210000, {0x1400021e198, 0x8}, {0x10103cdbc, 0x4}, {0x1010a8260?, 0x1400020e0c0})
        /Users/avengerweb/go/pkg/mod/github.com/wailsapp/wails/v2@v2.1.0/internal/binding/binding.go:202 +0x558
github.com/wailsapp/wails/v2/internal/binding.(*Bindings).getMethods(0x14000210000, {0x1010c80a0?, 0x1400000f3c8})
        /Users/avengerweb/go/pkg/mod/github.com/wailsapp/wails/v2@v2.1.0/internal/binding/reflect.go:138 +0xc4c
github.com/wailsapp/wails/v2/internal/binding.(*Bindings).Add(0x14000210000, {0x1010c80a0?, 0x1400000f3c8?})
        /Users/avengerweb/go/pkg/mod/github.com/wailsapp/wails/v2@v2.1.0/internal/binding/binding.go:62 +0x2c
github.com/wailsapp/wails/v2/internal/binding.NewBindings(0x1400000f3f8, {0x14000021550, 0x1, 0x100effee0?}, {0x14000111e58, 0x4, 0x1400011e480?}, 0x0)
        /Users/avengerweb/go/pkg/mod/github.com/wailsapp/wails/v2@v2.1.0/internal/binding/binding.go:50 +0x1ec
github.com/wailsapp/wails/v2/internal/app.(*App).Run(0x14000021560?)
        /Users/avengerweb/go/pkg/mod/github.com/wailsapp/wails/v2@v2.1.0/internal/app/app_bindings.go:28 +0xa4
github.com/wailsapp/wails/v2/pkg/application.(*Application).Run(0x1400000f3e0)
        /Users/avengerweb/go/pkg/mod/github.com/wailsapp/wails/v2@v2.1.0/pkg/application/application.go:69 +0x168
github.com/wailsapp/wails/v2.Run(0x140000010e0?)
        /Users/avengerweb/go/pkg/mod/github.com/wailsapp/wails/v2@v2.1.0/wails.go:14 +0xb0

```

[...]

PR #2186, opened 2022-12-09
Fix issue with anonymous structures #2038

avengerweb:
Add full support for anonymous structures

Empty anonymous (no fields defined) structures should return instance of Object (empty - {}) instead of try creating `new null()` that cause JS error. Such behavior make any structure with empty structure as a field unusable.


leaanthony (approved):
Sorry for being slow on this!

leaanthony (approved):

leaanthony (approved):

PR #2187, opened 2022-12-09
Add dummy systray_mac to fix unit tests

avengerweb:
Unit tests are broken after https://github.com/wailsapp/wails/commit/b84a2e52553463fc743b4e4b4e587734eed1ee52 because it has build restriction only for windows.

This PR is fixing tests for MacOS by adding Systray structure that will produce NotImplemented error if someone try to use it on Mac; however it still be buildable and testable.

### DIFF
--- a/v2/internal/platform/systray.go
+++ b/v2/internal/platform/systray.go
@@ -1,5 +1,3 @@
-//go:build windows
-
 package platform
 
 import (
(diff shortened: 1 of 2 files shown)


stffabi:
Awesome, thanks so much for taking your time to fix this 🙏.

LGTM, just a little comment: 
I suspect the same iasue applies to Linux. Would it be possible for you to change this to include the dummy systray for all platforms other than Windows?

leaanthony:
I'll handle the Linux fix in a different PR 👍

stffabi (approved):

stffabi:
Thanks again for taking your time fixing this 🙏

PR #2190, opened 2022-12-09
[Linux] Fix right mouse button event

leaanthony:
Allows the right mouse button event through whilst retaining the existing functionality of showing no context menu.
@mholt - please let me know if this fixes your issue 👍 

Also fixes the systray experiment compile issues on linux


leaanthony:
Any thoughts on this @mholt?

PR #2191, opened 2022-12-10
Update libwebkit package name for Fedora

NullCode1337:
Without this change, `wails doctor` reports libwebkit not being installed (even though it already is)


leaanthony (commented):
Keeping both lines in there should work and have the benefit of working on more versions. Could you please try that?

NullCode1337:
> Keeping both lines in there should work and have the benefit of working on more versions. Could you please try that?

sure thing

PR #2194, opened 2022-12-10
[bindings] Pass ts_generation flags as env variable

stffabi:
Otherwise this will clash if an app also uses flags internally


leaanthony (approved):
Thanks for tidying this up @stffabi 🙏

PR #2196, opened 2022-12-12
[dev] Remove redundant steps that are also executed during build.Build

stffabi:


OlegGulevskyy:
I was wondering why we do that, but always forgot to ask.. :D

stffabi:
Thanks for approving 🙏

PR #2203, opened 2022-12-13
[assetserver] Inject runtime/IPC into all index html files

### LINKED ISSUE #2047: Runtime and IPC only injected on index.html file
### Description

My frontend is vanilla JS, so I have multiple HTML files. It's a good-ol', boring website from the early 2000s. Relatively little client-side JS (no routing, etc).

I noticed that `runtime.js` and `ipc.js` are only injected for `index.html`:

https://github.com/wailsapp/wails/blob/638caf72f0a85c9b29dc4d4e993d2f80e01491b7/v2/internal/frontend/assetserver/assetserver.go#L168-L187

This came after some head-scratching as to why `Uncaught TypeError: window.go is undefined` errors kept appearing but it worked fine on the index page.

Additionally, hot reloading doesn't seem to work on any non-index file, but maybe that's a separate issue.

### To Reproduce

1. Make a file that is not index.html
2. Navigate to that file in the front-end
3. The console shows errors because the runtime and IPC are not available

### Expected behaviour

I think it would make more sense to either inject these on all HTML pages, or to not inject them at all.

### Screenshots

_No response_

### Attempted Fixes

Fortunately, I can work around this by manually adding them:

```
<script src="/wails/ipc.js"></script>
<script src="/wails/runtime.js"></script>
```

**Proposed solutions:**

1) Inject those on all HTML pages
2) Don't inject them on any pages (make the user do it in their HTML)
3) Since the user already has to import App.js to access bound methods, maybe just include/import those files as part of that, so it can happen implicitly?

### System Details

```shell
Wails CLI v2.1.0

System
------
OS:		Pop!_OS
Version: 	22.04
ID:		pop
Go Version:	go1.19.1
Platform:	linux
Architecture:	amd64

Wails
------
Version: 		v2.1.0
Revision:		e545a7747f97ffc1f130b61d0b2e3f5de3b23279
Modified:		false
Package Manager: 	apt

Dependency	Package Name		Status		Version
----------	------------		------		-------
*docker 	docker.io 		Available 	20.10.12-0ubuntu4
gcc 		build-essential 	Installed 	11.3.0
libgtk-3 	libgtk-3-dev 		Installed 	3.24.33-1ubuntu2
libwebkit 	libwebkit2gtk-4.0-dev 	Installed 	2.36.8-0ubuntu0.22.04.1
npm 		npm 			Installed 	8.5.1~ds-1
*nsis 		nsis 			Available 	3.08-2
pkg-config 	pkg-config 		Installed 	0.29.2

* - Optional Dependency

Diagnosis
---------
  - docker: sudo apt install docker.io
  - nsis: sudo apt install nsis



If Wails is useful to you or your company, please consider sponsoring the project:


```
```


### Additional context

Thank you for Wails. Its value is ramping up!!

### PR DESCRIPTION
**Pull request by stffabi**
It will also be injected into all html files returned when requesting a folder path.

Fixes #2047

### DIFF
--- a/v2/pkg/assetserver/assetserver.go
+++ b/v2/pkg/assetserver/assetserver.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"math/rand"
 	"net/http"
-	"net/http/httptest"
 	"strings"
 
 	"golang.org/x/net/html"
@@ -111,23 +110,60 @@ func (d *AssetServer) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 		return
 	}
 
-	header := rw.Header()
 	if d.servingFromDisk {
-		header.Add(HeaderCacheControl, "no-cache")
+		rw.Header().Add(HeaderCacheControl, "no-cache")
+	}
+
+	handler := d.handler
+	if req.Method != http.MethodGet {
+		handler.ServeHTTP(rw, req)
+		return
 	}
 
 	path := req.URL.Path
-	switch path {
-	case "", "/", "/index.html":
-		recorder := httptest.NewRecorder()
-		d.handler.ServeHTTP(recorder, req)
-		for k, v := range recorder.HeaderMap {
-			header[k] = v
+	if path == runtimeJSPath {
+		d.writeBlob(rw, path, d.runtimeJS)
+
+	} else if path == runtimePath && d.runtimeHandler != nil {
+		d.runtimeHandler.HandleRuntimeCall(rw, req)
+
+	} else if path == ipcJSPath {
+		content := d.runtime.DesktopIPC()
+		if d.ipcJS != nil {
+			content = d.ipcJS(req)
+		}
+		d.writeBlob(rw, path, content)
+
+	} else if script, ok := d.pluginScripts[path]; ok {
+		d.writeBlob(rw, path, []byte(script))
+
+	} else if d.isRuntimeInjectionMatch(path) {
+		recorder := &bodyRecorder{
+			ResponseWriter: rw,
+			doRecord: func(code int, h http.Header) bool {
+				if code == http.StatusNotFound {
+					return true
+				}
+
+				if code != http.StatusOK {
+					return false
+				}
+
+				return strings.Contains(h.Get(HeaderContentType), "text/html")
+			}}
+
+		handler.ServeHTTP(recorder, req)
+
+		body := recorder.Body()
+		if body == nil {
+			// The body has been streamed and not recorded, we are finished
+			return
 		}
 
-		switch recorder.Code {
+		code := recorder.Code()
+		switch code {
 		case http.StatusOK:
-			content, err := d.processIndexHTML(recorder.Body.Bytes())
+			content, err := d.processIndexHTML(body.Bytes())
 			if err != nil {
 				d.serveError(rw, err, "Unable to processIndexHTML")
 				return
@@ -138,34 +174,12 @@ func (d *AssetServer) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 			d.writeBlob(rw, indexHTML, defaultHTML)
 
 		default:
-			rw.WriteHeader(recorder.Code)
+			rw.WriteHeader(code)
 
 		}
 
-	case runtimeJSPath:
-		d.writeBlob(rw, path, d.runtimeJS)
-
-	case runtimePath:
-		if d.runtimeHandler != nil {
-			d.runtimeHandler.HandleRuntimeCall(rw, req)
-		} else {
-			d.handler.ServeHTTP(rw, req)
-		}
-
-	case ipcJSPath:
-		content := d.runtime.DesktopIPC()
-		if d.ipcJS != nil {
-			content = d.ipcJS(req)
-		}
-		d.writeBlob(rw, path, content)
-
-	default:
-		// Check if this is a plugin script
-		if script, ok := d.pluginScripts[path]; ok {
-			d.writeBlob(rw, path, []byte(script))
-			return
-		}
-		d.handler.ServeHTTP(rw, req)
+	} else {
+		handler.ServeHTTP(rw, req)
 	}
 }
 
@@ -228,4 +242,13 @@ func (d *AssetServer) logError(message string, args ...interface{}) {
 	if d.logger != nil {
 		d.logger.Error("[AssetServer] "+message, args...)
 	}
+}
+
+func (AssetServer) isRuntimeInjectionMatch(path string) bool {
+	if path == "" {
+		path = "/"
+	}
+
+	return strings.HasSuffix(path, "/") ||
+		strings.HasSuffix(path, "/"+indexHTML)
 }
--- a/v2/pkg/assetserver/body_recorder.go
+++ b/v2/pkg/assetserver/body_recorder.go
@@ -1,0 +1,61 @@
+package assetserver
+
+import (
+	"bytes"
+	"net/http"
+)
+
+type bodyRecorder struct {
+	http.ResponseWriter
+	doRecord func(code int, header http.Header) bool
+
+	body        *bytes.Buffer
+	code        int
+	wroteHeader bool
+}
+
+func (rw *bodyRecorder) Write(buf []byte) (int, error) {
+	rw.writeHeader(buf, http.StatusOK)
+	if rw.body != nil {
+		return rw.body.Write(buf)
+	}
+	return rw.ResponseWriter.Write(buf)
+}
+
+func (rw *bodyRecorder) WriteHeader(code int) {
+	rw.writeHeader(nil, code)
+}
+
+func (rw *bodyRecorder) Code() int {
+	return rw.code
+}
+
+func (rw *bodyRecorder) Body() *bytes.Buffer {
+	return rw.body
+}
+
+func (rw *bodyRecorder) writeHeader(buf []byte, code int) {
+	if rw.wroteHeader {
+		return
+	}
+
+	if rw.doRecord != nil {
+		header := rw.Header()
+		if len(buf) != 0 {
+			if _, hasType := header[HeaderContentType]; !hasType {
+				header.Set(HeaderContentType, http.DetectContentType(buf))
+			}
+		}
+
+		if rw.doRecord(code, header) {
+			rw.body = bytes.NewBuffer(nil)
+		}
+	}
+
+	if rw.body == nil {
+		rw.ResponseWriter.WriteHeader(code)
+	}
+
+	rw.code = code
+	rw.wroteHeader = true
+}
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by mholt** [comment]
Cool, thank you for doing this :partying_face: I will try to test it soon

File: v2/internal/frontend/assetserver/assetserver.go [hunk 1]
```diff
@@ -226,3 +246,12 @@ func (d *AssetServer) logError(message string, args ...interface{}) {
 		d.logger.Error("[AssetServer] "+message, args...)
 	}
 }
+
+func (AssetServer) isRuntimeInjectionMatch(path string) bool {
+	if path == "" {
+		path = "/"
+	}
+
+	return strings.HasSuffix(path, "/") ||
```
leaanthony, line 255: Do we also need to explicitly check for a `.html` extension?

stffabi: That's only a precondition, it really only injects the runtime in the end if the content-type of the result is 'text/html'. See line 122

leaanthony:
I need to stop trying to review on the mobile app 😅

v2/internal/frontend/assetserver/assethandler.go
```diff
@@ -129,13 +129,14 @@ func (d *assetHandler) serveFSFile(rw http.ResponseWriter, req *http.Request, fi
...
-		url := req.URL.Path
-		if url != "" && url[len(url)-1] != '/' {
+		if !isDirectoryPath {
 			// If the URL doesn't end in a slash normally a http.redirect should be done, but that currently doesn't work on
-			// WebKit WebVies (macOS/Linux).
-			// So we handle this as a file that could not be found.
-			return os.ErrNotExist
+			// WebKit WebViews (macOS/Linux).
+			// So we handle this as a specific error
+			return fmt.Errorf("a directory has been requested without a trailing slash, please add a trailing slash to your request")
```
leaanthony, line 139: I'm wondering if there's a consequence to just accepting this as we already know it's a directory and I think the `path.Join()` below would still work? Maybe it's best to error but thought I'd ask your thoughts on it 👍

v2/internal/frontend/assetserver/assethandler.go
```diff
@@ -150,6 +151,8 @@ func (d *assetHandler) serveFSFile(rw http.ResponseWriter, req *http.Request, fi
 		if err != nil {
 			return err
 		}
+	} else if isDirectoryPath {
+		return fmt.Errorf("a file has been requested with a trailing slash, please remove the trailing slash from your request")
```
leaanthony, line 155: Same as above

leaanthony (approved):
Thanks for doing this @stffabi !

v2/internal/frontend/assetserver/assethandler.go
stffabi: Yeah, from our side everything would work, but I think this would give some problems for the users. 

Let's consider the following case, someone navigates to  `/test` which is a folder and would be handled by this case. The file `/test/index.html` is found and would be served. So for the WebView the url would still be `/test`. If the `index.html` contains some relative references e.g. `<script src="main.js"></script>`, the webview would then request `/main.js` instead of `/test/main.js`. But the user expects `/test/main.js` to be loaded.

stffabi: Normally Webservers would to a HTTP redirect in this case, but we currently can't do that since WebKit on macOS and Linux doesn't support HTTP 30x redirects.

v2/internal/frontend/assetserver/assethandler.go
stffabi: This would result in the same problem but in reverse order, a file in the root folder would be returned as file in a sub folder. So relative references would then be loaded from the sub folder instead of the root.

mholt:
One more question: Should the runtime and IPC be injected into the root index.html file even if the URL is not `/`? I'm trying an SPA for something, and currently I can only get the runtime and IPC scripts to be added if the URL is `/` and not `/foo/`, even though the response for `/foo/` is also the contents of `/index.html` just like `/` is, except without runtime & IPC.

stffabi:
> One more question: Should the runtime and IPC be injected into the root index.html file even if the URL is not `/`? I'm trying an SPA for something, and currently I can only get the runtime and IPC scripts to be added if the URL is `/` and not `/foo/`, even though the response for `/foo/` is also the contents of `/index.html` just like `/` is, except without runtime & IPC.

Yeah that's what this PR should do, it should inject into all html
Files, depending on the content type of the result when retrieving a folder url.
Can you please provide a minimal repro for this?

mholt:
@stffabi Thanks. Yeah, I think so:

```
$ wails init -n myproject -t vanilla
```

I think this step is optional, but I already did it, so I'll document it here: add a file called `html/foo/index.html` in the `frontend` directory, with a simple hello world: `<p>Hello world!</p>`

Run `wails dev`, then `curl http://localhost:34115/ | grep ipc.js` and you will see that the IPC is injected. But run `curl http://localhost:34115/foo/ | grep ipc.js` and you will see that it is not, even though if you remove the grep it's the same /index.html file that's being served.

Shouldn't IPC+runtime be injected into the root index file regardless of the URL path?

Thanks for your help!

stffabi:
@mholt thanks for this guide. I've just tested it with this PR and everything worked as expected.

Both curl `curl http://localhost:34115/ | grep ipc.js` and `curl http://localhost:34115/foo/ | grep ipc.js` returned successfully the patched HTML file and contained both `ipc.js` and `runtime.js`

I've tested on macOS but wouldn't expect this to differ between platforms. Would it be possible for you to check if you are using the right wails replace in your go.mod? Just to be sure.

```go
replace github.com/wailsapp/wails/v2 v2.2.0 => github.com/stffabi/wails/v2 v2.0.0-beta.20.0.20221220124958-d3b156dff78e
```

mholt:
Oh, I think I had the replace wrong :facepalm: Thanks

Now I'm getting it injected twice: once on the root index.html, and then after my page fetches `/foo/index.html` and injects its response into `<body>`, it's included there too. And I'm getting errors like `Callback 'application.App.SearchItems-3449944401' not registered!!!` -- probably something dumb on my end, but I just wanted to double-check.

Thanks!

mholt:
@stffabi I had a chance to go over this with @leaanthony which helped clear up my misunderstandings. And I am better equipped to clarify my request :)

So, even with this patch, if you do:

1. `wails init -n myproject -t vanilla`
2. `wails dev`
3. `curl -v http://localhost:34115/foo`

you'll notice that the HTTP response has the contents of `/index.html`, even though `/foo` was requested. This is the correct and expected behavior of Wails, serving `index.html` because there's no file called `foo`. However, if you `grep ipc.js` or `grep runtime` you won't find them. Yet if you `curl -v http://localhost:34115/` then you will see both ipc.js and runtime.js are injected.

I think what needs to happen is Wails should be injecting runtime and IPC into the index file regardless of the URL path, or regardless of whether a file is found at the URL path, I mean. Lea pointed to this code:


```
	path := req.URL.Path
	switch path {
	case "", "/", "/index.html":
		recorder := httptest.NewRecorder()
		d.handler.ServeHTTP(recorder, req)
		for k, v := range recorder.HeaderMap {
			header[k] = v
		}

		switch recorder.Code {
		case http.StatusOK:
			content, err := d.processIndexHTML(recorder.Body.Bytes())
			if err != nil {
				d.serveError(rw, err, "Unable to processIndexHTML")
				return
			}
			d.writeBlob(rw, indexHTML, content)

		case http.StatusNotFound:
			// *** right here, might need to processIndexHTML? ***
			d.writeBlob(rw, indexHTML, defaultHTML)
```

This is needed, I think, because not all SPAs use hashes for routing. In other words, some use `window.location.hash` for routing, but fragment URLs aren't desirable or possible sometimes, so `window.location.pathname` is used for routing instead.

Does that make sense?

I'm sorry for the confusion about this issue, I hope this helps clarify what I need, at least, and what I think is correct behavior. :)

leaanthony:
Perhaps we could introduce a flag to indicate what to do when URLs are "not found"? I'm thinking that an "SPA" mode could be used as well as allowing the dev to supply an html file to serve in this instance. Thoughts?

mholt:
@leaanthony Doesn't the AssetServer already invoke a user's handler if a resource isn't found?

leaanthony:
I believe we present a default 404 page.

stffabi:
@mholt I suspect there's still some confusion about all of this.

> you'll notice that the HTTP response has the contents of /index.html, even though /foo was requested. This is the correct and expected behavior of Wails, serving index.html because there's no file called foo.

What you are seeing that this curl `curl -v http://localhost:34115/foo` returns index.html is a `vite` thing and is not a functionality of Wails itself and therefore only applies to `wails dev`. In a production build you would see just a 404 when fetching the `/foo` url.

The code you pointed to is before the request is handed over to `vite`, and `vite` does internally return the `index.html` when returning from `d.handler`. So wails code is not involved in all of this and the location `// *** right here, might need to processIndexHTML? ***` is never reached, because `vite` returns a 200 status. Doing a curl to `vite` directly `curl -v http://localhost:3000/foo` one can see that it already does this fallback thing and returns a 200 status.

> I think what needs to happen is Wails should be injecting runtime and IPC into the index file regardless of the URL path, or regardless of whether a file is found at the URL path, I mean.

I can't see any way how we could detect that `vite` is returning the `index.html` for a missing `/foo`, unless there's a way to disable this behaviour in `vite` and doing this fallback handling in Wails itself.

Maybe we should meet us all together in a discord channel for a more 1:1 discussion on all of this. Maybe that would be easier for discussing this.

@leaanthony
 > Perhaps we could introduce a flag to indicate what to do when URLs are "not found"? I'm thinking that an "SPA" mode could be used as well as allowing the dev to supply an html file to serve in this instance. Thoughts?

This could be achieved by the developer with an AssetServer middleware and by adding custom `RuntimeInjectionMatchers`.

stffabi:
> @leaanthony Doesn't the AssetServer already invoke a user's handler if a resource isn't found?

If the file is not found on the Assets fs.FS it goes to the AssetHandler.

leaanthony:
Agree about the discord chat. Seems like the async chat here is confusing us all :-) Like I was amazed vite returns data on :34115 :-)

leaanthony:
@mholt - do you think this is good to go as is?

mholt:
Sorry I stepped away from this for a while. Once I discovered `<meta name="wails-options" content="noautoinject" />` I was using that. And at one point it was working, however now it seems to not be working. This is my HTML:

```
<!DOCTYPE html>
<html>
	<head>
		<title>Timelinize</title>

		<meta charset="utf-8">
		<meta name="viewport" content="width=device-width, initial-scale=1.0">

		<!-- add wails runtime and IPC scripts ourselves -->
		<meta name="wails-options" content="noautoinject" />
		<script src="/wails/ipc.js"></script>
		<script src="/wails/runtime.js"></script>
...
```

But it still seems to be injected by Wails (you can see the script tags twice; the ones at the top are injected):

image

This is confirmed by the console, which says that wails is connected... twice:

image

I might be misunderstanding something about this still...

stffabi:
@leaanthony I've just rebased the PR to the master branch.

leaanthony (approved):

Issue #1788, opened 2022-08-24
Notification support


tonyho:
### Is your feature request related to a problem? Please describe.

We want to show a notification toast to hint information to users, but it seems no such function in Wails and its roadmap.

### Describe the solution you'd like

Notification should be added.

### Describe alternatives you've considered

Toast function which is similiar to Android Toas.

Lyimmi:
I don't know if there is a plan for notifications, but beeep is a good package for native notifications IMO: 
https://github.com/gen2brain/beeep .

leaanthony:
Added to Roadmap 👍

Lyimmi:
As @leaanthony suggested, I'm trying to put together an API proposal and I'm looking at the notification capabilities of Windows, macOs and linux. Windows has the most extensive functionality, linux and macOS has a "simpler" API.

My question would be, what is the goal, a common API with functions that work everywhere, or one where each OS can have its own settings, or some form of hybrid from the two with an override style solution.

**Case one:**
A common API for all supported OS with "limited" functionality.
 - title
 - message
 - icon
 - image (only windows)
 - actions

We may lose some functionality on windows, but when you create a notification it would behave the same way on every OS.

**Case two:**
An API with separate options for each OS.

One would have to configure every notification for each OS they want to support, probably the only common options for a notification would be the title and a message (and maybe action buttons).

**Case three:**
An API with a "limited" common functionality but with separate options for each OS.

The main API would work as described in the first case with the basic functionality, but has OS specific options like timing, sound, position, or for windows status/download bars, etc.

leaanthony:
Awesome @Lyimmi ! As far as API goes, it's ok to have a single struct of options for all platforms where options are ignored if not supported 👍

Lyimmi:
Here is my first draft of the Notification API. I think if we want to have more than a title, message and an icon, we need to have OS specific options. As far as I can tell Electron only supports these three options as default.

Even the timeout is different on each os, on linux it's set in milliseconds, on windows it's "short" and "long", on macOs its not available. Handling the actions / buttons is totally different as well.

Maybe the click on the notification could be handled commonly as well.

It is more of a conversation starter then a complete API.

```golang

// LinuxActionInvokedHandler handles signals returned by activating actions
type LinuxActionInvokedHandler func(signal *LinuxNotificationActionInvokedSignal, options ...interface{})

// LinuxNotificationAction represents a Notification action for a notification
type LinuxNotificationAction struct {

[...]

PR #2206, opened 2022-12-14
Feature/1788 notification support

lyimmi:
**This PR adds the runtime API for notifications. #1788**

Currently the Linux implementation is the main target for this PR, macOS and windows is not complete, more like a placeholder for now. All three Linux, Windows and macOS implementation is capable of showing basic notifications.

**Application options**
I created a new `AppID` property in the App options, because all three implementation requires it. It defaults to "wails".

**Icons / Images / Sounds**
Each operating system has its own ways to load images and sounds, but mostly they use absolute paths. To be able to use embedded resources and to create a common API, before a notification is sent the images & sounds are written to a process and resource specific tmp file and a path to that file is used. These files are written on each notification.

Current functionality that is working. I wrote the documentation as well.
- Linux:
  - icon
  - title
  - message
  - actions
    - callback
  - urgency
  - notification replacement
  - sound
  - on close callback
  - on show callback
  - notification close
- Windows:
  - icon
  - title
  - message
  - timeout
  - actions (actions are only shown signals are not handled)
- macOS:
  - title
  - message
  - timeout
  - actions (actions are only shown signals are not handled)

**Below is an example app.go file that can be used as a test in a clean wails project**

```golang
package main

import (
	"context"
	_ "embed"
	"fmt"
	"github.com/wailsapp/wails/v2/pkg/runtime"
	"log"
	"time"
)

type App struct {
	ctx context.Context
}

func NewApp() *App {
	return &App{}
}

func (a *App) startup(ctx context.Context) {
	a.ctx = ctx
}

var prevID = uint32(0)
var count = 1

//go:embed build/appicon.png
var appicon []byte

func (a *App) Greet(name string) string {

	closeChan := make(chan bool)

	err := runtime.SendNotification(a.ctx, runtime.NotificationOptions{
		AppIcon: appicon,
		Title:   "Hello!",
		Message: fmt.Sprintf("Hello %s, previous notification id: %d; count :%d", name, prevID, count),
		Timeout: 60 * time.Second,
		Close:   closeChan,
		LinuxOptions: &runtime.LinuxNotificationOptions{
			Sound: &runtime.LinuxNotificationSound{
				Name:     "message-sent-email",
				File:     nil,
				Suppress: false,
			},
			Actions: []runtime.LinuxNotificationAction{
				{Key: "default", Label: "", OnAction: func(notificationID uint32) {
					log.Println(notificationID)
				}},
			},
			ReplacesID: prevID,
			Urgency:    1,
			OnShow: func(notificationID uint32) {
				prevID = notificationID
				count++
			},
			OnClose: func(notificationID uint32, reason string) {
				log.Println(notificationID, reason)
			},
		},
	})

	if err != nil {
		return err.Error()
	}

	go func() {
		time.Sleep(5 * time.Second)
		closeChan <- true
	}()

	return fmt.Sprintf("Hello %s, It's show time!", name)
}

```


Lyimmi:
Halfway I forgot to use the logger and used fmt instead. I'll fix that.

Lyimmi:
The close channel currently could cause a few problems.
If a user closed the message, the close channel is never drained and could hang forever, or leak goroutines.

Would it be OK, to return a close function and an error? Similar to context witch cancel?

leaanthony:
What's the close channel for? To indicate the notification was closed? Would a callback be an option?

Lyimmi:
It should be used to close / remove a notification programmatically, for example it's not needed anymore or you want to replace it without showing two messages.

As far as I can tell only Linux has the option to replace an existing notification by Id.

Lyimmi:
It should be used to close / remove a notification programmatically, for example it's not needed anymore or you want to replace it without showing two messages.

As far as I can tell only Linux has the option to replace an existing notification by Id.

Currently the Linux options has an onClose callback that is exactly what you are describing it could be moved out to the common options if all three platforms are able to report it back.

leaanthony (approved):

leaanthony:
@Lyimmi Just noticed the macOS build failed: https://github.com/wailsapp/wails/actions/runs/3860653108/jobs/6581118661

Lyimmi:
Bummer, sorry. I think a package is missing from go mod.

I'll fix it.

Lyimmi:
@leaanthony I looked over and tested it on all tree platforms again.

Fixed the documentations page number after merging master, the clipboard page had the same index.

- macOS:
   - Fixed the missing package in go mod
- Windows:
   - Fixed the sound options
   - Updated the docs to warn about character encoding in the notification.
- Linux:
   - Changed the notification urgency option to a type to be more straightforward.
   - Updated the docs for this change.

I have a question, currently on Windows and mac the notification actions only shown there is no functionality behind them. Shouldn't we remove them from the API until they work properly? The docs say they are only shown, but I think it could cause some misunderstanding.

leaanthony (approved):

leaanthony:
Yeah seems like a good idea.

Lyimmi:
I'll remove the actions from mac and windows.

ValentinTrinque:
Regarding MacOS notifications, I think we should use the native ones. Those relaying on the terminal capabilities seems nice at first, but in the end, it just confuse the users. It am currently working on an implementation in objc that should play well with the notification center. It's not fully functional at the moment, and... it requires CGO, which may not fit the "No CGO" claim of this project. Let me know if you prefer native app notification.

stffabi:
> Regarding MacOS notifications, I think we should use the native ones. Those relaying on the terminal capabilities seems nice at first, but in the end, it just confuse the users (cannot change the icon, clicking the notification opens the terminal and not the app).
> 
> I am currently working on an implementation in objc that should play well with the notification center. It's not fully functional at the moment, and... it requires CGO, which may not fit the "No CGO" claim of this project. Let me know if you prefer native app notification.

That would be awesome to have a native implementation for this on macOS. The "No CGO" claim is currently only for Windows, we would be more than happy if you plan to contribute this to the project.

ValentinTrinque:
@stffabi I am thinking out loud, we could support both CGO and No-CGO™ within our build directive. That way no contraints. As a kind of "progressive enhancement". 🤔 

Anyway, i will past you my **non-working** code, as reference. For some reason, I can get the code running, but the MacOS authorization refuses to show. My diggings led me to "packaging issues" but I don't think that's that. The package seems just fine. I think it has more to do with macos application lifecyle. Maybe someone with more knowledge on how MacOS works can fix it.

ValentinTrinque:
Here is the MacOS code:
https://github.com/vegaprotocol/vegawallet-desktop/blob/180bd9ab1c7b828961bc1d5d9500edd4a90d308f/os/notification/notify_darwin.h

Here is the CGO binding:
https://github.com/vegaprotocol/vegawallet-desktop/blob/180bd9ab1c7b828961bc1d5d9500edd4a90d308f/os/notification/notify_darwin.go

ValentinTrinque:
@leaanthony  well, I think we both hit the same wall. These are the bits that doesn't work

ValentinTrinque:
Here is what I understand. The new notification center is very peaky. You need to request access. But how can it ensure it's not a software that tries to abuse it ? By looking at the packaging properties, suchas a the signature, the certificate, and all...

That sounds like a good explanation but it could also be a completely made up one. That's just what I understood. Not a MacOS dev, so it's really a wild guess.

leaanthony:
> Here is the MacOS code:
> https://github.com/vegaprotocol/vegawallet-desktop/blob/180bd9ab1c7b828961bc1d5d9500edd4a90d308f/os/notification/notify_darwin.h
> 
> Here is the CGO binding:
> https://github.com/vegaprotocol/vegawallet-desktop/blob/180bd9ab1c7b828961bc1d5d9500edd4a90d308f/os/notification/notify_darwin.go

@ValentinTrinque - Thanks for sharing these links. Is there anything extra you do? Everything I tried crashed.

Issue #2053, opened 2022-11-05
Flag to disable backend rebuild in dev


hkeide:
### Is your feature request related to a problem? Please describe.

I use autosave in my IDE and rebuilding the Go code and thereby resetting the runtime state is too intrusive for me.

### Describe the solution you'd like

Add a flag `-nogorebuild` to `wails dev` to disable the automatic backend rebuild logic

### Describe alternatives you've considered

Adding `-e dummyextension` to the `dev` command seems to accomplish what I want, but I imagine other people may want this behavior too. 

Alternatively this trick could be added to the docs since it seems to work fine.

### Additional context

Happy to create a PR to modify docs or to add an explicit flag if this is desirable.

leaanthony:
Yeah feel free to raise a PR for these 👍👍

OlegGulevskyy:
Me too during my project dev on wails I realised how much go rebuild on each save `can` be inconvenient, not always is, of course.
I am guessing @hkeide don't have plans to pick up on this, so I could take a swing at this.

To take this a step further, I am thinking of a `wails.json` property to add a glob pattern of files to ignore, when save is occurred. This should give some flexibility - if you want, you ignore all, or just specific files that you don't want go to rebuild.
What do you think @leaanthony ? I am not familiar with the file watcher that `wails` use but it shouldn't be too much of a problem, I imagine?

OlegGulevskyy:
By the way @leaanthony , I think this is a duplicate of #2032 , if I am not mistaken.

leaanthony:
Similar but not the same.

haukened:
Fixed in PR #2870

PR #2214, opened 2022-12-15
Allow ignore paths for dev

OlegGulevskyy:
Based on the issue #2053 the idea here is simple:
- adding a property ( `ignorereloadpaths` ) to `wails.json` file that accepts an array of globs
- you can mute folders, nested folders etc
- you can mute separate files in the project root folder or nested folders
- to mute all files, just add `"*"` glob

Need to update schema and come up with couple of unit tests, but the functionality of PR should be working as expected


leaanthony:
@OlegGulevskyy - I'm not sure but I think this may have been added via a different PR?

OlegGulevskyy:
Yes, I think this was already done by someone else, maybe in different manner but in the same spirit

PR #2215, opened 2022-12-15
Add nosyncgomod flag

leaanthony:
This is an alternative approach to not having the go.mod synced to your wails CLI version. It fixes the tests on master.

@mholt - could you please confirm this works for you? Adding `-nosyncgomod` to `dev` or `build` should trigger it 👍

### DIFF
--- a/v2/cmd/wails/build.go
+++ b/v2/cmd/wails/build.go
@@ -107,9 +107,11 @@ func buildApplication(f *flags.Build) error {
 		return err
 	}
 
-	err = gomod.SyncGoMod(logger, f.UpdateWailsVersionGoMod)
-	if err != nil {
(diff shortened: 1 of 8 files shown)


stffabi (approved):
LGTM 👍

mholt:
Hi, sorry for the late reply. Been down with the flu for a few days! I will try this out ASAP, but I also don't want you to be blocked by me while I catch up with things. The previous patch that disables syncing implicitly based on a revision/non-tagged commit works well for me, and it looks like this patch simply adds an explicit CLI flag to forcefully disable syncing. If that's right, then I think this is great too. At a glance it doesn't look like it would _break_ anything, at least for my use case. :+1:

PR #2218, opened 2022-12-18
Update libwebkit package name for Fedora

NullCode1337:
(added a new commit in response to member comment)

Without this change, wails doctor reports libwebkit not being installed (even though it already is)
Old version of libwebkit has been retained for backwards compatibility


leaanthony:
@NullCode1337 just confirming this has been tested.

NullCode1337:
> @NullCode1337 just confirming this has been tested.

It has
image-3

leaanthony (approved):
Thanks for this!

leaanthony (approved):

Issue #1743, opened 2022-08-15
Common Clipboard Support


leaanthony:
It would be good to provide an API for getting and setting the clipboard in both Go and JS.

csterritt:
I second this -- and I haven't forgotten to document the current situation, with respect to what I found, I've just... forgotten :-)

Sorry. I did look at doing a doc update, and I *think* what I need to do is do a new version? It seemed like Docusaurus wouldn't even recognize the existence of a new file unless I did one. Is that right? Is there a better way to add a new file?

Thanks!

misitebao:
We can create a new package to develop extension functions, similar to [""](https://github.com/vueuse/vueuse) for Vue.

Lyimmi:
I don't know what is the status of this ticket, but I looked into this.
If it is enough for us to handle text on the clipboard, I think this could be solved relatively easily.

I think we could create two runtime functions with "text" in their name to leave the option open for images etc. in the future.
 - GetClipboardText()
 - SetClipboardText()

On windows we could use [something like this.](https://github.com/atotto/clipboard/blob/master/clipboard_windows.go) (didn't test it my self)
On macOS we could use [something like this.](https://github.com/atotto/clipboard/blob/master/clipboard_darwin.go) (didn't test it my self)
On Linux of course it's more complicated, because of X11 and Wayland, however we could use GTK as a "middleware" and with that we could solve it on both window manager without any user installable dependency.

My working Linux experiment, uses gtk to get/set the current clipboard on X11 and Wayland as well.

```C
static gchar* GetClipboardText() {
	GtkClipboard *clip = gtk_clipboard_get(GDK_SELECTION_CLIPBOARD);
	return gtk_clipboard_wait_for_text(clip);
}

static void SetClipboardText(gchar* text) {
	GtkClipboard *clip = gtk_clipboard_get(GDK_SELECTION_CLIPBOARD);
	gtk_clipboard_set_text(clip, text, -1);
}

```

```golang
func (f *Frontend) GetClipboardText() string {
	var text string
	var wg sync.WaitGroup
	wg.Add(1)
	invokeOnMainThread(func() {
		ctxt := C.GetClipboardText()
		defer C.free(unsafe.Pointer(ctxt))
		text = C.GoString(ctxt)
		wg.Done()
	})
	wg.Wait()
	return text
}

func (f *Frontend) SetClipboardText(text string) {
	invokeOnMainThread(func() {
		ctxt := (*C.gchar)(C.CString(text))
[...]

PR #2228, opened 2022-12-22
Add common text clipboard support

Lyimmi:
This PR adds basic text clipboard functionality to GO and JS runtime as discussed in this issue #1743

I implemented the clipboard functions on Windows, Linux and macOS, however I could not test it on mac. Windows and Linux is tested and works.

The JS runtime uses the GO implementation in the background via runtime system call because the javascript API prompts the user for permission, using the GO functionality circumvents this problem.

The windows implementation uses win32 and has no dependency, on Linux I used GTK to handle both X11 and Wayland, on macOS I used os.exec with pbcopy / pbpaste.

GO runtime
```golang
err := runtime.ClipboardSetText(ctx, "write this to clipboard")

text, err := runtime.ClipboardGetText(ctx)
```

JS runtime
```javascript
ClipboardSetText("write this to clipboard").then(status => {
    console.log(status);
}).catch(err => {
    console.error(err)
})

ClipboardGetText().then(text => {
    console.log(text)
}).catch(err => {
    console.error(err);
});
```


leaanthony:
Just testing this on macOS and it works great @Lyimmi!  There's some native APIs to access the clipboard on macOS but I'm ok with this approach for v2. Awesome job 👍

leaanthony (approved):

omega32:
Hi. Is this working under Linux?
It works perfectly on Windows, but I get an error trying to access the clipboard under Linux.

Lyimmi:
Hi, It should, I tried this on a bunch of distros.
Could you please open an issue for this with a bit more info and the error message?

PR #2231, opened 2022-12-23
Improve `crowdin` init to be able to work locally

OlegGulevskyy:
This just refactors the way `crowdin` initialised, to not throw away from the docs build time should there be no personal token in env variables.
If there is no token, it will log a warning message, but the rest of the functionality is available to the developer to keep on contributing


Lyimmi:
Tried it out works for me nicely!

leaanthony (approved):

leaanthony:
Thanks @OlegGulevskyy !

PR #2110, opened 2022-11-19
[assetserver] Add support for serving the index.html file when requesting a directory

stffabi:
Furthermore this PR improves some error messages for the AssetServer Options and adds some additional checks. 

Fixes #2109

PR #2235, opened 2022-12-23
[assetHandler] Add request path checks

stffabi:
I've split out these checks from PR #2203 because that PR might take longer until it has been fully discussed. But still it would be good to have these checks integrated before the next release. Especially because this is a follow-up for PR #2110 which is not yet released.


leaanthony (approved):
Nice one!

Issue #2264, opened 2022-12-31
Support WebView Hardware acceleration option


lyimmi:
### Is your feature request related to a problem? Please describe.

Sometimes we need to be able to disable Hardware / GPU acceleration in webview.
I'd like to make it an app option, so developers can set it on start from a config (This way an application's users could switch it on and off for themselves theoretically), or baked in.

I have a solution for this on Windows and Linux, I'll make a PR when its ready.

### Describe the solution you'd like

I'd like to add a new WebviewGpuAcceleration property to the app options. With the following structure.

```golang 

// WebviewGpuAcceleration values used for determining the hardware acceleration policy.
//   - WebviewGpuAccelerationOff
//   - WebviewGpuAccelerationOn
//   - WebviewGpuAccelerationOnDemand
type WebviewGpuAcceleration int

const (
	// WebviewGpuAccelerationOn Hardware acceleration is always enabled, even for websites not requesting it.
	WebviewGpuAccelerationOn WebviewGpuAcceleration = 0
	// WebviewGpuAccelerationOnDemand Hardware acceleration is enabled/disabled as request by web contents. (linux only falls back to "on" on windows)
	WebviewGpuAccelerationOnDemand WebviewGpuAcceleration = 1
	// WebviewGpuAccelerationOff Hardware acceleration is always disabled, even for websites requesting it.
	WebviewGpuAccelerationOff WebviewGpuAcceleration = 2
)
```

Usage would look like this:

```golang

err := wails.Run(&options.App{
  Title:            "wailst",
  Width:            1024,
  Height:           768,
  Assets:           assets,
  BackgroundColour: &options.RGBA{R: 27, G: 38, B: 54, A: 0},
  OnStartup:        app.startup,
  Frameless:        true,
  CSSDragProperty:  "--wails-draggable",
  CSSDragValue:     "drag",
  Bind: []interface{}{
	  app,
  },
  Windows: &windows.Options{
	  WindowIsTranslucent:  true,
	  WebviewIsTransparent: true,
  },
  WebviewGpuAcceleration: options.WebviewGpuAccelerationOff, // <- it would be set here
})
```

### Additional context

On Windows it can be set via browser flags in chromium.
On Linux it can be set via `webkit_settings_set_hardware_acceleration_policy(s,o)`
On macOs as far as I can tell this is not an option to be set. (Can't find anything about it in the docs.)

leaanthony:
[...]

PR #2266, opened 2023-01-01
Support GPU acceleration settings on Windows & Linux

Lyimmi:
This PR adds support for webview GPU settings on Windows and Linux as discussed in issue #2264.

**Windows:**

- Created a new option `WebviewGpuIsDisabled` in windows.Options
- Modified the new webview loader to be able to accept chromium switches.
- When `WebviewGpuIsDisabled` set to true the webview will be initialized with the `--disable-gpu` switch.

**Linux:**

- Created a new option `WebviewGpuPolicy` in linux.Options.
- Created a new type `WebviewGpuPolicy` and three constants to be used as options.
  - WebviewGpuPolicyAlways
  - WebviewGpuPolicyOnDemand
  - WebviewGpuPolicyNever
- When the webview is set up the hardware acceleration policy is set based on the option. The default setting is WebviewGpuPolicyAlways, if it is not a valid option for the system GTK ignores it and sets an appropriate value.

**Mac:**
This setting is not available.

### DIFF
--- a/v2/internal/frontend/desktop/linux/window.go
+++ b/v2/internal/frontend/desktop/linux/window.go
@@ -214,7 +214,7 @@ gboolean close_button_pressed(GtkWidget *widget, GdkEvent *event, void* data)
     return TRUE;
 }
 
-GtkWidget* setupWebview(void* contentManager, GtkWindow* window, int hideWindowOnClose) {
+GtkWidget* setupWebview(void* contentManager, GtkWindow* window, int hideWindowOnClose, int gpuPolicy) {
 	GtkWidget* webview = webkit_web_view_new_with_user_content_manager((WebKitUserContentManager*)contentManager);
 	//gtk_container_add(GTK_CONTAINER(window), webview);
 	WebKitWebContext *context = webkit_web_context_get_default();
@@ -228,6 +228,21 @@ GtkWidget* setupWebview(void* contentManager, GtkWindow* window, int hideWindowO
 
 	WebKitSettings *settings = webkit_web_view_get_settings(WEBKIT_WEB_VIEW(webview));
 	webkit_settings_set_user_agent_with_application_details(settings, "wails.io", "");
+
+	if(gpuPolicy < 0 || gpuPolicy > 2) {
+		gpuPolicy = 1; // fall back to on_demand when an unknown policy is set
+	}
+	switch (gpuPolicy) {
+		case 0:
+		  webkit_settings_set_hardware_acceleration_policy(settings, WEBKIT_HARDWARE_ACCELERATION_POLICY_ALWAYS);
+		  break;
+		case 1:
+		  webkit_settings_set_hardware_acceleration_policy(settings, WEBKIT_HARDWARE_ACCELERATION_POLICY_ON_DEMAND);
+		  break;
+		case 2:
+		  webkit_settings_set_hardware_acceleration_policy(settings, WEBKIT_HARDWARE_ACCELERATION_POLICY_NEVER);
+		  break;
+	}
 	return webview;
 }
 
@@ -706,7 +721,12 @@ func NewWindow(appoptions *options.App, debug bool) *Window {
 	C.webkit_user_content_manager_register_script_message_handler(result.cWebKitUserContentManager(), external)
 	C.setupInvokeSignal(result.contentManager)
 
-	webview := C.setupWebview(result.contentManager, result.asGTKWindow(), bool2Cint(appoptions.HideWindowOnClose))
+	webview := C.setupWebview(
+		result.contentManager,
+		result.asGTKWindow(),
+		bool2Cint(appoptions.HideWindowOnClose),
+		C.int(appoptions.Linux.WebviewGpuPolicy),
+	)
 	result.webview = unsafe.Pointer(webview)
 	buttonPressedName := C.CString("button-press-event")
 	defer C.free(unsafe.Pointer(buttonPressedName))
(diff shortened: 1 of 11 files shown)


v2/internal/frontend/desktop/linux/window.go
```diff
@@ -228,6 +228,21 @@ GtkWidget* setupWebview(void* contentManager, GtkWindow* window, int hideWindowO
 
 	WebKitSettings *settings = webkit_web_view_get_settings(WEBKIT_WEB_VIEW(webview));
 	webkit_settings_set_user_agent_with_application_details(settings, "wails.io", "");
+
+	if(gpuPolicy < 0 || gpuPolicy > 2) {
+		gpuPolicy = 1; // fall back to on_demand when an unknown policy is set
```
stffabi, line 233: Should we add this as the `default` case?

v2/internal/frontend/desktop/windows/go-webview2/pkg/edge/create_env_native.go
```diff
@@ -12,7 +12,7 @@ import (
 	"golang.org/x/sys/windows"
 )
 
-func createCoreWebView2EnvironmentWithOptions(browserExecutableFolder, userDataFolder string, environmentCompletedHandle *iCoreWebView2CreateCoreWebView2EnvironmentCompletedHandler) error {
+func createCoreWebView2EnvironmentWithOptions(browserExecutableFolder, userDataFolder string, environmentCompletedHandle *iCoreWebView2CreateCoreWebView2EnvironmentCompletedHandler, additionalBrowserArgs string) error {
```
stffabi, line 15: Should we return an error or panic if `additionalBrowserArgs` is not empty and the old loader is used?

stffabi (approved):
LGTM. Awesome, thanks for doing this.

Just two minor comments and maybe you could also already add the changelog entry.

v2/internal/frontend/desktop/windows/go-webview2/pkg/edge/create_env_native.go
Lyimmi: I don't think so, it is not really an error, it's just ignored.

Maybe we should write a warning to the terminal, that the arguments are ignored with this loader?

stffabi: It's clearly something that a developer wanted but is ignored. 

Since we say we will drop support I would as a developer prefer to get an error if something doesn't work instead of just a silent ignore or just a warning.

stffabi: We might add more flags in the near future and those might change the behaviour in a more relevant way, e.g enabling some features. And then a silent ignore would just frustrate developers if it's silently ignored.

stffabi: Or let's just implement the support on the old loader by setting the env variable here `WEBVIEW2_ADDITIONAL_BROWSER_ARGUMENTS` instead of using the EnvironmentOptions.

https://github.com/wailsapp/wails/blob/d3b4105f7598ba51501541ebd4ee564ca75a70c6/v2/internal/frontend/desktop/windows/go-webview2/webviewloader/native_module.go#L154-L164

v2/internal/frontend/desktop/linux/window.go
Lyimmi: You mean that make the "on demand" policy the default?
We could do that, I set the "always" as a default because for me on popOS 22.04 and on a raspberry GTK sets it to "always" by default.

Lyimmi:
I found a bug. If you don't set the linux options, the setup crashes. I'm fixing it now!

v2/internal/frontend/desktop/linux/window.go
stffabi: I meant instead of this if and set gpuPolicy we could also add a default case to the switch which does set on_demand.

Lyimmi: Ahh yes... I don't know why I did it this way. That would be cleaner. I'll changit.

Lyimmi: Changed and committed.

v2/internal/frontend/desktop/windows/go-webview2/pkg/edge/create_env_native.go
Lyimmi: I didn't see that option. We couléd do that, but I don't know how we should do that.

Pass the addition arguments further down to the preventEnvAndRegistryOverrides?

stffabi: Yeah, just passing them should be fine.
You can also leave that away and I'll do it in a future PR.

Lyimmi: So for now leave it as is or should I throw an error?

stffabi: You can leave it as is and I'll add it.

Lyimmi:
> LGTM. Awesome, thanks for doing this. +1
> 
> Just two minor comments and maybe you could also already add the changelog entry.

Thanks @stffabi, I added the change lo as well!

stffabi:
You're welcome, thanks for doing this 🙏

leaanthony (approved):
This is great @Lyimmi !!

PR #2267, opened 2023-01-02
[windows] Add support for AdditionalBrowserArgs to the legacy Webview2Loader

stffabi:
Follow-up PR for #2266


leaanthony (approved):
LGTM. Just checking the switch in arguments was intended?

stffabi:
Yeah was intended because we don't need that anywhere and it could probably clash with the env variable.

PR #2269, opened 2023-01-03
Add EnableFraudulentWebsiteDetection to options

### LINKED ISSUE #2163: Disable SmartScreen / FraudulentWebsite detection

Microsoft uses SmartScreen per default in WebView2 to detect if a user is accessing a phishing or malware website. The same applies to Apple, they use a "FraudulentWebsite" detection service.

Both services send information about the navigation to Microsoft/Apple. It's not clear if this only applies to External Websites or also to custom protocols or what is included.

It would be great to have the possibility to disable this detection. Maybe even better per default disable it and the developer needs to opt-in.

How to disable it:
- Apple [fraudulentWebsiteWarningEnabled](https://developer.apple.com/documentation/webkit/wkpreferences/3335219-fraudulentwebsitewarningenabled?language=objc)  
- Microsoft: https://github.com/MicrosoftEdge/WebView2Feedback/issues/834

_No response_

### Additional context

_No response_

stffabi:
This allows to explicitly opt-in to scan services for fraudulent
content, such as malware or phishing attempts.

Fixes #2163

### DIFF
--- a/v2/internal/frontend/desktop/darwin/Application.h
+++ b/v2/internal/frontend/desktop/darwin/Application.h
@@ -17,7 +17,7 @@
 #define WindowStartsMinimised 2
 #define WindowStartsFullscreen 3
 
-WailsContext* Create(const char* title, int width, int height, int frameless, int resizable, int fullscreen, int fullSizeContent, int hideTitleBar, int titlebarAppearsTransparent, int hideTitle, int useToolbar, int hideToolbarSeparator, int webviewIsTransparent, int alwaysOnTop, int hideWindowOnClose, const char *appearance, int windowIsTranslucent, int debug, int windowStartState, int startsHidden, int minWidth, int minHeight, int maxWidth, int maxHeight);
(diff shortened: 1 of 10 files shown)


stffabi:
@leaanthony do you think it should be opt-in or opt-out? I tend it to be opt-in but that will change the behaviour of existing apps.

leaanthony (approved):
LGTM

leaanthony:
> @leaanthony do you think it should be opt-in or opt-out? I tend it to be opt-in but that will change the behaviour of existing apps.

I think it should be opt-in for v2 so that it doesn't change behaviour and opt-out for v3.

stffabi:
I've updated the PR to invert the logic. @leaanthony would you mind taking another look?

leaanthony:
I had a think about this overnight.... I don't recall anyone reporting this warning so maybe disabling by default is actually fine?

stffabi:
Reverted back to the original version.

leaanthony:
Nice one! Thanks for doing that 🙏

PR #2275, opened 2023-01-06
linux separate window c and go file

Lyimmi:
In this PR I moved the C code from v2/internal/frontend/desktop/linux/window.go to its own separate window.c and window.h files.

- I tried to reorganize the C code to make it a bit more readable.
- Moved the typedefs to the window.h file to both window.go and window.c can access them.
- Changed the function names to make them "consistent":
  - functions called from window.go start with an uppercase letter.
  - functions only called from window.c start with a lowercase letter and made them static.

I did not change the functionality, only the structure and the formatting of the code.

### DIFF
--- a/v2/internal/frontend/desktop/linux/window.c
+++ b/v2/internal/frontend/desktop/linux/window.c
@@ -1,0 +1,694 @@
+#include <JavaScriptCore/JavaScript.h>
+#include <gtk/gtk.h>
+#include <webkit2/webkit2.h>
+#include <stdio.h>
+#include <limits.h>
+#include <stdint.h>
+#include "window.h"
+
+// These are the x,y,time & button of the last mouse down event
+// It's used for window dragging
+static float xroot = 0.0f;
+static float yroot = 0.0f;
+static int dragTime = -1;
+static uint mouseButton = 0;
+
+// casts
+void ExecuteOnMainThread(void *f, gpointer jscallback)
+{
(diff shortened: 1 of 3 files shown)


leaanthony:
Thanks for doing this. Looks tidy!

Did you try changing a method (in .h and .C) and see if the changes got picked up?

Lyimmi:
When I renamed the functions and moved them around I built the test app constantly and did not notice any issue with changes not getting built.

Lyimmi:
I read about the issue you mentioned, that the go file is not changed the imported C code is not recompiled.

I did not face this, but to be sure I'll try to reproduce it.

Lyimmi:
Ok, it was fast.

So it seems that go doesn't pick up the changes, if the modified c/h file is not next to the go file that is included in. That is why I didn't run into this.
Because my window.c, window.h, window.go files are in the same directory.

On the image below you can see the file tree.
test.h is included in window.c, test_sub.h is included in test.h.
`window.c <- test.h <- test_sub.h`

If I change test.h go recompiles it every time, but if change test_sub.h it uses the cache and the changes are ignored.

image

leaanthony:
Thanks for testing that @Lyimmi! I remember hitting an issue with the Go cache and .h files a while back. This may well have been the scenario.

Lyimmi:
I think so too.

leaanthony (approved):
LGTM!

PR #2288, opened 2023-01-10
[windows] Disable frameless decorations during fullscreen

stffabi:
Otherwise this leads to thin transparent lines on the window borders on older Windows versions.


leaanthony (approved):
LGTM!

Issue #2287, opened 2023-01-09
Theme switch issue (Windows 10)


oSethoum:
### Description

~~When the window is frameless it's unable to be resized from the edges or the sides (left, bottom, right),~~ the second issue is when I change theme between Dark and Light using `WindowSetLightTheme` or `WindowSetDarkTheme` the changes don't apply immediately. Also, after upgrading to the master branch changing the `BackgroundColour` also don't apply immediately. 

#### ~~frameless resize issue~~
~~1- Make sure you are on windows 10
2- Make the window frameless and try to resize it from sides (left, right, bottom) or edges.~~

#### background color issue
1- Make sure you are on windows 10
2- Keep the frame of the window to see the changes
3- Change Theme or BackgroundColour and see no immediate update. 

### Expected behaviour

~~1- Frameless window to be resizable.~~
2- Immediate change after changing the theme or the backgroundcolour.

### Screenshots

https://user-images.githubusercontent.com/88779394/211408892-61a11a55-53f9-4730-8279-d1e350e43948.mp4

### System Details

```shell
# System

OS           | Windows 10 Pro
Version      | 2009 (Build: 19045)
ID           | 22H2
Go Version   | go1.19.4
Platform     | windows
Architecture | amd64

# Wails

Version  | v2.3.1
Revision | ecc451937495018ca3ffee9dfbb41e3c08cc100c
Modified | true

# Dependencies

Dependency | Package Name | Status    | Version
WebView2   | N/A          | Installed | 108.0.1462.76
npm        | N/A          | Installed | 9.2.0
*upx       | N/A          | Installed | upx 4.0.1
*nsis      | N/A          | Available |
* - Optional Dependency
```

### Additional context

This was tested on the master branch. 

Edit: The resize issue has been resolved with PR #2289

stffabi:
Thanks for creating this issue.

I've just tested this on a clean `Windows 10 Pro  | 2006 (Build: 19045) | 22H2` and wasn't able to reproduce the resize issue. The test was done with a clean `wails init` project. Do you see the same problem when using a clean `wails init` project?

oSethoum:
@stffabi I just `wails init` new wails app and still not resizing in frameless mode, I used the command
```console
wails init -n app -t react-ts
```

stffabi:
I was able to reproduce it with the `react-ts` template. PR #2289 will fix that.

stffabi:
[...]

PR #2289, opened 2023-01-10
[runtime] Frameless resize now sets the cursor on documentElement

stffabi:
Otherwise if the body does not cover the full viewport the cursor won't be shown outside of its rectangle.


stffabi:
This PR fixes the resizing issues mentioned in #2287

leaanthony (approved):

PR #2299, opened 2023-01-14
[windows] Some improvements to fullscreen

stffabi:
- Do not show a white border for windows in fullscreen mode and when DisableWindowIcon is used
- Prevent a short showing of the window titlebar for frameless window when switching fullscreen/unfullscreen


leaanthony (approved):
LGTM! Thanks for taking the time to do this 🙏

PR #2302, opened 2023-01-15
[runtime] Improve `--wails-draggable` by starting a drag immediately but only for the first click in a series

stffabi:
This removes the sometime lagging experience when starting to drag on Windows. But it's still possible to use double-click events for adding e.g. a WindowToggleMaximise.

Tested on:
- [x] Windows
- [x] macOS
- [x] Linux


stffabi:
This doesn't work well on Linux, double click events really often get swallowed. So the previous solution did work far better on Linux. I'll update the PR to use the old solution conditionally depending on the platform.

stffabi:
Updated version pushed which defers the drag to `mousemove` for Linux.

leaanthony (approved):
Awesome! Absolute beast mode!
This will get easier when we split the JS to the different platforms.

Issue #2321, opened 2023-01-20
System Tray Memory Leak - Darwin


bdwyertech:
I have noticed a memory leak in the icon setting code for Darwin.  You can test this by having a goroutine rapidly toggle the icon and monitor memory usage.

I managed to fix this on (https://github.com/fyne-io/systray/commit/31f066d1602c490bbcad29db1b77c187ea793aab) ~~but I am not sure how to achieve similar here (not a C developer)~~.  Over there, ARC mode is enabled, and I simply wrapped some code in an autorelease pool.

I believe the memory leak exists somewhere in this function.
https://github.com/wailsapp/wails/blob/5ef32760f05f58daf8f05b15036562d380467b9d/v3/pkg/application/systemtray_darwin.go#L38-L53

Finally, this project is simply fantastic!  I love the fact that systray support is being added.  I'm currently using Fyne but would love to swap over to webview methodology for my app.

leaanthony:
Thanks for taking the time to report this and look at the entirely experimental v3 branch. I didn't enable arc as I thought it would be risky with CGO and the objc runtime knowing when it could free an object. Any thoughts on that?

bdwyertech:
> Thanks for taking the time to report this and look at the entirely experimental v3 branch. I didn't enable arc as I thought it would be risky with CGO and the objc runtime knowing when it could free an object. Any thoughts on that?

Explicit alloc/autorelease is easier to understand than the magic of more autorelease pools imho. Typically a func like imageFromBytes using alloc would return the object with autorelease set.  Because we call imageFromBytes from Go and feed it to setIcon, we can’t do that because it releases too soon.  Perhaps a future optimization may be to consolidate this setIcon logic into a single function instead of playing pass the pointer in CGO. This is an overoptimization now though imho 😊

PR #2322, opened 2023-01-20
Fix memory leak on Darwin when toggling system tray icon

bdwyertech:
alloc should always be paired with autorelease if not running in ARC mode.

Resolves https://github.com/wailsapp/wails/issues/2321

Related: https://github.com/fyne-io/systray/pull/34


leaanthony:
Thanks for this 🙏🙏🙏

leaanthony (approved):

PR #2324, opened 2023-01-21
Updating and Adding

raguay:
I updated the descriptions for EmailIt, ScriptBar, and EmailIt. I also added BullitenBoard.


leaanthony (approved):
Thanks for the updates 🙏

Issue #2332, opened 2023-01-26
Wrong models.ts generated with nested struct


lzjluzijie:
### Description

I have two structs defined in go:

```go
type User struct {
	ID    string `json:"id"`
	Name  string `json:"name"`
	Token string `json:"token"`
}

type Config struct {
	User   `json:"user"`
	RoomID string `json:"roomID"`
}
```

and wails generated `models.ts` like
```typescript
export namespace config {
	
	export class Config {
	    id: string;
	    name: string;
	    token: string;
	    roomID: string;
	
	    static createFrom(source: any = {}) {
	        return new Config(source);
	    }
	
	    constructor(source: any = {}) {
	        if ('string' === typeof source) source = JSON.parse(source);
	        this.id = source["id"];
	        this.name = source["name"];
	        this.token = source["token"];
	        this.roomID = source["roomID"];
	    }
	}
	export class User {
	    id: string;
	    name: string;
	    token: string;
	
	    static createFrom(source: any = {}) {
	        return new User(source);
	    }
	
	    constructor(source: any = {}) {
	        if ('string' === typeof source) source = JSON.parse(source);
	        this.id = source["id"];
	        this.name = source["name"];
	        this.token = source["token"];
	    }
	}

}
```
which is not nested, but in reality the object printed in console is nested:
![图片](https://user-images.githubusercontent.com/7992620/214955847-6d878fb9-5d83-4319-a221-1cddd8b64529.png)

### To Reproduce

Please see description.

### Expected behaviour

Ideally wails should generated `models.ts` with nested class, as written in Go, or the object should not be nested to match the current `models.ts`.

### System Details

```shell
DEB | Using go webview2loader
Wails CLI v2.3.1

Scanning system - Please wait (this may take a long time)...Done.

# System

OS           | Windows 10 Pro for Workstations
Version      | 2009 (Build: 22621)
ID           | 22H2
Go Version   | go1.19.4
Platform     | windows
Architecture | amd64

# Wails

Version | v2.3.1

# Dependencies

Dependency | Package Name | Status    | Version
WebView2   | N/A          | Installed | 109.0.1518.61
npm        | N/A          | Installed | 8.19.2
*upx       | N/A          | Installed | upx 4.0.1
*nsis      | N/A          | Available |
* - Optional Dependency

# Diagnosis

Your system is ready for Wails development!
Optional package(s) installation details:
  - nsis : More info at https://wails.io/docs/guides/windows-installer/

[...]

PR #2333, opened 2023-01-27
Possible fix for #2332. Handle anonymous fields with JSON tags as known.

sebastian0x62:


leaanthony:
Hey there 👋 can you please provide more context for this PR like what it changes and why? Future us will appreciate it 😅

sebastian0x62:
Tried to offer the intended behavior from the referenced issue. When bindings include a struct with an embedded anonymous other struct, generated .ts file and runtime TS type are not the same. Flat vs hierarchy.

This PR handles those fields as not anonym, when a struct tag for JSON marshalling is provided. This way the generated .ts file matches the runtime type.

sebastian0x62:
@leaanthony , I've read, for v3 the whole binding generation stuff is changed. In this case we also can delete this PR since a manual solution with changing such structs manually is still possible.

leaanthony:
@sebastian0x62 - closing as you suggested. Anon structs supported in v3 🎉

PR #2338, opened 2023-02-01
Fix #2280 - code block highlights

### LINKED ISSUE #2280: Unexpected code highlighting
### Have you read the Documentation Contribution Guidelines?

- [X] I have read the [Documentation Contribution Guidelines](https://wails.io/community-guide#ways-of-contributing).

### Description

Some of the code blocks in the docs have changed.  The highlights are specified with line numbers but those were not updated.

For example https://wails.io/docs/howdoesitwork#method-binding 

howdoesitwork

When the highlights were last set (commit dce5fd442fe2d68d0bea68f82f890f6bb9e5a568), the equivalent of lines 17 (`app := &App{}`) and 27 (`app,`) were highlighted.

After that commit, two more commits added lines to the code block without updating the highlights.

It looks like the wails.io is using Docusaurus.  It supports [special comments for highlighting lines](https://docusaurus.io/docs/markdown-features/code-blocks#highlighting-with-comments).

### Self-service

- [ ] I'd be willing to address this documentation request myself.

BuckeyeCoder:
Code reviewed the entire site's code blocks for those that use a highlight. Found 3 that needed adjusting. 

Only other change was from the Markdown linter which prefers underscores for italic text instead of asterisks for some reason. Either are allowed and assuming we all use the same linter I allowed the linter's opinionated change. 

Did not see an AUTHORS.md to add to; if you need that to maintain MIT let me know and we'll add it.

### DIFF
--- a/website/docs/guides/dynamic-assets.mdx
+++ b/website/docs/guides/dynamic-assets.mdx
@@ -11,7 +11,7 @@ By installing a custom AssetsHandler, you can serve your own assets using a cust
 
 In our example project, we will create a simple assets handler which will load files off disk:
 
-```go title=main.go {16-35,49}
+```go title=main.go {17-36,49}
 package main
 
 import (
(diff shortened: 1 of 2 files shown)


leaanthony:
> Did not see an AUTHORS.md to add to; if you need that to maintain MIT let me know and we'll add it.

We keep a list of contributors in the main project documentation, which is rendered here: https://wails.io/credits/#contributors
The source of this is: https://github.com/wailsapp/wails/blob/master/.all-contributorsrc

PR #2342, opened 2023-02-02
use appargs when generating bindings

Eriner:
This change makes binding generation respect `appargs` in `wails.json`.

--

I ran into a bug where wails was not generating JS bindings for my (exported) struct methods.

This was because the binding generation did not respect `appargs` and was silently failing during binding generation.

My project only executes wails if the `gui` argument is passed, as in: `./myprogram gui`. Prior to this change, when adding `"appargs": "gui"` to the `wails.json` file, everything worked *except* binding generation.

--

Also, sorry for the code formatting noise, my editor has a mind of its own.

### DIFF
--- a/v2/pkg/commands/bindings/bindings.go
+++ b/v2/pkg/commands/bindings/bindings.go
@@ -21,12 +21,12 @@ type Options struct {
 	GoModTidy        bool
 	TsPrefix         string
 	TsSuffix         string
+	AppArgs          string
 }
 
 // GenerateBindings generates bindings for the Wails project in the given ProjectDirectory.
 // If no project directory is given then the current working directory is used.
 func GenerateBindings(options Options) (string, error) {
-
 	filename, _ := lo.Coalesce(options.Filename, "wailsbindings")
 	if runtime.GOOS == "windows" {
 		filename += ".exe"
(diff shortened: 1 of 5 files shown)


v2/pkg/commands/bindings/bindings.go
```diff
@@ -67,7 +67,9 @@ func GenerateBindings(options Options) (string, error) {
 	env = shell.SetEnv(env, "tsprefix", options.TsPrefix)
 	env = shell.SetEnv(env, "tssuffix", options.TsSuffix)
 
-	stdout, stderr, err = shell.RunCommandWithEnv(env, workingDirectory, filename)
+	// TODO: split arguments by matching quotes
```
Eriner, line 70: This should be fixed prior to merge

Eriner:
Args are also not passed to the executables when platform-specific builds are created, however the fix for this is more involved than this simple PR.

floreks:
It would be great to get proper bindings generation when UI requires an argument to run. Any chance of getting this any time soon?

v2/pkg/commands/bindings/bindings.go
leaanthony: Who is fixing this? 😄

leaanthony:
@floreks - you happy to take ownership of this PR to move it forward?

leaanthony:
Closing for now. If anyone wants to take ownership of it please comment.

PR #2352, opened 2023-02-06
Patch 1

Nexus26404:
update function notifyListeners in events.js


Nexus26404:
newEventListenerList will delete wrong listener if it call splice method in forward for-loop.

leaanthony (approved):
Good catch! Thanks for this 🙏

Issue #2290, opened 2023-01-10
MacOS: wails build -o outputName ignored.


Snider:
### Description

When building an application on MacOS the -o arg is ignored, and the configuration value in the code is used.

### To Reproduce

```yaml
macos:
    runs-on: macos-latest
    steps:
      # Checkout code
      - uses: actions/checkout@v3
        with:
          repository: letheanVPN/desktop
          submodules: recursive
      - uses: dAppServer/wails-build-action@v2
        with:
          build-name: letheanTest
          build-platform: darwin/universal
          wails-build-webview2: "embed"
          wails-version: "v2.3.1"
          go-version: ^1.18
          sign: "false"
 ```

### Expected behaviour

The output name to be the same as provided with -o

### Screenshots

<img width="992" alt="image" src="https://user-images.githubusercontent.com/631881/211657355-a868deff-18d0-49f1-a769-25c75674dba5.png">

### System Details

```shell
https://github.com/dAppServer/wails-build-action/actions/runs/3885309312/jobs/6629024935
```

### Additional context

linux and windows will use the -o value; not sure if you build from Linux for macOS if it builds fine or not.

https://github.com/dAppServer/wails-build-action/actions/runs/3885309312
Workflow file to test the GHA: https://github.com/dAppServer/wails-build-action/blob/11-action-fails-with-wails-231-on-macos-due-to-ditto-error/.github/workflows/lethean-desktop.yml

Workflow file to test using GHA: https://github.com/dAppServer/wails-build-action/blob/main/.github/workflows/lethean-desktop.yml

gwynforthewyn:
I've been taking a look at this, and I'd like to develop the fix. Here's what I've figured out so far.

Building a macOS app comes in two parts: compiling the go binary, and packaging the .app. 

### go binary 
In v2/pkg/commands/build/build.go, in the execBuildApplication function, building the go binary has 2 logical branches for darwin; the if statement is on https://github.com/wailsapp/wails/blob/master/v2/pkg/commands/build/build.go#L266

If we are building a "universal" build (compiling for both arm64 and amd64 architectures), then on https://github.com/wailsapp/wails/blob/master/v2/pkg/commands/build/build.go#L311 the go compilation's OutputFile name is set to the value of the `-o` flag; if only one architecture is being built, this doesn't happen presently.

### macos packaging the .app

[...]

PR #2427, opened 2023-02-26
Set bundlename for single targets

leaanthony:
Another attempt at fixing the bundle output for Mac @Snider @gwynforthewyn

### DIFF
--- a/v2/cmd/wails/build.go
+++ b/v2/cmd/wails/build.go
@@ -181,7 +181,10 @@ func buildApplication(f *flags.Build) error {
 			macTargets := targets.Filter(func(platform string) bool {
 				return strings.HasPrefix(platform, "darwin")
 			})
-			if macTargets.Length() == 2 {
+			switch macTargets.Length() {
+			case 1:
+				buildOptions.BundleName = fmt.Sprintf("%s.app", desiredFilename)
+			case 2:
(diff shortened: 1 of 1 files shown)


gwynforthewyn:
@leaanthony , where are you getting the output from that's guiding this ongoing change?

leaanthony:
I reread the original issue which stated that the problem was in the bundle name, so I looked at the packaging code and sure enough, it's hardwired to the name of the project.

gwynforthewyn:
@leaanthony  ah! Thank you! I had asked in https://github.com/wailsapp/wails/issues/2290#issuecomment-1422036611 about addressing the CFBundleName,  but it was brought in as part of a conversation about whether I should also update the window title and then the bundle name got lost as a detail in that other conversation. 

Let me know if you want me to take a look through the whole plist file to see if there are any other pieces of metadata to update; I'm not 100% sure if you'd like that  stand with the next revision of the wails cli including a change to the build system

leaanthony:
If @Snider can confirm this is working then I'm happy to merge and release 2.4

stffabi:
Seems like specifying the `-o` flag doesn't have an effect anymore. Was this an intended change? I thought that we fixed that with another PR.

leaanthony:
The intent was the opposite :-) `-o` is ignored for the bundle name :(

Issue #1507, opened 2022-06-29
Disable the Webview2 security policy


jinpy666:
### Is your feature request related to a problem? Please describe.

For some crawlers, multi-account management of web projects is very meaningful ~ hope to support. Thank you so much for your team's work.

### Describe the solution you'd like

CORS requests need to be supported.
Read and write cookies and storages of any site.

misitebao:
You can follow this issue.

https://github.com/MicrosoftEdge/WebView2Feedback/issues/2558

yildirimcagri:
Hi, you can use the [--disable-web-security](https://stackoverflow.com/questions/3102819/disable-same-origin-policy-in-chrome), browser flag (at your own risk since this disables security policies such as cross origin restrictions :) ) to achieve what you are looking for. We don't intend to provide an API for this at the moment due to security risks involved. Hope this helps!

jinpy666:
Thank you~ @yildirimcagri @misitebao

KiddoV:
This would answer my #1375 as well, but I am still not sure how do you put flag ``--disable-web-security`` in ``wails``?

leaanthony:
Yeah, we could look at just adding a passthrough string to the options. I haven't really had a chance to consider the implications though

Qendolin:
The CORS policy really is a pain since reading local files is a common use-case for desktop apps and 'proxying' them via the asset handler is not always feasible and presents a security risk if not properly implemented. I hope this issue is resolved soon.

leaanthony:
It's scheduled. If you want to raise a PR sooner, feel free 👍

fynxiu:
``` golang
  os.Setenv("WEBVIEW2_ADDITIONAL_BROWSER_ARGUMENTS", "--disable-web-security")
```

stffabi:
> ```go
>   os.Setenv("WEBVIEW2_ADDITIONAL_BROWSER_ARGUMENTS", "--disable-web-security")
> ```

That won't work anymore since PR #1771 has been merged and will also not work with new native go-webview2loader (currently experimental) merged with PR #1974.

stffabi:
Yeah supporting this is now pretty straight forward with the new go-webview2loader.

From a personal view I'm still not 100% sure if this should be supported right out of the box. Because that easily opens an app for several security problems if one doesn't take care what will be loaded into the frontend.

haranobuhardo:
> It's scheduled. If you want to raise a PR sooner, feel free 👍

[...]

Issue #1350, opened 2022-04-25
Ignore the ssl certificate verify


chenxiao1990:
### Is your feature request related to a problem? Please describe.

in h5  i  wan't  request wss://192.168.xxx.xxx   , no Officially signed ssl certificate.

### Describe the solution you'd like

can set webview to Ignore  ssl certificate verify

leaanthony:
Please add platform you are using

chenxiao1990:
Platform:       windows

leaanthony:
I believe this will covered by whatever solution we use for #1507 👍

PR #2476, opened 2023-03-09
feat(linux): added WebviewTlsErrorsPolicyIgnore option to ignore TLS …

marcio199226:
Hi, I think this in somehow related to this issues: #1507 #1350 
With this PR we can ignore TLS related errors only for Linux at least for now, I'm not a C developer so it took me a while how to add this functionality.

Anyway self signed certs should not be a problem anymore, though this may be a potential security issue so use with caution and only if you know what are you doing as this disables tls check for all sites not only for localhost.

This is my first contribution to this repo so if I omit something apologize.

Cheers!

p.s if it will be merged I will also update the docs part


leaanthony:
Thanks for taking the time to do this - I appreciate it. I think we do need to have a wider discussion around how to handle potentially dangerous settings like this. I'd be keen to know what other projects are doing.

marcio199226:
Maybe we could show any kind of notification while building for prod env what do you think ? This could potentially speed up some POC development

leaanthony:
What are your thoughts on this @stffabi ?

leaanthony:
I'm going to close this for now and look at what this means for v3 👍

PR #2490, opened 2023-03-14
Bugfix - Parsing structs with reference breaks types generation

OlegGulevskyy:
Whenever parsing Go type to JS type, remove any `*` from the input, so that regexp can correctly match what is required (extract `valueType`, `valuePackage` etc).


zllovesuki:
> Thanks for jumping on this so quickly @OlegGulevskyy ! @zllovesuki does this PR change what you're seeing?

I supposed I will change my binding to not return a pointer for now until we have a better solution. This hack of a bugfix has a strong code smell

leaanthony:
This was addressed in a different PR 👍

PR #2496, opened 2023-03-16
[webview2] Fix error handling in CreateWebResourceResponse

stffabi:
The old error handling fails on WebView2Runtimes > 110.0.1587.69 and will show a "connection refused" page and reloads after a short time. After the reload everything wroked as expected again.


stffabi:
There are potentially also other locations that doing the wrong error handling and don't take into account the `HRESULT` of the native methods and only rely on the `GetLastError`.

leaanthony (approved):
Nice sleuthing! 🔍 👍

stffabi:
Thanks for approving.

PR #2497, opened 2023-03-16
Fix incorrect rendering of FAQ section header

boreq:


misitebao (approved):
Good fix.

boreq:
> Good fix.

I always make sure that my contributions are extremely important and valuable.

PR #2509, opened 2023-03-20
Use json.Marshal instead of strconv.Quote to correctly support unicode

joshbuddy:
I came across a bug where unicode values such as 🫣 would not get correctly escaped. This is due to using `strconv.Quote` which would encode this value as "\U0001fae3" which is not a valid js string. This would, in turn, get turned into "U0001fae3" in the frontend.

To correctly represent these values we need to use the new js format for multi-rune characters "\u{0001fae3}"[1], but I thought, instead of figuring out the correct quoting strategy, why not just use `json.Marshal` here instead. Certainly open to finding a different approach if you don't think this is a good idea.

Also decided to panic on error here as this error should never come up, and I saw similar panics in the code encountering things like this.

[1] https://exploringjs.com/es6/ch_unicode.html#sec_escape-sequences

### DIFF
--- a/v2/internal/frontend/desktop/darwin/frontend.go
+++ b/v2/internal/frontend/desktop/darwin/frontend.go
@@ -20,7 +20,6 @@ import (
 	"html/template"
 	"log"
 	"net/url"
-	"strconv"
 	"unsafe"
 
 	"github.com/wailsapp/wails/v2/pkg/assetserver"
@@ -337,7 +336,11 @@ func (f *Frontend) processMessage(message string) {
 }
 
 func (f *Frontend) Callback(message string) {
-	f.ExecJS(`window.wails.Callback(` + strconv.Quote(message) + `);`)
+	escaped, err := json.Marshal(message)
+	if err != nil {
+		panic(err)
+	}
+	f.ExecJS(`window.wails.Callback(` + string(escaped) + `);`)
 }
 
 func (f *Frontend) ExecJS(js string) {
--- a/v2/internal/frontend/desktop/linux/frontend.go
+++ b/v2/internal/frontend/desktop/linux/frontend.go
@@ -81,7 +81,6 @@ import (
 	"net/url"
 	"os"
 	"runtime"
-	"strconv"
 	"strings"
 	"text/template"
 	"unsafe"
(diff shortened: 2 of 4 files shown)


leaanthony:
Thanks for opening this and raising the issue. What testing have you done on this so far and on what platforms? Thanks!

joshbuddy:
@leaanthony I was only able to test this on Mac, but after making this change all the weird unicode issues I was having went away.

leaanthony:
@tmclane / @Lyimmi - Is this something that can be easily tested on Linux? 🙏

Lyimmi:
> @tmclane / @Lyimmi - Is this something that can be easily tested on Linux? pray

I checked the basic behavior, go side responds with a unicode smiley, the frontend prints it as an escaped string. If I apply @joshbuddy's changes the unicode is parsed properly on the frontend. 

I'm currently on an Ubuntu 22.04 LTS

**master:**
Screenshot from 2023-04-11 14-26-34

**PR applied:**
Screenshot from 2023-04-11 14-28-06

v2/internal/frontend/desktop/darwin/frontend.go
```diff
@@ -337,7 +336,11 @@ func (f *Frontend) processMessage(message string) {
 }
 
 func (f *Frontend) Callback(message string) {
-	f.ExecJS(`window.wails.Callback(` + strconv.Quote(message) + `);`)
+	escaped, err := json.Marshal(message)
+	if err != nil {
+		panic(err)
```
tmclane, line 341: Can we make it not panic here?  That would look like a crash for a user and be really hard to locate.

tmclane: Likewise for the other ones.

tmclane (commented):
I'll test it in a bit.  Looks like @Lyimmi already did though.

v2/internal/frontend/desktop/darwin/frontend.go
leaanthony: What would you do though? If your callbacks fail then you have a pretty catastrophic error. I'm trying to think of an alternative

joshbuddy: As far as I can tell, a panic is appropriate here because `json.Marshal` should never return an error in response to marshaling a string, so if it did that, I would consider that to be unspecified behavior.

On a separate note, I notice that json.Marshall will do a certain amount of html escaping on a string. I can't think how it would be a problem, but worth noticing. See: https://pkg.go.dev/encoding/json#Marshal

> String values encode as JSON strings coerced to valid UTF-8, replacing invalid bytes with the Unicode replacement rune. So that the JSON will be safe to embed inside HTML <script> tags, the string is encoded using HTMLEscape, which replaces "<", ">", "&", U+2028, and U+2029 are escaped to "\u003c","\u003e", "\u0026", "\u2028", and "\u2029". This replacement can be disabled when using an Encoder, by calling SetEscapeHTML(false).

joshbuddy:
@leaanthony done

leaanthony (approved):
Thanks for doing this 🙏

PR #2520, opened 2023-03-22
docs: optimize code block display style

misitebao:
Optimize code block display style, easy to display and translate.


leaanthony:
Agree. Much better!

PR #2527, opened 2023-03-25
Fix dates in changelog

antimatter96:
Please let me know if the dates for other languages need to be made.
Keep up the good work 👍🏻 🚀


leaanthony:
Oh nice catch!

PR #2561, opened 2023-04-06
update wml events to be parseable

tmclane:
When testing the `v3/examples/plain` application I noticed that the `clicking` was failing to be parsed on Mac and on the Linux port as well.

I was getting the following message:

```
[INFO] Wails: Window: 1, Method: GET Request: wails://wails/wails/runtime?method=events.Emit&args=%22clicked%22
[MessageProcessor] Error: Error parsing event: json: cannot unmarshal string into Go value of type application.CustomEvent
```

This change updates the wml.js to send an object with the `name` attribute equal to the `eventType` passed to it's `sendEvent` function.
This at least allows it to parse on the backend but I am not sure this is what was intended.


leaanthony:
Oh nice! Did the wml example not work? I must have changed it....

leaanthony:
I've added in a bunch of changes. The original PR was ok in itself but I found that `CustomEvent` was actually a standard JS Object, so renamed it. Of course, to be consistent it needs to be renamed in Go too and boom! a bunch of files have changed. This totally was an issue so thanks for picking it up 👍 If you're ok with the changes I just pushed then we can merge it.

tmclane (commented):
I can't approve this since I started the PR originally.  :+1:

tmclane:
I have no problem of course with the changes you want.  I figured there was something more that was missing other than it being a string only instead of an object.

leaanthony (approved):

tmclane:
@leaanthony Just rebased and force-pushed on top of `master`.  It is ready to be merged if you would do the honor.

leaanthony (approved):

leaanthony:
Thanks for getting us here @tmclane 🙏

PR #2562, opened 2023-04-06
refactor menu handling

tmclane:
- refactor 'processMenu' out of 'macosMenu' into 'Menu'
- update menuImpl interface to ensure methods required to create menus are present
- implement menuImpl methods in macosMenu

This refactoring was done to avoid having to reimplement the overall 'processMenu' code for each platform.

### DIFF
--- a/v3/pkg/application/application_darwin.go
+++ b/v3/pkg/application/application_darwin.go
@@ -170,7 +170,7 @@ func (m *macosApp) setApplicationMenu(menu *Menu) {
 	}
 	menu.Update()
 
-	// Convert impl to macosMenu object
+	// Convert impl to macosMenu object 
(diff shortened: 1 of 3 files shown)


leaanthony:
I'm unsure if the steps required for both Linux and Windows follow this same logic. Keen to hear your feedback on that

tmclane:
I was doing this refactor in order to allow for the Linux menu creation to work the same way.   I was going to base the Linux version on this actually.  I presume the same menu structure would hold wouldn't it?

Let me experiment with the Linux implementation on top of this and that will help decide if I am correct. 
This is a draft for that reason.

tmclane:
FYI: I'm going to throw this away most likely.

PR #2578, opened 2023-04-12
Update zypper.go

wgjtyu:
the name of `libwebkit` in opensuse-tumbleweed is `webkit2gtk3-soup2-devel`


v2/internal/system/packagemanager/zypper.go
```diff
@@ -33,7 +33,7 @@ func (z *Zypper) Packages() packagemap {
 			{Name: "gtk3-devel", SystemPackage: true, Library: true},
 		},
 		"libwebkit": []*Package{
-			{Name: "webkit2gtk3-devel", SystemPackage: true, Library: true},
+			{Name: "webkit2gtk3-soup2-devel", SystemPackage: true, Library: true},
```
leaanthony, line 36: Could you please try adding this as a second entry in the slice rather than replace the current one? 👍

wgjtyu: ok, I will create another PR later.

PR #2579, opened 2023-04-12
Bugfix/remove miniaturize

tmclane:
Removes the `minimize` call from the `webviewWindowImpl` and the code which it was calling.
This appeared to be duplicate logic to the `minimise` call and was therefore redundant.


leaanthony:
Nice one @tmclane 🎉

Issue #2582, opened 2023-04-12
macOS: Window management shortcuts like Cmd+Ctrl+F don't work


mholt:
### Description

Wails windows, both `Frameless: true` and `Frameless: false`, are not able to be controlled by macOS utilities like Magnet (which aligns/tiles windows) and keyboard shortcuts like Cmd+Ctrl+F which full-screens an application.

### To Reproduce

Run (presumably) any Wails app, and try to fullscreen with Cmd+Ctrl+F. Note that this shortcut does work on other applications.

If you have Magnet or other tiling utilities installed, try using those to control the window.

### Expected behaviour

Application should fullscreen. Utilities like Magnet should be able to resize and move the window.

### Attempted Fixes

Tried toggling Frameless, playing with Fullscreen options, and other options.App settings, but no luck.

### System Details

```shell
Wails CLI v2.4.1

Scanning system - Please wait (this may take a long time)...Done.

# System

OS           | MacOS   
Version      | 13.2.1  
ID           | 22D68   
Go Version   | go1.20.1
Platform     | darwin  
Architecture | arm64   

# Wails

Version | v2.4.1

# Dependencies

Dependency                | Package Name | Status    | Version
Xcode command line tools  | N/A          | Installed | 2396   
npm                       | N/A          | Installed | 9.4.0  
*Xcode                    | N/A          | Available |        
*upx                      | N/A          | Available |        
*nsis                     | N/A          | Available |        
* - Optional Dependency

# Diagnosis

Your system is ready for Wails development!
Optional package(s) installation details: 
  - Xcode: Available at https://apps.apple.com/us/app/xcode/id497799835
  - upx : Available at https://upx.github.io/
  - nsis : More info at https://wails.io/docs/guides/windows-installer/
```

### Additional context

Thanks for working on a really hard problem (Wails in general I mean) :blush:

stffabi:
> If you have Magnet or other tiling utilities installed, try using those to control the window.

https://rectangleapp.com seems to work fine with a Wails application, the tiling features worked all as expected.

> Cmd+Ctrl+F

Does not work currently and probably needs an internal handler and do a fullscreen.

stffabi:
Can confirm we do need to add a menu item for `Cmd+Ctrl+F` to work, same for `Cmd+M` to support minimize.

But have no idea what's needed for Magnet.

stffabi:
There are also other missing shortcuts
[...]

PR #2586, opened 2023-04-12
[v2, darwin] Add some missing default shortcuts

stffabi:
See also #2582


mholt:
From my brief testing, this works! Thank you!

PR #2587, opened 2023-04-12
Feature/explicit callbacks

tmclane:
Functional updates:

- update WebviewWindow to implement functions for each of the callback types currently in use
  -  CallError
  - CallResponse
  - DialogError
  - DialogResponse
- update MessageProcessor to utilize the above functions on the window
- move responsibility of creating the javascript for the callback to the webview window functions
- ensure 'result' data is JSON marshallable or panic

Non-Functional updates:

- Sorted the `webviewImpl` interface functions

Note: My original intent with this changeset was to add `call.Context|call.Response` type interfaces to the function calls as well.  I decided not to do that at this time.
This simplifies what i was doing to enable the `websocket server plugin`

### DIFF
--- a/v3/pkg/application/messageprocessor_call.go
+++ b/v3/pkg/application/messageprocessor_call.go
@@ -4,19 +4,16 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
-	"strconv"
 )
 
 func (m *MessageProcessor) callErrorCallback(window *WebviewWindow, message string, callID *string, err error) {
 	errorMsg := fmt.Sprintf(message, err)
 	m.Error(errorMsg)
-	msg := "_wails.callErrorCallback('" + *callID + "', " + strconv.Quote(errorMsg) + ");"
-	window.ExecJS(msg)
+	window.CallError(callID, errorMsg)
 }
 
 func (m *MessageProcessor) callCallback(window *WebviewWindow, callID *string, result string, isJSON bool) {
-	msg := fmt.Sprintf("_wails.callCallback('%s', %s, %v);", *callID, strconv.Quote(result), isJSON)
-	window.ExecJS(msg)
+	window.CallResponse(callID, result)
 }
 
 func (m *MessageProcessor) processCallMethod(method string, rw http.ResponseWriter, _ *http.Request, window *WebviewWindow, params QueryParams) {
--- a/v3/pkg/application/messageprocessor_dialog.go
+++ b/v3/pkg/application/messageprocessor_dialog.go
@@ -5,19 +5,16 @@ import (
 	"fmt"
 	"net/http"
 	"runtime"
-	"strconv"
 )
 
 func (m *MessageProcessor) dialogErrorCallback(window *WebviewWindow, message string, dialogID *string, err error) {
 	errorMsg := fmt.Sprintf(message, err)
 	m.Error(errorMsg)
-	msg := "_wails.dialogErrorCallback('" + *dialogID + "', " + strconv.Quote(errorMsg) + ");"
-	window.ExecJS(msg)
+	window.DialogError(dialogID, errorMsg)
 }
 
 func (m *MessageProcessor) dialogCallback(window *WebviewWindow, dialogID *string, result string, isJSON bool) {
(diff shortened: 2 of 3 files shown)


v3/pkg/application/messageprocessor_dialog.go
```diff
@@ -5,19 +5,16 @@ import (
...
 func (m *MessageProcessor) dialogErrorCallback(window *WebviewWindow, message string, dialogID *string, err error) {
 	errorMsg := fmt.Sprintf(message, err)
 	m.Error(errorMsg)
-	msg := "_wails.dialogErrorCallback('" + *dialogID + "', " + strconv.Quote(errorMsg) + ");"
-	window.ExecJS(msg)
+	window.DialogError(dialogID, errorMsg)
 }
 
 func (m *MessageProcessor) dialogCallback(window *WebviewWindow, dialogID *string, result string, isJSON bool) {
-	msg := fmt.Sprintf("_wails.dialogCallback('%s', %s, %v);", *dialogID, strconv.Quote(result), isJSON)
```
leaanthony, line 19: If we remove `isJSON` here, do multiple file select dialogs still work?

v3/pkg/application/messageprocessor_call.go
```diff
@@ -4,19 +4,16 @@ import (
...
 func (m *MessageProcessor) callErrorCallback(window *WebviewWindow, message string, callID *string, err error) {
 	errorMsg := fmt.Sprintf(message, err)
 	m.Error(errorMsg)
-	msg := "_wails.callErrorCallback('" + *callID + "', " + strconv.Quote(errorMsg) + ");"
-	window.ExecJS(msg)
+	window.CallError(callID, errorMsg)
 }
 
 func (m *MessageProcessor) callCallback(window *WebviewWindow, callID *string, result string, isJSON bool) {
-	msg := fmt.Sprintf("_wails.callCallback('%s', %s, %v);", *callID, strconv.Quote(result), isJSON)
```
leaanthony, line 18: I think if we can resolve the `isJSON` question then we could get rid of this method and call it on the window directly.

v3/pkg/application/messageprocessor_dialog.go
leaanthony: Perhaps this method could always just decode the payload? I believe a string is a valid json payload. https://github.com/wailsapp/wails/blob/master/v3/internal/runtime/desktop/dialogs.js#L35-L45

v3/pkg/application/messageprocessor_call.go
tmclane: That's a good point.  Everywhere the call existed it was passing a `true` already so I think might should create a `type` that enforces the object passed in to have a `ToJSON` function implemented in order to be acceptable. 
What do you think of that?

v3/pkg/application/messageprocessor_dialog.go
tmclane: Let me check the `file dialog` stuff.  I have to be honest in that I didn't test that since I haven't gotten that reimplemented in Linux yet.  That is the next thing on my todo list after the menus.

tmclane: I'll check it on the `mac`.

leaanthony: https://github.com/wailsapp/wails/pull/2587/files#diff-5f2cc1ceaf1ea4e502ed0aca10ee2858b1f6978b278b2200e27c5d8746e9a8deL66 uses `false`

tmclane: Just tested multiple dialogs on Mac.  Works fine without any other changes so far.

v3/pkg/application/messageprocessor_call.go
leaanthony: I think we should be able to get away with parsing it as json and if it fails just return the raw value. I'll try looking into this tomorrow 👍

tmclane: Just as an aside. The Dialogs work just fine with this code.  They always return a string, so it seems it's OK. Although I think we should update such that you can't possibly put in a bad value.

tmclane:
This functionality was already merged as part of a related branch.

PR #2593, opened 2023-04-14
Update zypper.go (opensuse-tumbleweed)

wgjtyu:
- Add package(webkit2gtk3-soup2-devel) for `libwebkit`.
- Add package(pkgconf-pkg-config) for `pkg-config`.
- Change environment variable before running command to query information of the package.

### DIFF
--- a/v2/internal/system/packagemanager/zypper.go
+++ b/v2/internal/system/packagemanager/zypper.go
@@ -33,6 +33,7 @@ func (z *Zypper) Packages() packagemap {
 			{Name: "gtk3-devel", SystemPackage: true, Library: true},
 		},
 		"libwebkit": []*Package{
+			{Name: "webkit2gtk3-soup2-devel", SystemPackage: true, Library: true},
 			{Name: "webkit2gtk3-devel", SystemPackage: true, Library: true},
 		},
 		"gcc": []*Package{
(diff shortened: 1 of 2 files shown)


v2/internal/system/packagemanager/zypper.go
```diff
@@ -60,7 +62,9 @@ func (z *Zypper) PackageInstalled(pkg *Package) (bool, error) {
 	if pkg.SystemPackage == false {
 		return false, nil
 	}
-	stdout, _, err := shell.RunCommand(".", "zypper", "info", pkg.Name)
+	var env []string
+	env = shell.SetEnv(env, "LANGUAGE", "en_US.utf-8")
```
leaanthony, line 66: Do we know this isn't going to break non US English installations?

wgjtyu: Yes, my OS was Chinese, if we don't set `LANGUAGE` env variable here, the `zypper info pkgName` command will return result in Chinese, so `wails doctor` will nott work properly here.

leaanthony:
Please update the changelog in `website/src/pages/changelog.mdx` 🙏

wgjtyu:
OK, updated.

leaanthony (approved):

leaanthony (approved):

Issue #2548, opened 2023-04-01
Type safe source for generated typescript models


dominictwlee:
### Is your feature request related to a problem? Please describe.

Currently, we can pass any plain ol' object  to the constructor of a model class.  I think it'd be nice to be able to pass in a typesafe object that matches the types generated for the class.

### Describe the solution you'd like

```
type UserSourceInput = {
  name: string;
  description: string;
  id: string;
}

function validUserSourceInput(source: unknown): source is UserSourceInput {
  // validate source here, check keys and value types?
}

export class User {
	  name: string;
	  description: string;
	  id: string;

         // instead of any, we narrow the type here
	  static createFrom(source: UserSourceInput | string) {
	      return new User(source);
	  }

	  constructor(source: UserSourceInput | string) {
	      if ('string' === typeof source) {
                  const parsedSource = JSON.parse(source);

                  // type guard
                 if (validUserSourceInput(parsedSource)) {
                      source = parsedSource
                  }
               }
	      this.name = source["name"];
	      this.description = source["description"];
	      this.id = source["id"];
	  }
}
```

### Describe alternatives you've considered

My current workaround is something like:
```
export type ModelInput<T> = Omit<T, "convertValues">;
```

leaanthony:
Closing this for now as it's been discussed on discord.

PR #2600, opened 2023-04-16
[v3] add typesafe model source to template

dominictwlee:
As mentioned in #2548, this is an attempt to narrow source type to match model fields.

@leaanthony let me know what you think.


leaanthony:
Thanks for looking at this. So this is an attempt to allow a Partial to populate an object rather than a string? I think I'd need someone more familiar with TS to review this. I'll ask on the discord server 👍

plihelix:
The problem is mostly because of type limits when going from go -> json -> ts.


From what I can tell, if the generated models encounter a type that isn't generated it will parse the JSON just fine (into any type) but will fail to convert due to the 'null' type being passed into the conversion function. 

Rather than a null, this should stay as any (#1) and the 'asMap' flag should be set to true #2 so that it is converted to a map instead of discarded.

```ts
// example generated models.ts
export class Person {
	name: string;
	// Go type: iso8601    <- Correctly identified type not  defined and generated
	born: any;              // <- 'any' type for JSON parsing
        ...
	
	constructor(source: any = {}) {
	    if ('string' === typeof source) source = JSON.parse(source);  <- Parse JSON input
	    this.name = source["name"];
	    this.born = this.convertValues(source["born"], null);                <- #1 defaults to null
	    ...
	}
	
        // Generic conversion function that is used in the constructor.
        // Given the source any and the target type convert the JSON into
        // the appropriate TS type. 
        // Optionally return a k:v pair map for unknown data.
        convertValues(a: any, classs: any, asMap: boolean = false): any {
	    if (!a) {
	        return a;
            }
	    if (a.slice) {
	        return (a as any[]).map(elem => this.convertValues(elem, classs));
	    } else if ("object" === typeof a) {
	        if (asMap) {                                                                          <- #2 returns a pair map of the json
	            for (const key of Object.keys(a)) {
	                a[key] = new classs(a[key]);
	            }
	            return a;
	        }
	        return new classs(a);                                                 <- pushing a value into a null gets a null
	    }
	    return a;
	}
}
```

dominictwlee:
> Thanks for looking at this. So this is an attempt to allow a Partial to populate an object rather than a string? I think I'd need someone more familiar with TS to review this. I'll ask on the discord server 👍

Yeah. This is a case of making things a bit more convenient when initializing a model.

i.e.
```
const person = new Person({
  name: 'hello'; // typed as name: string, typescript won't show an error
   buddy: 123 // doesn't exist in source type, will show a TS error
})
```

leaanthony:
Necromancing this as this is something we are looking at in the current v3 alpha. IF you'd like to join in on the discussion, feel free to join us [here](https://discord.com/channels/1042734330029547630/1184977285225398393).

Issue #2590, opened 2023-04-13
Cookies cannot be set on macos


helloxz:
### Description

I am on the macos platform. I can normally set cookies in dev mode, but setting cookies after build is invalid.

I did a comparison check with debug mode and found this to be the case.

### To Reproduce

run `wails dev` ok

```js
document.cookie = "username=John Doe; expires=" + new Date(new Date().getTime() + (24 * 60 * 60 * 1000)).toUTCString() + "; path=/";
```

run `wails build -debug` failed

```js
document.cookie = "username=John Doe; expires=" + new Date(new Date().getTime() + (24 * 60 * 60 * 1000)).toUTCString() + "; path=/";
```

### Expected behaviour

Cookies can be set correctly and effectively

### System Details

```shell
Wails CLI v2.4.1

Scanning system - Please wait (this may take a long time)...Done.

# System

OS           | MacOS   
Version      | 13.3.1  
ID           | 22E261  
Go Version   | go1.20.3
Platform     | darwin  
Architecture | amd64   

# Wails

Version | v2.4.1

# Dependencies

Dependency                | Package Name | Status    | Version  
Xcode command line tools  | N/A          | Installed | 2397     
npm                       | N/A          | Installed | 9.5.0    
*Xcode                    | N/A          | Available |          
*upx                      | N/A          | Installed | upx 4.0.2
*nsis                     | N/A          | Available |          
* - Optional Dependency

# Diagnosis

Your system is ready for Wails development!
Optional package(s) installation details: 
  - Xcode: Available at https://apps.apple.com/us/app/xcode/id497799835
  - nsis : More info at https://wails.io/docs/guides/windows-installer/

 ♥   If Wails is useful to you or your company, please consider sponsoring the project:
https://github.com/sponsors/leaanthony
```

leaanthony:
Is your expectation that webviews support cookies? Do they?

helloxz:
* `wails dev`(support cookies)
* `wails build`(no suppot cookies)

I hope the wails build can also support setting cookies

helloxz:
Excuse me, have you reproduced this BUG?

evorts:
In my opinion, having no cookie support does not posed a critical issue. Since it can be handled by either LocalStorage (which hadn't been an issue for me) or perhaps by the `golang`/`backend` side of the wails app. Cookie imho much more appropriate for web development which run on full fledge browser. Again, this is just my opinion

misitebao:
[...]

Issue #2523, opened 2023-03-23
[linux] Weird behaviour with HTTPS with wails build command


marcio199226:
### Description

Hi I have a backend here: https://server.mchat.ovh/ with TLS enabled and the certificates generated by lets encrypt certbot.
When I running my app with **wails dev** everything goes ok, when running builded app I encounter this error message in webview console **"Failed to load resource: WebKit encountered an internal error"**

I think that the problem is caused by Origin header that in case of **wails dev** it's set to **localhost:34115** and in case of wails build it's set to **wails://wails** and this custom protocol cause some troubles.

I can't figure out how can I fix it.
I even tried some local http proxy like this:
```
package main

import (
	"embed"
	"log"
	"net/http"
	"net/http/httputil"
	"net/url"

	"github.com/wailsapp/wails/v2"
	"github.com/wailsapp/wails/v2/pkg/logger"
	"github.com/wailsapp/wails/v2/pkg/options"
	"github.com/wailsapp/wails/v2/pkg/options/linux"
	"github.com/wailsapp/wails/v2/pkg/options/mac"
	"github.com/wailsapp/wails/v2/pkg/options/windows"
)

//go:embed frontend/dist
var assets embed.FS

func handler(w http.ResponseWriter, r *http.Request) {
	localProxyUrl, _ := url.Parse("https://server.mchat.ovh/")
	w.Header().Set("Access-Control-Allow-Methods", "POST, GET, OPTIONS, PUT, DELETE")
	w.Header().Set("Access-Control-Allow-Headers", "Accept, Content-Type, Content-Length, Accept-Encoding, X-CSRF-Token, Authorization")

	proxy := httputil.NewSingleHostReverseProxy(localProxyUrl)

	r.URL.Host = localProxyUrl.Host
	r.URL.Scheme = localProxyUrl.Scheme
	r.Header.Set("X-Forwarded-Host", r.Header.Get("Host"))
	r.Host = localProxyUrl.Host

	proxy.ServeHTTP(w, r)
}

func main() {
	// Create an instance of the app structure
	app := NewApp()

	go func() {
		http.HandleFunc("/", handler)

		log.Println("Serving on localhost:8080")
		log.Fatal(http.ListenAndServe(":8080", nil))
	}()

	// Create application with options
	err := wails.Run(&options.App{
		Title:             "mchat-desktop",
		Width:             1024,
		Height:            768,
		MinWidth:          1024,
		MinHeight:         768,
		DisableResize:     false,
[...]

Issue #2090, opened 2022-11-14
[linux] Can play audio in wails dev but not in wails build


cvilsmeier:
### Description

My wails app must play audio when user clicks a 'play' button. I'm using the following code for setup:

    import alertMp3 from './assets/sounds/alert.mp3';
    let audio = new Audio(alertMp3);

and later, to actually starting the audio:

    audio.play();

The thing is working perfectly well when I use

    wails dev

but the sound could not be loaded (and therefore not played) when using

    wails build

I've created a small sample project for reproducing the error: https://github.com/cvilsmeier/pino

### To Reproduce

1. git clone https://github.com/cvilsmeier/pino
2. cd pino
3. wails dev
4. wails build && build/bin/pino

### Expected behaviour

A sound must be playing both in `wails dev` and in `wails build`

### Attempted Fixes

I've tried it with data URLs, and that works:
Instead of my setup code 

     import alertMp3 from './assets/sounds/alert.mp3';
     let audio = new Audio(alertMp3);

I'm using this code:

    let audio = new Audio("data:audio/mpeg;base64, //uQxAAAAAAAAAA...(long data string)");

Then the audio is loaded and played correctly.

### System Details

```shell
Wails CLI v2.2.0

    Scanning system - Please wait (this may take a long time)...Done.

    System
    ------
    OS:		Debian GNU/Linux
    Version: 	11
    ID:		debian
    Go Version:	go1.19.3
    Platform:	linux
    Architecture:	amd64

    Wails
    ------
    Version: 		v2.2.0
    Package Manager: 	apt

    Dependency	Package Name		Status		Version
    ----------	------------		------		-------
    *docker 	docker.io 		Available 	20.10.5+dfsg1-1+deb11u2
    gcc 		build-essential 	Installed 	12.9
    libgtk-3 	libgtk-3-dev 		Installed 	3.24.24-4+deb11u2
    libwebkit 	libwebkit2gtk-4.0-dev 	Installed 	2.38.2-1~deb11u1
    npm 		npm 			Installed 	8.19.2
    *nsis 		nsis 			Available 	3.06.1-1
    pkg-config 	pkg-config 		Installed 	0.29.2

    * - Optional Dependency

    Diagnosis
    ---------
    Your system is ready for Wails development!
    Optional package(s) installation details: 
    - docker: sudo apt install docker.io
    - nsis: sudo apt install nsis



    If Wails is useful to you or your company, please consider sponsoring the project:
[...]

Issue #2026, opened 2022-10-31
error Embed twitch


longjin1991:
### Description

I embed twitch video to app wails, code:

<div id="twitch-embed">

<script src="[https://player.twitch.tv/js/embed/v1.js"></script>

<script type="text/javascript">
new Twitch.Player("twitch-embed", {
video: "1621322762",

parent: ['wails.localhost', window.location.host],
});
</script>

### To Reproduce

when run: wails dev.
it normal operation, show video and play it.
when run: wails build
it successful, but video twitch not show, not view.
console in devtool error:
Refused to frame 'https://player.twitch.tv/' because an ancestor violates the following Content Security Policy directive: "frame-ancestors https://wails.localhost/".

### Expected behaviour

i want when run wails build, twitch video show and play normal.

### Screenshots

https://imgur.com/a/T1ZEVYj

### System Details

```shell
System
------
OS:             Windows 10 Pro for Workstations
Version:        2009 (Build: 19045)
ID:             22H2
Go Version:     go1.19.2
Platform:       windows
Architecture:   amd64

Wails
------
Version:        v2.1.0

Dependency      Package Name    Status          Version
----------      ------------    ------          -------
WebView2        N/A             Installed       107.0.1418.24
npm             N/A             Installed       8.19.2
*upx            N/A             Available
*nsis           N/A             Available

* - Optional Dependency

Diagnosis
---------
Your system is ready for Wails development!
Optional package(s) installation details:
  - upx : Available at https://upx.github.io/
  - nsis : Available at https://nsis.sourceforge.io/Download
```

### Additional context

i think the problem comes from the url: https://wails.localhost/
how to change url start:https://wails.localhost/ to example: http:/localhost:1234/
thank you.

stffabi:
Thanks for using Wails and reporting this issue.

Unfortunately I don’t see a way we could support that in a production build. On Linux and macOS we have to use a custom scheme to serve the embedded assets, which is “wails://wails/“. On Windows this was not possible in the early days, so as a workaround Wails is currently using “http://wails.localhost/”, but since this is now supported by WebView2 we are going to switch that probably soon to have a consistent behaviour.

There are several discussions on the twitch dev forum about developers that can’t use the player in a iOS WkWebView app, where one also needs to use a custom scheme: https://discuss.dev.twitch.tv/t/embedded-player-within-ios-web-based-mobile-app-e-g-ionic-not-working/36227

PR #2610, opened 2023-04-18
[v2, dev] Use custom schemes for in-app dev mode

stffabi:
This fixes some long-standing inconsistencies between dev mode builds and production builds but is a breaking change.  Dev mode uses custom scheme for Vite versions >= 3.0.0 and for older it still behaves in the old way.

Relates to
- #2590
- #2523
- #2090
- #2026
- #1568

### DIFF
--- a/v2/cmd/wails/internal/dev/dev.go
+++ b/v2/cmd/wails/internal/dev/dev.go
@@ -22,6 +22,7 @@ import (
 	"github.com/wailsapp/wails/v2/cmd/wails/flags"
 	"github.com/wailsapp/wails/v2/cmd/wails/internal/gomod"
 	"github.com/wailsapp/wails/v2/cmd/wails/internal/logutils"
+	"golang.org/x/mod/semver"
 
 	"github.com/wailsapp/wails/v2/pkg/commands/buildtags"
 
@@ -34,6 +35,10 @@ import (
 	"github.com/wailsapp/wails/v2/internal/process"
 	"github.com/wailsapp/wails/v2/pkg/clilogger"
 	"github.com/wailsapp/wails/v2/pkg/commands/build"
+)
+
+const (
+	viteMinVersion = "v3.0.0"
 )
 
 func sliceToMap(input []string) map[string]struct{} {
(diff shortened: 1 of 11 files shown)


leaanthony:
Do you think using the old URL for Vite < 3 would be reasonable? It'd mean it wouldn't break compatibility but we could spam the terminal with all the benefits of upgrading to v3?

stffabi:
> This is great! Perhaps we should link the tickets we believe will be fixed by this?

Yeah that's a good idea.

stffabi:
> Do you think using the old URL for Vite < 3 would be reasonable? It'd mean it wouldn't break compatibility but we could spam the terminal with all the benefits of upgrading to v3?

This would need some more if-cases so theoretically it's feasible. I'm not sure if that gives a better experience for the developer. In the current version of the PR it's clear that when upgrading to the next release there's a breaking change and the user needs to update vite. Otherwise it would still work as in older releases and would then introduce the breaking change just by the means of updating vite to a newer version.
As an end user of Wails I would personally prefer to have specific version that introduces the breaking change, and not by just updating Vite. 
Another solution would to add a config value, but I don't know if that's really better. Especially we are only fixing an inconsistency in between dev/prod mode and all people that didn't encounter such an inconsistency by using a feature in dev that doesn't exist in prod should not be affected by this breaking change.

stffabi:
I've update the initial comment to list issues that I've found relating to such inconsistencies.

stffabi:
@leaanthony I've marked the PR ready for review, let me know how you would like to proceed with it.

leaanthony:
I think if it's not a huge problem, we should use the default behaviour if vite < 3 but issue a large warning about it. I think this will be a small minority because v3 has been standard in v2 for a while, however if their projects just fail to work then that would be a bad experience. There may be some weird reason they can't upgrade or perhaps they are using a remote template. I'm happy to do that work if you're done with it 😅

stffabi:
Ok, will do that. Unfortunately we still had some old vite v2 in our templates. 😞

stffabi:
@leaanthony PR has been updated

stffabi:
> I remember from the first commit (can't get to it after force push) that `ViteServerURLChan` was used. According to my IDE, it looks like it's only sent a value but the value isn't read. Am I reading this wrong?

Didn't change anything there, it's still being read here `v2/cmd/wails/internal/dev/dev.go` line 235

leaanthony:
That is so weird! I wonder if it was a gui issue with a small window as I can see it just fine on desktop... 🤯

leaanthony (approved):
Thanks so much for adding in the legacy support 🙏 This feature should help a lot.

stffabi:
You're welcome. I hope this will get us rid of all those "working in development, not working in production" tickets.

PR #2618, opened 2023-04-21
[darwin] add getPrimaryScreen/getScreens to impl

tmclane:
Adding in the missing impl methods for Darwin


leaanthony:
Ooops! Thanks man!

PR #2637, opened 2023-05-01
feat(cli): use Spinner instead of Print

misitebao:
- [x] Use `Spinner` instead of `Print` to optimize user experience.

<details>
  <summary>Preview</summary>
  
wails-doctor

</details>


leaanthony (approved):
This is great! Thanks!

Issue #2640, opened 2023-05-02
Proposal: Remove the `v1` content from the default branch


misitebao:
Since `v1` has been suspended for maintenance, the `v1` related content has been moved to the `v1` branch.

To do this just do:
- Set the `v1` branch as a protected branch, this is to prevent the branch from being accidentally deleted.
- Check and merge the PR corresponding to this Issue.

PR #2641, opened 2023-05-02
chore: remove the `v1` content from default branch

misitebao:
Resolve: https://github.com/wailsapp/wails/issues/2640

After merging, please set the `v1` branch as a protected branch.


leaanthony:
What is the impact to projects using v1? For instance Wally is still using it. Extract from `go.mod`:

`github.com/wailsapp/wails v1.16.7`

misitebao:
> What is the impact to projects using v1? For instance Wally is still using it. Extract from `go.mod`:
> 
> `github.com/wailsapp/wails v1.16.7`
> 
> I'm thinking the tags should still be there...?

I don't think this will affect it. When go imports a package, it first looks for it based on the tag, which has nothing to do with the branch.

leaanthony:
If @stffabi is ok with this then I think we're good. I can't think of anything right now that might be an issue.

stffabi:
I'm also fine doing this. Should also make it clear `v1` is EoL. Thanks @misitebao for doing this.

misitebao:
> Should also make it clear `v1` is EoL.

This can be explained in README or documentation, because `go module` does not have a command like `npm deprecate`. 🤣

stffabi:
> > Should also make it clear `v1` is EoL.
> 
> @stffabi @leaanthony This can be explained in README or documentation, because `go module` does not have a command like `npm deprecate`. 🤣

I think I didn't express myself right. I meant it's already very clear `v1` is unmaintained, but with removing the code from master this also makes it somehow final. So everything is fine 😄

leaanthony (approved):

PR #2688, opened 2023-05-26
feat(website): update the `Crowdin` branch

misitebao:
# Description

- [x] Update the `Crowdin` branch corresponding to the v2 document. 

Put the `v2` document into the v2 branch of Crowdin, and update the translation progress of each language of the previously obtained project to the translation progress of each language of the corresponding branch of the obtained project.(This is also to ensure that adding v3 documents to Crowdin will not affect the translation progress data of v2 documents.)

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -11,9 +11,7 @@
     "clear": "docusaurus clear",
     "serve": "docusaurus serve",
     "write-translations": "docusaurus write-translations",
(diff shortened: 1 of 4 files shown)


website/package.json
misitebao, line 1: The scripts here are not used and have been migrated to Taskfile.

leaanthony (approved):
Nice one!

PR #2689, opened 2023-05-27
chore: add `Prettier` as format tool

misitebao:
# Description

- [x] Add `Prettier` as format tool
- [x] Add workflow to auto format Markdown(x) files

Add `Prettier` tool to the root directory of each major version and its corresponding `website` directory to format all `.md` and `.mdx` files. However, to reduce unnecessary work, all files synchronized by Crowdin are ignored because formatting these files is meaningless.

### DIFF
--- a/.github/workflows/format-markdown-files.yml
+++ b/.github/workflows/format-markdown-files.yml
@@ -1,0 +1,37 @@
+name: Format Markdown Files
+
+on:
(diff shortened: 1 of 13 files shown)


leaanthony (approved):
Thanks for doing this! Let's try it out :-)

PR #2745, opened 2023-06-26
[v2, runtime] Separate runtime js debug build

mmghv:
# Description

Currently, the runtime javascript is only built for `production` and `dev`, so in `debug` the minified version of the runtime is served.
This PR separates the `debug` build of the runtime so it includes the un-minified version with the source map.

# How Has This Been Tested?

- [x] Windows
- [x] macOS
- [x] Linux

## How to test :

In devtools, check `wails/runtime.js` in the `source` tab 

`wails dev` => `wails/runtime.js` is un-minified, `wails/desktop` exists.
`wails build -debug` => `wails/runtime.js` is un-minified, `wails/desktop` exists.
`wails build -devtools` => `wails/runtime.js` is minified, `wails/desktop` doesn't exist.


leaanthony:
Thanks for doing this @mmghv 🙏 I think we can simplify this as the debug and dev versions are identical. I had a quick look at the inclusion logic and I think if we completely remove the `runtime_dev*` files (I prefer debug!) and then change the build tags logic at the top of `runtime_debug_desktop.go` to be: `//go:build !production || (production && debug)` then I think that covers all scenarios and removes the duplicate. We could then also get rid of `build:runtime-desktop-dev` script in `package.json`. Thoughts?

mmghv:
@leaanthony Done 👍

leaanthony (approved):
Really nice 👍

PR #2750, opened 2023-06-29
[v3] Fix examples background color

mmghv:
Some examples had a white background behind a white text on windows.


leaanthony (approved):
Thanks! 🙏

PR #2774, opened 2023-07-16
[v2/windows] implement full file path resolver

### LINKED ISSUE #1090: [FEAT]: Native drap and drop for file and folder with complete filepath
Currently, on Wails v1.16.8 as for windows development it depends on IE11 (as Windows development use the mshtml library which is only compatible with IE11), and On IE11, the drag and drop is not supported for folder. Whereas with Chrome, we can have file or folder drag and drop functionality.

Feature that can be suitable would be native drag and drop with absolute path of dropped file/folder and if the dependency to IE11 can be removed.

**Describe alternatives you've considered**
Currently, We have tried the following approaches but not complete solution was found:
1. Use [ngx-file-drop](https://www.npmjs.com/package/ngx-file-drop) library for drag and drop, it worked fine for chrome but for IE11, only file drop was working fine and not folder. But was not providing absolute path of neither file not folder..
2. HTML Drag and drop: Failed for both file and folder.
3. Drag and drop on Input Type=File: Failed for both file and folder.
4. Created our own directive even it failed for both file and folder.

**Additional context**
Screenshot from example Feature Request

ayatkyo:
# Description

This PR add a feature for resolving the full path of js [File](https://developer.mozilla.org/en-US/docs/Web/API/File).

Currently only tested in Windows 11 with WebView2 114.0.1823.79.

Fixes #1090 but only for Windows with WebView2 version >= 1.0.1774.30.

## Usage Demo

Source code of this demo is on https://github.com/ayatkyo/wails-2-file-drop-test

https://github.com/wailsapp/wails/assets/2143364/54a8781f-c098-4bbf-bd32-cd8550690255

## Type of change
  
## Another Solution

The previous solution that I tried before using this is using an overlay view in front of the webview, so it is something like this:
- Create a hidden Overlay view in front of the webview
- In the JS side, listen to `dragenter` event then get the target element size and position using `event.target.getBoundingClientRect()`
- Update Overlay size and position using that information
- Show Overlay and handle file drop event
- Send dropped file list to js side

But this solution will trigger `dragleave` to target the element on the JS side as soon as we show the Overlay view, but we can trigger `dragover` programmatically to webview with the mouse position (I have not tried this).

# Checklist:

### DIFF
--- a/v2/internal/frontend/desktop/windows/frontend.go
+++ b/v2/internal/frontend/desktop/windows/frontend.go
@@ -502,6 +502,10 @@ func (f *Frontend) setupChromium() {
 	if err != nil {
 		log.Fatal(err)
 	}
+	err = chromium.CheckCanAccessAdditionalWebMessageObjects()
+	if err != nil {
+		log.Fatal(err)
+	}
 	err = settings.PutAreDefaultContextMenusEnabled(f.devtools || f.frontendOptions.EnableDefaultContextMenu)
 	if err != nil {
 		log.Fatal(err)
--- a/v2/internal/frontend/desktop/windows/frontend.go
+++ b/v2/internal/frontend/desktop/windows/frontend.go
@@ -19,6 +19,7 @@ import (
 	"sync"
 	"text/template"
 	"time"
+	"unsafe"
 
 	"github.com/bep/debounce"
 	"github.com/wailsapp/go-webview2/pkg/edge"
@@ -27,6 +28,7 @@ import (
 	"github.com/wailsapp/wails/v2/internal/frontend/desktop/windows/win32"
 	"github.com/wailsapp/wails/v2/internal/frontend/desktop/windows/winc"
 	"github.com/wailsapp/wails/v2/internal/frontend/desktop/windows/winc/w32"
+	"github.com/wailsapp/wails/v2/internal/frontend/dispatcher"
 	wailsruntime "github.com/wailsapp/wails/v2/internal/frontend/runtime"
 	"github.com/wailsapp/wails/v2/internal/logger"
 	"github.com/wailsapp/wails/v2/internal/system/operatingsystem"
@@ -38,6 +40,10 @@ import (
 const startURL = "http://wails.localhost/"
 
 type Screen = frontend.Screen
+
+type File struct {
+	Path string `json:"path"`
+}
 
 type Frontend struct {
 
@@ -451,6 +457,7 @@ func (f *Frontend) setupChromium() {
 	}
 
 	chromium.MessageCallback = f.processMessage
+	chromium.MessageWithAdditionalObjectsCallback = f.processMessageWithAdditionalObjects
 	chromium.WebResourceRequestedCallback = f.processRequest
 	chromium.NavigationCompletedCallback = f.navigationCompleted
 	chromium.AcceleratorKeyCallback = func(vkey uint) bool {
@@ -687,6 +694,62 @@ func (f *Frontend) processMessage(message string) {
 			f.logger.Info("Unknown message returned from dispatcher: %+v", result)
 		}
 	}()
+}
+
+func (f *Frontend) processMessageWithAdditionalObjects(message string, sender *edge.ICoreWebView2, args *edge.ICoreWebView2WebMessageReceivedEventArgs) {
+	if strings.HasPrefix(message, "file:") {
+		callbackID := message[5:]
+		objs, err := args.GetAdditionalObjects()
+		if err != nil {
+			f.logger.Error(err.Error())
+			return
+		}
+
+		if objs == nil {
+			return
+		}
+
+		_count, err := objs.GetCount()
+		if err != nil {
+			f.logger.Error(err.Error())
+			return
+		}
+
+		files := []File{}
+		count := *(*uint32)(unsafe.Pointer(&_count))
+		for i := uint32(0); i < count; i++ {
+			_file, err := objs.GetValueAtIndex(i)
+			if err != nil {
+				f.logger.Error("cannot get value at %d : %s", i, err.Error())
+				return
+			}
+
+			file := *(**edge.ICoreWebView2File)(unsafe.Pointer(&_file))
+			filepath, err := file.GetPath()
+			if err != nil {
+				f.logger.Error("cannot get path for object at %d : %s", i, err.Error())
+				return
+			}
+
+			files = append(files, File{
+				Path: filepath,
+			})
+		}
+
+		callbackMessage := &dispatcher.CallbackMessage{
+			CallbackID: callbackID,
+		}
+		callbackMessage.Result = files
+
+		messageData, err := json.Marshal(callbackMessage)
+		if err != nil {
+			f.logger.Error("cannot marshal result: %s", err)
+			return
+		}
+
+		f.Callback(string(messageData))
+		return
+	}
 }
 
 func (f *Frontend) Callback(message string) {
--- a/v2/internal/frontend/desktop/windows/frontend.go
+++ b/v2/internal/frontend/desktop/windows/frontend.go
@@ -701,16 +701,12 @@ func (f *Frontend) processMessageWithAdditionalObjects(message string, sender *e
 		callbackID := message[5:]
 
 		objs, err := args.GetAdditionalObjects()
-		defer objs.Release()
-
 		if err != nil {
 			f.logger.Error(err.Error())
 			return
 		}
 
-		if objs == nil {
-			return
-		}
+		defer objs.Release()
 
 		count, err := objs.GetCount()
 		if err != nil {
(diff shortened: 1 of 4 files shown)


leaanthony:
I think we need to think about the scenario where the webview runtime doesn't support this feature. Any thoughts on how the developer would manage that?

ayatkyo:
> I think we need to think about the scenario where the webview runtime doesn't support this feature. Any thoughts on how the developer would manage that?

Maybe we can make a function to check if this feature is available with `CanResolveFilePaths`, so the developer can handle other case based on that value.

Because another solution is using a different way to get the file path, for example when using overlay as a drop target, we cannot use the same `ResolveFilePaths` function that accepts the file list, but we need to listen for `file-drop` event in js side.

Maybe it will be something like this:
- If `CanResolveFilePaths` = true, handle file drop on the target element using `ResolveFilePaths`
- Else, listen `wails-file-drop` event on the target element.

As far as I know, right now only WebView2 provides an API to resolve the js `File` path on the native side. WebView2 on another platform is on their [Roadmap](https://learn.microsoft.com/en-us/microsoft-edge/webview2/roadmap).

v2/internal/frontend/desktop/windows/frontend.go
```diff
@@ -689,6 +696,62 @@ func (f *Frontend) processMessage(message string) {
...
+		files := []File{}
+		count := *(*uint32)(unsafe.Pointer(&_count))
+		for i := uint32(0); i < count; i++ {
+			_file, err := objs.GetValueAtIndex(i)
+			if err != nil {
+				f.logger.Error("cannot get value at %d : %s", i, err.Error())
+				return
+			}
+
+			file := *(**edge.ICoreWebView2File)(unsafe.Pointer(&_file))
```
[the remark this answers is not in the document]
ayatkyo: There is no explanation about it in the docs. But I think it is better to release the resource to avoid memory leaks, or maybe it will be cleared by the garbage collector? honestly, I don't know much about Golang memory 😅.

stffabi, line 727: I think the following should be equivalent and should make reasoning about the pointer conversion more understandable.
```suggestion
			file := (*edge.ICoreWebView2File)(unsafe.Pointer(_file))
```

stffabi (changes requested):
Awesome job @ayatkyo, thanks so much for all your time you have put into this 🙏 

I've added some comments about COM memory management and pointer conversions.

ayatkyo:
> Awesome job @ayatkyo, thanks so much for all your time you have put into this 🙏
> 
> I've added some comments about COM memory management and pointer conversions.

Sorry for the late reply.

Thanks for the comments, I learn a lot from that, especially the pointer, I will update the code 🙏

If I not mistaken, we will need to update `ICoreWebView2File.go` to be something like this:

```go
// ...
func (i *ICoreWebView2File) Release() error {
  return i.vtbl.CallRelease(unsafe.Pointer(i))
}
// ...
```

Then call the `Release` function using `defer`

```go
file := (*edge.ICoreWebView2File)(unsafe.Pointer(_file))
defer file.Release()
```

Also for `ICoreWebView2ObjectCollectionView.go`

```go
// ...
func (i *ICoreWebView2ObjectCollectionView) Release() error {
  return i.vtbl.CallRelease(unsafe.Pointer(i))
}
// ...
```

```go
objs, err := args.GetAdditionalObjects()
defer objs.Release()
if err != nil {
  f.logger.Error(err.Error())
  return
}
```

v2/internal/frontend/desktop/windows/frontend.go
```diff
@@ -709,14 +712,13 @@ func (f *Frontend) processMessageWithAdditionalObjects(message string, sender *e
...
+		count, err := objs.GetCount()
 		if err != nil {
 			f.logger.Error(err.Error())
 			return
 		}
 
 		files := []File{}
-		count := *(*uint32)(unsafe.Pointer(&_count))
 		for i := uint32(0); i < count; i++ {
 			_file, err := objs.GetValueAtIndex(i)
```
ayatkyo, line 723: Maybe we also need to Release here? because we will return if there is an error in the next line.

leaanthony: Using `defer` is a common technique to make sure it's released no matter how the function exists. The defer could release all the memory allocated

ayatkyo: Thanks, I have updated the code

leaanthony: I have a sneaky suspicion that this will only release the last _file. I think it will also attempt a release if `err != nil`, which might fail spectacularly as `_file` is unlikely to have been allocated.

v2/internal/frontend/desktop/windows/frontend.go
```diff
@@ -689,6 +696,73 @@ func (f *Frontend) processMessage(message string) {
...
+					f.logger.Error("cannot release resource: %s", err.Error())
+				}
+			}()
+			if err != nil {
+				f.logger.Error("cannot get value at %d : %s", i, err.Error())
+				return
+			}
+
+			file := (*edge.ICoreWebView2File)(unsafe.Pointer(_file))
+			defer file.Release()
```
leaanthony, line 736: Ah we have this release here. That may end up being a double release.

leaanthony:
My suggestions for releasing would be:
- Create a defer function for each item you *know* is allocated (not before the error condition
- Pass in the thing you want to release as a parameter to that method: 
```
defer func(itemToRelease *item) { 
   //release the item 
}(item)
```

I *think* that should work as it copies the reference.

v2/internal/frontend/desktop/windows/frontend.go
ayatkyo: I have a question, is this casting will create a new allocation or it will reuse the previous allocation (returned by `GetValueAtIndex`)? 

If it is reused, then can we remove the defer function above, as it will be released here?

stffabi: This is just a pointer conversion as such, the reference counter will not be increased and you don't need to call `Release` an additional time. In contrast if you use `IUnknown.QueryInterface` you would need to call `Release` on both `_file` and `file`.

v2/internal/frontend/desktop/windows/frontend.go
```diff
@@ -689,6 +696,73 @@ func (f *Frontend) processMessage(message string) {
...
+		if err != nil {
+			f.logger.Error(err.Error())
+			return
+		}
+
+		files := []File{}
+		for i := uint32(0); i < count; i++ {
+			_file, err := objs.GetValueAtIndex(i)
+			defer func() {
+				err := _file.CallRelease(unsafe.Pointer(_file))
```
stffabi, line 725: If `GetValueAtIndex` has returned an error `_file` would be nil and this defer will then panic. I think it would be better to move it after the `if err != nil` block.

v2/internal/frontend/desktop/windows/frontend.go
```diff
@@ -689,6 +696,73 @@ func (f *Frontend) processMessage(message string) {
 	}()
 }
 
+func (f *Frontend) processMessageWithAdditionalObjects(message string, sender *edge.ICoreWebView2, args *edge.ICoreWebView2WebMessageReceivedEventArgs) {
+	if strings.HasPrefix(message, "file:") {
+		callbackID := message[5:]
+
+		objs, err := args.GetAdditionalObjects()
+		defer objs.Release()
```
stffabi, line 704: This should also be moved after the `if err != nil` block, so we don't panic if `GetAdditionalObjects` has returned an error.

ayatkyo:
I have updated the code, thanks a lot for guiding me on this PR @leaanthony @stffabi, I really appreciate that.

stffabi (approved):
LGTM, thanks for all your time you have put into this.

I suspect we also need to update `go-webview2` with the changed method signature for `GetCount`. But this part looks good to me.

ayatkyo:
Thank you for your response. You are right. I have made some changes to the code and created a new PR for `go-webview2`.

The previous code caused the application to crash when using WebView2 SDK versions < `1.0.1774.30` due to the unavailability of `GetAdditionalObjects`.

leaanthony:
I've commented on the go-webview PR so I think we're pretty close to getting it in. One question: Am I right in thinking that if a webview runtime is installed which does not support `GetAdditionalObjects`,that means DnD is unavailable by default?

ayatkyo:
Thanks for the reply, let's move the discussion to go-webview PR, sorry for the late reply.

> Am I right in thinking that if a webview runtime is installed which does not support GetAdditionalObjects,that means DnD is unavailable by default?

Yes, to get full file path, frontend needs to send the File object using `postMessageWithAdditionalObjects` then on the native side, we use `GetAdditionalObjects` alongside `GetValueAtIndex` to get the file object.

Unfortunately, we cannot simply just use `window.chrome.webview.postMessageWithAdditionalObjects` to check if this feature is supported, because I have checked that function is also available on WebView2 1.0.1722.45 which does not have `GetAdditionalObjects`.

leaanthony:
@ayatkyo - Do you still want to proceed with this PR? I think `go-webview2` is ready for this now.

PR #2782, opened 2023-07-19
Add Fullscreen HTML content hook

leaanthony:
# Description

This PR adds a hook into the `ContainsFullScreenElementChanged` event from WebView2 to fullscreen/unfullscreen automatically when elements request to go fullscreen.

It works, but the window moves slightly erratically. What seems to be happening is:
- Press button to go fullscreen - works reasonably well
- Press button to unfullscreen and the window returns to original size, but is left aligned with the window
- The unfullscreen happens but the window "goes to" the left to wrap the content
- It then realises it wasn't meant to be there and jumps back to the original position

https://github.com/wailsapp/wails/assets/1943904/44ac806a-b1cd-427d-ad1a-052173cdb329

### DIFF
--- a/v2/go.mod
+++ b/v2/go.mod
@@ -33,7 +33,7 @@ require (
 	github.com/tc-hib/winres v0.1.5
 	github.com/tidwall/sjson v1.1.7
 	github.com/tkrajina/go-reflector v0.5.6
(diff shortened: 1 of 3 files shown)


leaanthony:
Bumping to v3

PR #2891, opened 2023-09-08
docs: make the install command easier to copy

nguyenvanduocit:
# Description

just change doc to make the installation command easier to copy.


leaanthony (approved):
Nice one!

PR #2926, opened 2023-09-19
[v2, windows] Support async request processing on AssetServer

### LINKED ISSUE #2890: More efficient frontend image sending.

I am working on an application that generates images on the *"backend"* side and then sends them to the *"frontend"* as a preview.

From measurements, it seems that a lot of time is wasted on the very stage of preparing the data for transport and transporting them between the *"backend"* and the *"frontend"*. 
For context, the "heavy" image processing step takes approx. 300ms and the transport stage approx. +800ms.
This measurement was taken on Linux with the GNOME 44.4 DE and Wails version 2.5.1.

The current approach is to encode the result image to `base64`, send this string via a bind function to the *"frontend"* and set the `src` attribute of the `` element.

Above all, I would like to ask those more experienced in using Wails if I have overlooked some simpler obvious solution.

I've been thinking about providing an API that allows you to submit primitive types bypassing the logic of mapping data to JavaScript/TypeScript structures.

I was also thinking about the possibility of streaming results using other network protocols such as WebSockets/WebRTC, but an issue dedicated to this already exists.

If it turned out that the problem was caused, for example, by the very long `base64` interpretation time of the HTML engine. We could try to send a RGB byte array through the binding functions and iterate over a `<canvas>` element, which should work relatively efficiently thanks to the support for hardware acceleration. The problem is, however, that the amount of data (measured in bytes) to be sent to the *"frontend"* will be much larger.

Alternatively, we can try to send the image as a event payload, but this will require a lot of changes to the whole application, and the events themselves (according to what I read in a another issue), do not necessarily keep their order of emission.

### Additional context

_No response_

stffabi:
Fixes #2890


leaanthony (approved):
Thanks as always @stffabi 🙏 LGTM!

stffabi:
You're welcome and thanks for approving.

website/src/pages/changelog.mdx
```diff
@@ -18,6 +18,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added support for enabling/disabling swipe gestures for Windows WebView2. Added by @leaanthony in [PR](https://github.com/wailsapp/wails/pull/2878)
 - When building with `-devtools` flag, CMD/CTRL+SHIFT+F12 can be used to open the devtools. Added by @leaanthony in [PR](https://github.com/wailsapp/wails/pull/2915)
 
+### Changed
+
+- AssetServer requests are now processed asynchronously without blocking the main thread on Windows. Changed by @stffabi in [PR](https://github.com/wailsapp/wails/pull/2926)
+- AssetServer requests are now processed concurrently by spawning a goroutine per request. Changed by @stffabi in [PR](https://github.com/wailsapp/wails/pull/2926)
```
mmghv, line 24: Duplicate record?
When fixing this please merge it with the other `Changed` section, got mixed with the devtools PR, don't know how it didn't get marked as conflict.

stffabi: That's not a duplicate record, they both have the same prefix so they might look like duplicate but they aren't.

- The first describes that we now don't block the main-thread during the processing of a request on Windows. 
- The second describes we now spawn a goroutine per request, allowing them to be processed concurrently. Previously we had 10 worker goroutines processing requests.

stffabi: PR #2940 is up to fix the duplicate `changed` section that came in with the devtools PR.

leaanthony: Merged. Git black magic bites again

Issue #2099, opened 2022-11-16
<video/> tag can not be full screen


ghost:
### Description

'https://github.com/webview/webview/' 
Use the repository  to display full-screen buttons for full-screen video playback

-------------

But the same code
use `wails` Can not play full screen

### To Reproduce

vue3 html

```
<video controls width="500" >
        <source
          src="https://interactive-examples.mdn.mozilla.net/media/cc0-videos/flower.webm"
          type="video/webm"
        />
        <source
          src="https://interactive-examples.mdn.mozilla.net/media/cc0-videos/flower.mp4"
          type="video/mp4"
        />
      </video>
```

### Expected behaviour

The full-screen button should be displayed for full-screen playback

### System Details

```shell
------
OS:		MacOS
Version: 	12.3.1
ID:		21E258
Go Version:	go1.18.5
Platform:	darwin
Architecture:	amd64

Wails
------
Version: 	v2.1.0
```

leaanthony:
Hi 👋 Why are you referencing 'https://github.com/webview/webview/'?

758F08FE920B1765:
@leaanthony  hi
Because I suspect a browser kernel problem。 

[webview](https://github.com/webview/webview) This is just an example. 
It uses the webkit kernel, which renders `<video />`  perfectly and supports full-screen playback.

but wails webkit kernel， `<video />` Full screen playback， it doesn't work

leaanthony:
AFAIK they use the same engine: WkWebView

krishna8421:
Any update on this?, I want to make a video go full screen but unable to do it.

leaanthony:
@krishna8421 Please can you retry using the latest version. Remember to upgrade the version of Wails in you project's `go.mod`.

krishna8421:
@leaanthony Hey updated to v2.5.1, full screen working on the browser but not in the app. can you give some example codes or reference how to do it, maybe I am doing something wrong.

fkhadra:
@leaanthony  I'm wondering if allowing to set [elementFullscreenEnabled](https://developer.apple.com/documentation/webkit/wkpreferences/3917769-elementfullscreenenabled?language=objc) in the preferences on mac won't address the issue. I'll check

fkhadra:
So just tested and it seems to work as expected :). 
https://github.com/wailsapp/wails/assets/5574267/eb338fe3-437f-4f9f-9155-14d12145a7ba

leaanthony:
[...]

PR #2937, opened 2023-09-23
[v2 mac] Allow to specify webview preferences

### LINKED ISSUE #2919: [OSX] Allow to set tabFocusesLinks

"By default, Apple computers have an operating system (OS) level setting that limits the Tab key to "Text boxes and lists only". This can be confusing, if not frustrating, for those unaware of this setting and expect keyboard focus to be similar to navigating on a Windows machine." For example, links are not focusable by default when using the keyboard.
See https://www.a11yproject.com/posts/macos-browser-keyboard-navigation/ for full details.

Webkit exposes a setting to make links focusable using the keyboard under `WKPrefences.tabFocusesLinks`. This is documented [here](https://developer.apple.com/documentation/webkit/wkpreferences/2818595-tabfocuseslinks?language=objc).

I see 2 possible solutions:
- Always set this parameter to `true`, this way the application behaviour is more consistent across different OS when it comes to keyboard navigation.
- Allow the user to configure this parameter. For example, we could add a new option to `mac.Options.TabFocusesLinks: boolean`.

Locally, I've tried the first solution given it was the fastest to test and it works. I've added the line below.
```objective-c
// file: wails/v2/internal/frontend/desktop/darwin/WailsContext.m
// around line 219, after WKWebViewConfiguration instantiation 
config.preferences.tabFocusesLinks = true;
```

I would be happy to contribute if needed.

Currently, I'm handling the tab navigation with javascript. I create a tree walker with all focusable element and use it to navigate across focusable nodes.

```ts
const treeWalker = document.createTreeWalker(
    document.body,
    NodeFilter.SHOW_ELEMENT,
    {
      acceptNode: (node) =>
        (node as HTMLElement).tabIndex >= 0
          ? NodeFilter.FILTER_ACCEPT
          : NodeFilter.FILTER_SKIP,
    },
  );

  function handleTab(e: KeyboardEvent) {
    const focusNext = e.key === "Tab" && !e.shiftKey;
    const focusPrevious = e.key === "Tab" && e.shiftKey;

    if (focusNext || focusPrevious) {
      e.preventDefault();
      treeWalker.currentNode = document.activeElement as Node;

      let next = focusNext ? treeWalker.nextNode() : treeWalker.previousNode();

      if (next === null) {
        treeWalker.currentNode = treeWalker.root;
        next = focusNext ? treeWalker.nextNode() : treeWalker.lastChild();
      }

      (next as HTMLElement).focus();
    }
  }

  document.addEventListener("keydown", handleTab);
```

### Additional context

_No response_

fkhadra:
# Description

This PR allows to set some of the preferences for the Webview on osx, notably `tabFocusesLinks` and `textInteractionEnabled`. By default, `tabFocusesLinks` is disabled, preventing the user to select links with the tab key.  One could use javascript to workaround this issue, but it's very brittle and error prone. Having a way to turn this setting on increase the accessibility of the app.

Fixes #2919 

## Type of change
  
# How Has This Been Tested?
  
I've tested the changes locally with a project I'm currently working on by adding a replace statement in my go.mod to point to the version with my changes. I've added a recording:

https://github.com/wailsapp/wails/assets/5574267/8e7072f8-e2fb-470f-8102-99defaf08abe

- during the first app launch, `tabFocusesLinks` is enabled, pressing tab moves the focus ring to overview, settings, etc... Also I'm not unable to select any text as `textInteractionEnabled` is disabled
- after changing the settings, pressing tabs does not focus overview, settings, etc..., only button are focusable. And text is selectable again

In the recording, to set the preferences, I was using `u.Bool` directly, it requires end user to have the package in their dependencies. 

```go
require (
	github.com/leaanthony/u v1.0.1
	github.com/wailsapp/wails/v2 v2.6.0
	github.com/zalando/go-keyring v0.2.3
	go.uber.org/zap v1.26.0
	golang.org/x/net v0.15.0
)
```

It's not an issue per se but I've decided to use a more user friendly solution to avoid exposing implementation details to the end user. I've added 2 functions: `mac.EnablePreference: u.Bool` and `mac.DisablePreference`.

```go
Preferences: &mac.Preferences{
		TabFocusesLinks:           mac.EnablePreference(),
		TextInteractionEnabled: mac.DisablePreference(),
}
```

Regarding the name of the preferences, I've used the one from apple documentation.
The repository I've used in the recording is private but I can setup a simple app if needed.

- [ ] Windows
- [x] macOS
- [ ] Linux
  
## Test Configuration

```sh
# Wails
Version | v2.6.0

# System
┌─────────────────────────┐
| OS           | MacOS    |
| Version      | 13.5.2   |
| ID           | 22G91    |
| Go Version   | go1.21.0 |
| Platform     | darwin   |
| Architecture | arm64    |
└─────────────────────────┘

# Dependencies
┌──────────────────────────────────────────────────────────────────┐
| Dependency                | Package Name | Status    | Version   |
| Xcode command line tools  | N/A          | Installed | 2397      |
| Nodejs                    | N/A          | Installed | 18.17.0   |
| npm                       | N/A          | Installed | 9.6.7     |
| *Xcode                    | N/A          | Available |           |
| *upx                      | N/A          | Installed | upx 4.1.0 |
| *nsis                     | N/A          | Available |           |
└──────────────────── * - Optional Dependency ─────────────────────┘

# Diagnosis
  - Xcode: Available at https://apps.apple.com/us/app/xcode/id497799835

 ♥   If Wails is useful to you or your company, please consider sponsoring the project:
```
# Checklist:

### DIFF
--- a/v2/pkg/options/mac/preferences.go
+++ b/v2/pkg/options/mac/preferences.go
@@ -1,0 +1,21 @@
+package mac
+
+import "github.com/leaanthony/u"
+
+// Preferences allows to set webkit preferences
+type Preferences struct {
+	// A Boolean value that indicates whether pressing the tab key changes the focus to links and form controls.
+	// Set to false by default.
+	TabFocusesLinks u.Bool
+	// A Boolean value that indicates whether to allow people to select or otherwise interact with text.
+	// Set to true by default.
+	TextInteractionEnabled u.Bool
+}
+
+func EnablePreference() u.Bool {
+	return u.NewBool(true)
+}
+
+func DisablePreference() u.Bool {
+	return u.NewBool(true)
+}
--- a/v2/pkg/options/mac/preferences.go
+++ b/v2/pkg/options/mac/preferences.go
@@ -2,6 +2,9 @@ package mac
 
 import "github.com/leaanthony/u"
 
+var Enabled = u.True
+var Disabled = u.False
+
 // Preferences allows to set webkit preferences
 type Preferences struct {
 	// A Boolean value that indicates whether pressing the tab key changes the focus to links and form controls.
@@ -10,12 +13,4 @@ type Preferences struct {
 	// A Boolean value that indicates whether to allow people to select or otherwise interact with text.
 	// Set to true by default.
 	TextInteractionEnabled u.Bool
-}
-
-func EnablePreference() u.Bool {
-	return u.NewBool(true)
-}
-
-func DisablePreference() u.Bool {
-	return u.NewBool(true)
 }
--- a/v2/go.mod
+++ b/v2/go.mod
@@ -22,7 +22,7 @@ require (
 	github.com/leaanthony/go-ansi-parser v1.6.0
 	github.com/leaanthony/gosod v1.0.3
 	github.com/leaanthony/slicer v1.6.0
-	github.com/leaanthony/u v1.0.1
+	github.com/leaanthony/u v1.1.0
 	github.com/leaanthony/winicon v1.0.0
 	github.com/matryer/is v1.4.0
 	github.com/pkg/browser v0.0.0-20210911075715-681adbf594b8
--- a/v2/go.mod
+++ b/v2/go.mod
@@ -22,6 +22,7 @@ require (
 	github.com/leaanthony/go-ansi-parser v1.6.0
 	github.com/leaanthony/gosod v1.0.3
 	github.com/leaanthony/slicer v1.6.0
+	github.com/leaanthony/u v1.0.1
 	github.com/leaanthony/winicon v1.0.0
 	github.com/matryer/is v1.4.0
 	github.com/pkg/browser v0.0.0-20210911075715-681adbf594b8
--- a/v2/go.sum
+++ b/v2/go.sum
@@ -119,6 +119,8 @@ github.com/leaanthony/gosod v1.0.3/go.mod h1:BJ2J+oHsQIyIQpnLPjnqFGTMnOZXDbvWtRC
 github.com/leaanthony/slicer v1.5.0/go.mod h1:FwrApmf8gOrpzEWM2J/9Lh79tyq8KTX5AzRtwV7m4AY=
 github.com/leaanthony/slicer v1.6.0 h1:1RFP5uiPJvT93TAHi+ipd3NACobkW53yUiBqZheE/Js=
 github.com/leaanthony/slicer v1.6.0/go.mod h1:o/Iz29g7LN0GqH3aMjWAe90381nyZlDNquK+mtH2Fj8=
+github.com/leaanthony/u v1.0.1 h1:tJLpf9EsuCgSB02ojRxg8KRqMgRN6mCTvFwI55kxRFE=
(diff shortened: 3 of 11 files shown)


v2/pkg/options/mac/preferences.go
```diff
@@ -0,0 +1,21 @@
...
+	// Set to true by default.
+	TextInteractionEnabled u.Bool
+}
+
+func EnablePreference() u.Bool {
+	return u.NewBool(true)
+}
+
+func DisablePreference() u.Bool {
+	return u.NewBool(true)
```
leaanthony, line 20: I think this might be false? We can now use `u.True` and `u.False` as of `v1.1.0`

leaanthony: The other way we could do this is like this: 
```go
type Enabled = u.True
type Disabled = u.False
```

then you can have: 
```go
Mac: &mac.Options{
    Preferences: &mac.Preferences{
		TabFocusesLinks:        mac.Enabled,
		TextInteractionEnabled: mac.Disabled,
	}
}
```

leaanthony (changes requested):
Great PR! 🙏 
There's 1 thing we need to change and a couple of things to consider.

v2/pkg/options/mac/preferences.go
fkhadra: Even better, I've missed the upgrade to`v1.1.0`. You meant to use `var` instead of `type`, right?

leaanthony: Ah yes of course 😅

leaanthony (approved):

leaanthony:
There's an interesting error [here](https://github.com/wailsapp/wails/actions/runs/6289516348/job/17076078654?pr=2937#step:5:383) @fkhadra - any ideas? It does appear to work when i use it in an app. Just wondering what is affecting this test.

fkhadra:
Looking into it.

fkhadra:
Running the test suites locally yielded no errors. Still trying to understand what's going on 🤔 
Screenshot 2023-09-24 at 13 06 01

fkhadra:
I'm using an M1 mac btw, I'm wondering is the issue can be caused by the CPU arch.

fkhadra:
Regarding my latest commit,  why I've changed `BOOL` to `bool`. 
https://opensource.apple.com/source/objc4/objc4-706/runtime/objc.h.auto.html
```c
/// Type to represent a boolean value.
#if (TARGET_OS_IPHONE && __LP64__)  ||  TARGET_OS_WATCH
#define OBJC_BOOL_IS_BOOL 1
typedef bool BOOL;
#else
#define OBJC_BOOL_IS_CHAR 1
typedef signed char BOOL; 
// BOOL is explicitly signed so @encode(BOOL) == "c" rather than "C" 
// even if -funsigned-char is used.
#endif
```

fkhadra:
Yes I'll sleep a bit smarter than I woke up 😄. I was wondering if you'd like me to address [<video/> tag can not be full screen](https://github.com/wailsapp/wails/issues/2099) as well in this PR or I can open another one when this one is merged.

leaanthony:
Thanks for the fix! Let's keep the PRs separate 👍

leaanthony (approved):

leaanthony:
Thanks for this! 🙏

fkhadra:
Awesome! Ill open a new PR

Issue #1351, opened 2022-04-25
Integrated Single Process Option of Wails app


Purple-CSGO:
### Describe the solution you'd like

Currently wails does not have integrated option to make sure only 1 process is running at a time. Pls consider add this feature to wails. Thx.

KiddoV:
In the mean time, you can do something like this, not sure if it is a good method, but it seem to work for me.
```go
// Create a mutex to keep a program running on single instance. Return error if program is already running.
func CreateMutex(name string) (uintptr, error) {
	kernel32 := syscall.NewLazyDLL("kernel32.dll")
	procCreateMutex := kernel32.NewProc("CreateMutexW")
	ret, _, err := procCreateMutex.Call(0, 0, uintptr(unsafe.Pointer(syscall.StringToUTF16Ptr(name))))
	switch int(err.(syscall.Errno)) {
	case 0:
		return ret, nil
	default:
		return ret, err
	}
}
```
Call the function in the ``main.go`` just before ``app := NewApp()``.
```go
import (
    sysWindows "golang.org/x/sys/windows"
)
// Create mutex to keep app running in single instance
if _, mutexErr := CreateMutex("YourAppName"); mutexErr != nil {
        // Do whatever... for me, I call a messagebox...
	sysWindows.MessageBox(
		sysWindows.GetShellWindow(),
		syscall.StringToUTF16Ptr("Info-Hub Server application is already running!"),
		syscall.StringToUTF16Ptr("Warning"),
		sysWindows.MB_OK|sysWindows.MB_ICONEXCLAMATION|sysWindows.MB_SYSTEMMODAL) //See https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-messageboxw for details
	os.Exit(0) //Exit if the program is already running
}
```

avengerweb:
file locks is one of ways to go
https://github.com/postfinance/single

NanoNik:
There's another way making this on windows - Events.
This way your first process will know that new process is launched, and, for eaxample, bring your main window to front instead of openning second process.

[CreateEventW](https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-createeventw)
[SetEvent](https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-setevent)
[WaitForSingleObject](https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-waitforsingleobject)

[...]

PR #2918, opened 2023-09-17
Implement file association Open a file from Finder/Explorer

APshenkin:
# Description

Implements file association for macOS/Windows https://github.com/wailsapp/wails/issues/1830

Users can specify list of associated file types in `wails.json` config with following params
```
"ext": :"The extension (minus the leading period). e.g. png"
"name": "The name. e.g. PNG File"
"iconName": "The icon name without extension. Icons should be located in build folder. Proper icons will be generated from .png file for both macOS and Windows)"
"description": "Windows-only. The description. It is displayed on the `Type` column on Windows Explorer."
"role": "macOS-only. The app’s role with respect to the type. Corresponds to CFBundleTypeRole."
```

Also custom icons for files are supported 🚀

For Linux will try to investigate what can be done later

Few comments regarding implementation
MacOS: Darwin handle opening file by raising `openFile` event. Callback is added for app configuration to handle this correctly.
Windows: On Windows fileAssociation is done via NSIS installer. So apps should be distributed with it to have this feature. 
When associated file is opened, Windows launches new instance of the app and pass filePath as argument. So apps developers should parse args and handle situation properly. However it's common practic to frameworks like Electron or Tauri provide capabilities to lock single instance and handle cases, when new instance is opened. Seems wails v3 will have support of it https://github.com/wailsapp/wails/issues/1351 . But currently this plugin not supporting passing arguments from new instance for first one https://github.com/wailsapp/wails/blob/v3-alpha/v3/plugins/single_instance/plugin_windows.go
Linux: To setup association and icons changes in client system should be done. As wails doesn't support bundling for linux now, nothing was added in this PR for it, but in test repo that is provided below you can see how this can be done using nfmp

You can see how it works here.
- [x] [Windows](https://github.com/wailsapp/wails/assets/14344430/291bae19-0cd1-4591-a70d-b6a061c96063)
- [x] [MacOs](https://github.com/wailsapp/wails/assets/14344430/90310b9a-2d93-408f-ac16-7a6bc2b69649)
- [x] [Linux](https://github.com/wailsapp/wails/assets/14344430/5f943f3c-42f9-48c6-85b0-7a596ae2f6c3) – NOTE THIS IS NOT PART OF PR AS WAILS DOESN'T SUPPORT BUNDLING FOR LINUX

Fixes # (issue)

## Type of change
  
Please delete options that are not relevant.

- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
[...]

PR #2951, opened 2023-09-28
Implement Single instance lock feature with passing arguments to initial instance

APshenkin:
# Description

Implements Single instance lock feature with additional passing arguments to initial instance https://github.com/wailsapp/wails/issues/1351

There is implementation for single instance lock in V3 as plugin, but it doesn't allow to pass arguments of second instance to first one.

Now in V2 it's possible.
Also in plugin version there was option `ActivateAppOnSubsequentLaunch`. I decided to not put it here. As now we have callbacks when second instance is called, I think it should be on "developer" side how to treat this. For example somebody will want to handle it differently depends on what type of arguments were passed.

Also in addition Windows `runtime.Show()` and `runtime.WindowUnminimise()` were modified for better handling different scenarios related to change focus of apps.

You can see how it works here.
- [x] [Windows](https://github.com/wailsapp/wails/assets/14344430/a5f4d396-daad-4f91-a757-a8b4bd1bcdc8)
- [x] [MacOs](https://github.com/wailsapp/wails/assets/14344430/24e49d80-aa34-44bb-a448-4667c79f2dc6)
- [x] [Linux](https://github.com/wailsapp/wails/assets/14344430/f6377baf-5bd7-4881-aeb8-ed6564b6f87e)

## Type of change
  
# How Has This Been Tested?
  
Test repo is available here https://github.com/APshenkin/wails-open-file/tree/feature/single-lock
PR should be fetched and replace in go.mod should be adjusted

- [x] Windows
- [x] macOS
- [x] Linux
  
## Test Configuration

```
# Wails
Version  | v2.6.0
Revision | a419721dcdf16da7a4afd92b9ba63b2050229a26
Modified | true

# System
┌─────────────────────────┐
| OS           | MacOS    |
| Version      | 13.5.2   |
| ID           | 22G91    |
| Go Version   | go1.20.2 |
| Platform     | darwin   |
| Architecture | arm64    |
└─────────────────────────┘


# System
┌────────────────────────────────────┐
| OS           | Windows 10 Pro      |
| Version      | 2009 (Build: 22621) |
| ID           | 22H2                |
| Go Version   | go1.21.1            |
| Platform     | windows             |
| Architecture | arm64               |
└────────────────────────────────────┘                         

# System
┌────────────────────────────────────┐
| OS           | Ubuntu              |
| Version      | 22.04               |
| ID           | ubuntu              |
| Go Version   | go1.21.1            |
| Platform     | linux               |
| Architecture | arm64               |
└────────────────────────────────────┘

```

# Checklist:


APshenkin:
Should be reviewed after https://github.com/wailsapp/wails/pull/2918

APshenkin:
@leaanthony Maybe we can take a look on this too?

Also I have one more PR to enable deeplinks in wails as well 🚀
Will raise it once this one merged

v2/internal/frontend/desktop/windows/single_instance.go
```diff
@@ -0,0 +1,160 @@
...
+
+// SetupSingleInstance single instance Windows app
+func SetupSingleInstance(uniqueId string) {
+	id := "wails-app-" + uniqueId
+
+	className := id + "-sic"
+	windowName := id + "-siw"
+	mutexName := id + "sim"
+
+	_, err := windows.CreateMutex(nil, false, windows.StringToUTF16Ptr(mutexName))
```
leaanthony, line 44: I believe we need to also call `CloseHandle` for the mutex on shutdown

v2/internal/frontend/desktop/darwin/frontend.go
```diff
@@ -39,6 +39,7 @@ var messageBuffer = make(chan string, 100)
 var requestBuffer = make(chan webview.Request, 100)
 var callbackBuffer = make(chan uint, 10)
 var openFilepathBuffer = make(chan string, 100)
+var secondInstanceBuffer = make(chan options.SecondInstanceData, 100)
```
leaanthony, line 42: Am wondering if this needs to be 100? I imagine there's only normally 1 message getting processed at any time?

v2/pkg/options/options.go
```diff
@@ -165,6 +169,31 @@ func MergeDefaults(appoptions *App) {
 	processDragOptions(appoptions)
 }
 
+type SingleInstanceLock struct {
+	// uniqueId that will be used for setting up messaging between instances
+	UniqueId               string
+	Enabled                bool
```
leaanthony, line 175: Is there a scenario where this is false? As we are using a pointer to the struct, we could assume if it's available then the intent is to enable it?

leaanthony (commented):
Looks good on the whole. I have a few questions. Keen to hear what @stffabi and @tmclane think.

v2/internal/frontend/desktop/windows/single_instance.go
APshenkin: If I understood this docs correctly https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-createmutexw

> Use the [CloseHandle](https://learn.microsoft.com/en-us/windows/desktop/api/handleapi/nf-handleapi-closehandle) function to close the handle. The system closes the handle automatically when the process terminates. The mutex object is destroyed when its last handle has been closed.

Windows shutdown mutexes if process terminates. So it doesn't require explicitly closing

v2/internal/frontend/desktop/darwin/frontend.go
APshenkin: I'm not Golang expert, so not clearly know how buffered channels works.

Does it mean that if there is 100, then it can handle 100 opening at the same time?

If I will create channel with size 1 – then it will process openings sequentially?

v2/internal/frontend/desktop/windows/single_instance.go
leaanthony: I guess I misread something, somewhere 😬

v2/internal/frontend/desktop/darwin/frontend.go
leaanthony: Yes, correct. But it's unlikely that you'd ever get more than 1 at the same time. I think setting this to 5 would be fine?

v2/pkg/options/options.go
APshenkin: Just had in mind to make it more explicit, but we can just check that pointer to the struct exists. Happy to change it

v2/internal/frontend/desktop/darwin/frontend.go
APshenkin: Ok, that make sense. Will update values for them

tmclane (commented):
I would recommend setting this to one because all that means is that if two messages are sent at the same time the second routine  sending will block until the first message is processed. Setting it higher only ensures that the caller doesn't block on sending up to the limit.

v2/internal/frontend/desktop/windows/single_instance.go
```diff
@@ -0,0 +1,160 @@
...
+	class.Cursor = 0
+	class.Background = 0
+	class.MenuName = nil
+	class.ClassName = windows.StringToUTF16Ptr(className)
+	class.IconSm = 0
+
+	w32.RegisterClassEx(&class)
+
+	// create event window that will not be visible for user
+	hwnd := w32.CreateWindowEx(
```
stffabi, line 111: Have you tried using a [MessageOnly](https://learn.microsoft.com/en-us/windows/win32/winmsg/window-features#message-only-windows) window? I'm not sure if your message goes into the broadcast type of messages but I wouldn't expect it since it gets send to a specific HWND.

That might be the easiest to hide it from the user.

v2/internal/frontend/desktop/windows/single_instance.go
```diff
@@ -0,0 +1,160 @@
...
+		// The window technically has to be visible to receive WM_PAINT messages (which are used
+		// for delivering events during resizes), but it isn't displayed to the user because of
+		// the LAYERED style.
+		w32.WS_VISIBLE|w32.WS_POPUP,
+	)
+
+	return hwnd
+}
+
+func uintptrToString(cstr uintptr) string {
```
stffabi, line 148: Can we use the `x/sys` [UTF16PtrToString](https://pkg.go.dev/golang.org/x/sys/windows#UTF16PtrToString) for this? That should be possible by converting the `uintptr` to a `*uint16`.

stffabi (commented):
LGTM, that's an awesome PR thanks @APshenkin 🙏 

I only added two minor comments regarding the Windows code.

v2/internal/frontend/desktop/windows/single_instance.go
APshenkin: Updated. Thank you for hint!

v2/internal/frontend/desktop/windows/single_instance.go
APshenkin: I actually didn't know that such thing exists 😅. I just tried to reimplement it same way as Tauri did. But this MessageOnly window looks better. I tested it locally and seems they works same as our configuration.

Updated, thank you for hint!

leaanthony (approved):
👍 from me

leaanthony:
Thanks for the awesome PR 🙏

stffabi:
Awesome, thanks for your time @APshenkin

PR #3046, opened 2023-11-10
Feature/template consolidation

tmclane:
- Removed the duplicate files from each of the included template directories
- Update the list of templates dynamically out of the embed.FS
- Pull the description of the templates from template's `template.json` file
- Add Linux build target to the Taskfile.yaml in `_common`


leaanthony (approved):
Thanks @tmclane !!!!

PR #3081, opened 2023-11-27
Make contributing to docs easier

daertommy:
With this PR readers can suggest changes to the docs with an easy to use WYSIWYG markdown editor.

[This](https://holocron.so/github/pr/wailsapp/wails/master/editor) is how the editor looks like for this repo.

https://github.com/remorses/docusaurus-example/assets/31321188/43c8bc68-3668-4a25-bf1c-a70eceab7bcb

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -85,14 +85,15 @@ module.exports = async function configCreatorAsync() {
         ({
           docs: {
             sidebarPath: require.resolve("./sidebars.js"),
-            // Please change this to your repo.
-            editUrl: "https://github.com/wailsapp/wails/edit/master/website",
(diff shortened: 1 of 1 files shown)


leaanthony:
Congratulations on making a good looking project 🎉

We've had no feedback yet that editing docs is hard. Could consider it once holocron hits v1. My only concern is that these SSGs generally don't allow a fallback edit url should the first one be unavailable.

Good luck with the project 👍

daertommy:
The editor already redirects to another file in case the specified path does not exist, feel free to close this PR if you don’t find this tool useful 😀

leaanthony:
Will look at it in the future. Good luck! 👍

PR #3221, opened 2024-01-26
chore: add cfn-tracker to community docs

williamsjokvist:
# Description

Added [CFN Tracker](https://github.com/williamsjokvist/cfn-tracker) to the community docs. I listed features as well as the major tools and libraries that we used alongside Wails. I think it can be interesting for those to be listed, so people can check out the repo to see for example how SQLite can be implemented in a Wails app


leaanthony:
Thanks for this! We would love to see this in our showcase channel in Discord! https://discord.gg/YsEQQX3Zaa

PR #3250, opened 2024-02-11
implement native drag and drop for macOS

APshenkin:
# Description

Native drag and drop with full file path for macOS. https://github.com/wailsapp/wails/issues/1090
Also changed few things:
– changed `DragAndDrop` to be a link in app options
– renamed `DragAndDrop.Disable` to `DragAndDrop.DisableWebViewDrop` to be more clear about what we are disabling
– `EventsOff` import was missing draganddrop.js
– Fixed setting class property when moving hover to and away target element

1 UX nuance that I found: if you have Native drag and drop disabled and you try to drop file somewhere, where it's not allowed in UI, you can see animation where element "return back". When you enable native drag and drop you loose this effect. As well that some additional icons are not displayed (e.g. if you drop to `input` field). I think this is because we set `drop` and `dragover` listeners for entire window in `draganddrop.js` here https://github.com/wailsapp/wails/blob/f9df5645efb516ca4d7fb09c4e58f71f776582d9/v2/internal/frontend/runtime/desktop/draganddrop.js#L183
This is minor thing and I think we can go with it.

## Type of change
  
# How Has This Been Tested?
  
- [ ] Windows
- [x] macOS
- [ ] Linux
  
## Test Configuration

```
# System
| OS           | MacOS                                                                                                                       |
| Version      | 14.3                                                                                                                        |
| ID           | 23D56                                                                                                                       |
| Go Version   | go1.21.3                                                                                                                    |
| Platform     | darwin                                                                                                                      |
| Architecture | arm64                                                                                                                       |
| CPU          | Apple M1 Max                                                                                                                |
| GPU          | Chipset Model: Apple M1 Max Type: GPU Bus: Built-In Total Number of Cores: 32 Vendor: Apple (0x106b) Metal Support: Metal 3 |
| Memory       | 32GB                                                                                                                        |
```

# Checklist:

## Summary by CodeRabbit

- **New Features**
	- Introduced Drag & Drop support for macOS, enhancing file handling and interaction within the app.
- **Documentation**
	- Updated documentation to reflect the new Drag & Drop capabilities across all platforms and the modifications to the `DragAndDrop` option structure.


v2/internal/frontend/desktop/darwin/WailsContext.m
```diff
@@ -258,25 +258,28 @@ - (void) CreateWindow:(int)width :(int)height :(bool)frameless :(bool)resizable
                   forMainFrameOnly:false];
         [userContentController addUserScript:initScript];
     }
-    
-    self.webview = [WKWebView alloc];
+
+    self.webview = [WailsWebView alloc];
```
APshenkin, line 262: The important change is here. We use our extended implementation of WKWebkit here

leaanthony:
Thanks for this @APshenkin 🙏 I'd be keen to hear @pavelbinar and @lyimmi's thoughts. Awesome work man!

APshenkin:
@coderabbitai review

lyimmi:
Nice work @APshenkin!

I tested it on linux it works as expected.

On windows the `f.frontendOptions.DragAndDrop.Disable` check should be renamed to the new name in `v2/internal/frontend/desktop/windows/frontend.go:465`. After I replaced the check, it works on windows too.

Thanks for catching my mistakes in the JS files!

On the issue of loosing some functionality when D&D is set to disabled/enabled, I was thinking about that before and yes the cause of it is that all drop events are hidden from the webview.

The same thing happens when the `OnFileDrop` is registered on the JS side, you cannot drop to any type of input anymore. I think we could solve this by firing custom drop, or changed events based on the positions, but I'm not sure if its a good idea. I don't know how it would behave with the different frontend frameworks and that we could make a robust enough solution that works with all of them.

When someone wants to have the original behavior as well I think in that case they should handle the drop events manually to have a greater level of control over it. Maybe we should document this?

APshenkin:
@lyimmi thank you for the hint!. For some reason my IDE didn't replace it for Windows. Fixed

jakubpeleska:
Hi, guys. Thanks, @APshenkin and @leaanthony for a great job on the PR.

 I work with @pavelbinar and today, we were able to test it on MacOS, and it works well. 

The only issue I found is that the removal of the `wails-drop-target-active` class on `dragleave` doesn't work correctly if the target element has the size of the whole window. In that case, the class stays on the element even though `dragleave` was triggered. But that seems to be a general issue, not MacOS-specific.

lyimmi:
It seems that its a Javascript "behavior"...

We could use the "counter method" for this.

Add a new flag in `v2/internal/frontend/runtime/desktop/draganddrop.js`:
```javascript
const flags = {
    registered: false,
    defaultUseDropTarget: true,
    useDropTarget: true,
    prevElement: null,
    dragCount: 0, <-- this
};
```

Add a new listener:

```javascript
function onDragEnter(e) {
    if (!window.wails.flags.enableWailsDragAndDrop) {
        return;
    }
    e.preventDefault();

    if (!flags.useDropTarget) {
        return;
    }

    flags.dragCount++;
}

```

Zero the counter on drop:

```javascript
function onDrop(e) {
    if (!window.wails.flags.enableWailsDragAndDrop) {
        return;
    }
    e.preventDefault();

    if (!flags.useDropTarget) {
        return;
    }

    flags.dragCount = 0;
...
```

Decrement the counter on leave and clear the class when the counter is zero:

```javascript
function onDragLeave(e) {
    if (!window.wails.flags.enableWailsDragAndDrop) {
        return;
    }
    e.preventDefault();

    if (!flags.useDropTarget) {
        return;
    }
    flags.dragCount--;

    let targetElement = document.elementFromPoint(e.x, e.y);
    let cssDropValue = window.getComputedStyle(targetElement).getPropertyValue(window.wails.flags.cssDropProperty);
    if (cssDropValue) {
        cssDropValue = cssDropValue.trim();
    }
    if (flags.prevElement && (cssDropValue !== window.wails.flags.cssDropValue || flags.dragCount === 0))  {
        targetElement.classList.remove("wails-drop-target-active");
        flags.prevElement.classList.remove("wails-drop-target-active");
        // remove element as otherwise we will not update the class on the next dragover
        flags.prevElement = null;
    }
}
```

With this the class is reset correctly, I don't really like it but it works.
[Screencast from 2024-02-12 16:54:40.webm](https://github.com/wailsapp/wails/assets/8627125/80ac9abe-0cbc-4cd0-9aa8-aa44919cf113)

APshenkin:
@lyimmi @jakubpeleska 

I tried @lyimmi solution with macOS and for some reasons it didn't work.

So I came up with other way to solve it. Seems it works (at least on macOS 😅) and do this without counter which I think was the reason (as if something will register several times, then counter drop will not work).

Also I covered the case when you have nested elements in your `--wails-drop-target`. Before it didn't clean it properly I think.

 Can you guys check that it works on your side too?

https://github.com/wailsapp/wails/assets/14344430/895e6a13-5a43-4b81-b502-385f48abe372

jakubpeleska:
Hi everyone,
Thank you for the fixes. Overall, I believe the solution proposed by @APshenkin is superior. We have tested it and found it to work well; however, we identified one edge case where the behavior does not function as expected:
The removal of the `wails-drop-target-active` class on `dragleave` event does not work correctly if the target element is obscured by another window.
Please see the attached video for a demonstration of this issue.
https://github.com/wailsapp/wails/assets/48656252/02b1043a-4f4d-4112-8229-94c6b5cab0ea

pavelbinar:
Hi guys, I would suggest the following:

The primary goal of this PR is the low-level implementation of drag and drop - providing the URLs for the files and folders and registering the events for such actions.

The issues and implementation of the front-end are an addition. In our specific case, it is all we need, because we will still listen for the entire window (we do not need drop zones). Also, I can imagine other edge cases when you have multiple drop zones and you need to identify which one was used - this is not currently supported.

My point is:
Let's keep this lean and functional and leave the specific front-end implementation to each user (for now). This will allow us to ship this feature fast and at the same time not ship a half-baked, non-essential feature.

What do you think?

lyimmi:
I'm ok with just the events without the Javascript runtime part.

leaanthony:
Agree with @pavelbinar. Is there a way to separate the 2 concerns? Perhaps hooks? We could also add the ui part later as a flag or document it. Thoughts?

lyimmi:
@leaanthony it can be separated relatively simply.

We should register an event listener on the window for the drop for windows, that only handles the path resolution by calling `postMessageWithAdditionalObjects()`. Remove the `cssDropProperty` & `cssDropValue` from the app options.

Simplified to something like this, it could be moved to the main.js file under other listeners. This should be enough for windows.

```javascript

function onDrop(e) {
    if (!window.wails.flags.enableWailsDragAndDrop) {
        return;
    }
    e.preventDefault();

    if (window.chrome?.webview?.postMessageWithAdditionalObjects != null) {
        // process files
        let files = [];
        if (e.dataTransfer.items) {
            files = [...e.dataTransfer.items].map((item, i) => {
                if (item.kind === 'file') {
                    return item.getAsFile();
                }
            });
        } else {
            files = [...e.dataTransfer.files];
        }
        chrome.webview.postMessageWithAdditionalObjects(`file:drop:${e.x}:${e.y}`, files);
    }
}

window.addEventListener('drop', onDrop);

```

APshenkin:
@leaanthony @lyimmi 

maybe then merge this PR to the branch and then someone can pick up separation of JS part?

I'm not too familiar with Windows JS magic 😅

pavelbinar:
🚀

PR #3282, opened 2024-02-28
[v3] Add option for showing the toolbar in fullscreen mode on macOS

fbbdev:
# Description

When the MacTitleBar.UseToolbar option is enabled, the system toolbar is shown on macOS windows. Currently, when the windows are switched to fullscreen mode, the toolbar remains visible.

Although this might be desirable sometimes, in most common cases it is not the expected behaviour, as the full-screen toolbar cannot be made transparent and therefore clashes with the look of the application.

This patch introduces a new option in the `MacTitleBar` struct that when enabled hides the toolbar automatically upon entering fullscreen mode. It also enables the option by default in the predefined titlebar styles.

## Type of change

# How Has This Been Tested?
  
The new [`hide-macos-toolbar` example](https://github.com/fbbdev/wails/tree/feature/hide_fullscreen_toolbar/v3/examples/hide-macos-toolbar) creates two windows shocasing the two possible configurations of the new flag.

- [ ] Windows
- [x] macOS
- [ ] Linux
  
## Test Configuration

```
Go Version     | go1.22.0                                                                      
Modified       | true                                                                          
-buildmode     | exe                                                                           
-compiler      | gc                                                                            
CGO_CFLAGS     |                                                                               
CGO_CPPFLAGS   |                                                                               
CGO_CXXFLAGS   |                                                                               
CGO_ENABLED    | 1                                                                             
CGO_LDFLAGS    |                                                                               
GOAMD64        | v1                                                                            
GOARCH         | amd64                                                                         
GOOS           | darwin                                                                        
vcs            | git                                                                           
vcs.modified   | true                                                                          

# System
Name            | MacOS                                   
Version         | 12.6.6                                  
ID              | 21G646                                  
Branding        | Monterey                                
Platform        | darwin                                  
Architecture    | amd64                                   
Apple Silicon   | unknown                                 
Xcode cli tools | 2395                                    
CPU             | Unknown                                 
GPU             | Unknown                                 
Memory          | Unknown                                 

# Diagnosis
```

# Checklist:

- [x] I have updated `mkdocs-website/docs/en/changelog.md` with details of this PR

### DIFF
--- a/v3/pkg/application/webview_window_darwin.go
+++ b/v3/pkg/application/webview_window_darwin.go
@@ -443,18 +443,12 @@ void windowSetHideTitle(void* nsWindow, bool hideTitle) {
 }
 
 // Set Window use toolbar
-void windowSetUseToolbar(void* nsWindow, bool useToolbar, int toolbarStyle) {
+void windowSetUseToolbar(void* nsWindow, bool useToolbar) {
 	WebviewWindow* window = (WebviewWindow*)nsWindow;
 	if( useToolbar ) {
 		NSToolbar *toolbar = [[NSToolbar alloc] initWithIdentifier:@"wails.toolbar"];
 		[toolbar autorelease];
 		[window setToolbar:toolbar];
-
-		// If macos 11 or higher, set toolbar style
-		if (@available(macOS 11.0, *)) {
-			[window setToolbarStyle:toolbarStyle];
-		}
-
 	} else {
 		[window setToolbar:nil];
 	}
@@ -462,11 +456,15 @@ void windowSetUseToolbar(void* nsWindow, bool useToolbar, int toolbarStyle) {
 
 // Set window toolbar style
 void windowSetToolbarStyle(void* nsWindow, int style) {
+	WebviewWindow* window = (WebviewWindow*)nsWindow;
 	// use @available to check if the function is available
 	// if not, return
 	if (@available(macOS 11.0, *)) {
-		NSToolbar* toolbar = [(WebviewWindow*)nsWindow toolbar];
-		[toolbar setShowsBaselineSeparator:style];
+		NSToolbar* toolbar = [window toolbar];
+		if ( toolbar == nil ) {
+			return;
+		}
+		[window setToolbarStyle:style];
 	}
 }
 
@@ -477,6 +475,15 @@ void windowSetHideToolbarSeparator(void* nsWindow, bool hideSeparator) {
 		return;
 	}
 	[toolbar setShowsBaselineSeparator:!hideSeparator];
+}
+
+// Configure the toolbar auto-hide feature
+void windowSetShowToolbarWhenFullscreen(void* window, bool setting) {
+	WebviewWindow* nsWindow = (WebviewWindow*)window;
+	// Get delegate
+	WebviewWindowDelegate* delegate = (WebviewWindowDelegate*)[nsWindow delegate];
+	// Set height
+	delegate.showToolbarWhenFullscreen = setting;
 }
 
 // Set Window appearance type
@@ -745,10 +752,11 @@ void windowFocus(void *window) {
 */
 import "C"
 import (
-	"github.com/wailsapp/wails/v3/internal/assetserver"
-	"github.com/wailsapp/wails/v3/internal/runtime"
 	"sync"
 	"unsafe"
+
+	"github.com/wailsapp/wails/v3/internal/assetserver"
+	"github.com/wailsapp/wails/v3/internal/runtime"
 
 	"github.com/wailsapp/wails/v3/pkg/events"
 )
@@ -968,7 +976,6 @@ func (w *macosWebviewWindow) setEnabled(enabled bool) {
 }
 
 func (w *macosWebviewWindow) execJS(js string) {
-
 	InvokeAsync(func() {
 		if globalApplication.performingShutdown {
 			return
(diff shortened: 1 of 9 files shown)


v3/pkg/application/webview_window_darwin.go
```diff
@@ -443,30 +443,28 @@ void windowSetHideTitle(void* nsWindow, bool hideTitle) {
...
-			[window setToolbarStyle:toolbarStyle];
-		}
-
 	} else {
 		[window setToolbar:nil];
 	}
 }
 
 // Set window toolbar style
 void windowSetToolbarStyle(void* nsWindow, int style) {
```
fbbdev, line 458: The `windowSetToolbarStyle` had previously a wrong implementation and was unused. I changed it to the correct implementation and used it to replace the code that was previously at lines 452–457.

v3/pkg/application/webview_window_darwin.go
```diff
@@ -443,30 +443,28 @@ void windowSetHideTitle(void* nsWindow, bool hideTitle) {
...
 	// use @available to check if the function is available
 	// if not, return
 	if (@available(macOS 11.0, *)) {
-		NSToolbar* toolbar = [(WebviewWindow*)nsWindow toolbar];
-		[toolbar setShowsBaselineSeparator:style];
+		NSToolbar* toolbar = [window toolbar];
+		if ( toolbar == nil ) {
+			return;
+		}
+		[window setToolbarStyle:style];
```
fbbdev, line 467: The `windowSetToolbarStyle` had previously a wrong implementation and was unused. I changed it to the correct implementation and used it to replace the code that was previously at lines 452–457.

v3/pkg/application/webview_window_darwin.go
```diff
@@ -443,30 +443,28 @@ void windowSetHideTitle(void* nsWindow, bool hideTitle) {
 }
 
 // Set Window use toolbar
-void windowSetUseToolbar(void* nsWindow, bool useToolbar, int toolbarStyle) {
+void windowSetUseToolbar(void* nsWindow, bool useToolbar) {
```
fbbdev, line 446: The signature of this method has changed because the style configuration is now done through the `windowSetToolbarStyle` method, as per the comment at line 467.

v3/pkg/application/webview_window_darwin.go
```diff
@@ -1146,11 +1153,12 @@ func (w *macosWebviewWindow) run() {
 			C.windowSetHideTitleBar(w.nsWindow, C.bool(titleBarOptions.Hide))
 			C.windowSetHideTitle(w.nsWindow, C.bool(titleBarOptions.HideTitle))
 			C.windowSetFullSizeContent(w.nsWindow, C.bool(titleBarOptions.FullSizeContent))
-			if titleBarOptions.UseToolbar {
```
fbbdev, line 1149: This check has been removed as the involved methods work correctly in any case.

v3/pkg/application/webview_window_darwin.m
```diff
@@ -241,6 +241,9 @@ - (void)webView:(nonnull WKWebView *)webView stopURLSchemeTask:(nonnull id<WKURL
         }
     }
 }
+- (NSApplicationPresentationOptions)window:(NSWindow *)window willUseFullScreenPresentationOptions:(NSApplicationPresentationOptions)proposedOptions {
```
fbbdev, line 244: The main change is here: a new property on the delegate interface is used as a flag to either keep the default presentation options, or enable the `NSApplicationPresentationAutoHideToolbar` option, thus hiding the toolbar.

leaanthony:
Thanks for spending time on this and making such a comprehensive PR 🙏 As we currently have no API for adding anything to the toolbar, I'm wondering if this should be default behaviour? If later on, we have APIs to add elements to the toolbar, we could reintroduce this flag, but maybe invert it: `ShowToolbarWhenFullscreen`? Thoughts?

fbbdev:
Thank you for taking the time to review the PR!

As regards flipping the flag, I too had thought about it afterwards and completely second the proposal. This way, the original behaviour from v2 becomes the default and the change is completely transparent to the average consumer.

As regards removing the flag altogether, it makes no difference to me as I don't actually need it. However, I'd like to argue briefly in favour of keeping it:

  1. I feel like the general direction of v3 has been to give the developer greater control upon the appearance of the application, and this change would be in line with that trend;
  2. it is essentially zero-cost and transparent, especially in the `Show*` form;
  3. now that native window handles are available, someone might put them to use by filling the toolbar; after all, the option is called `UseToolbar`, not `InsetTitleBar` 😉
  4. and finally, that setting cannot be easily altered from outside through the native API, as it requires changing some logic in the window delegate, thus leaving no other option apart from fragile workarounds.

Let me know what you think about it, and afterwards I can edit the patch.

leaanthony:
You make some great points above and on the whole, I agree. Simply changing the Hide to Show might be enough for this PR. If you wanted to take it one step further, how about a method to set the value on the `Window` interface to allow setting it at runtime?  This would also require a change to allow JS to call it. You really don't have to and nobody has asked for it (yet) so I'll leave that with you.

fbbdev:
Ok, thanks! I am gonna edit the PR.

As regards adding the method, no other title bar parameter is reconfigurable atm, so it feels strange to have a method just for this one option. I won't go for it right now. As I read your message I thought I'd whip up a further PR for titlebar reconfiguration in general, but I keep getting second thoughts about it, as I just can't see the use case...

fbbdev:
Please note that I am gonna force-push here to keep things clean

leaanthony:
Great stuff! Thanks so much 👍

fbbdev:
Thank you!!

PR #3280, opened 2024-02-27
[v3] Export Window.UnFullscreen method from JS runtime module

fbbdev:
# Description

The Window.UnFullscreen method is available to the JS runtime, but is not exported. This patch fixes the problem.
The newly introduced [`window-api` example](https://github.com/fbbdev/wails/tree/fix-unfullscreen/v3/examples/window-api) (same as PR #3279) allows reproducing the problem and testing after the change: before the change, clicking the UnFullscreen button results in a JS exception being thrown.

## Type of change
  
Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?
  
I run the [`window-api` example](https://github.com/fbbdev/wails/tree/fix-unfullscreen/v3/examples/window-api) after the change and verified that the UnFullscreen button works as intended.

- [ ] Windows
- [x] macOS
- [ ] Linux
  
## Test Configuration

```Wails CLI      | v3.0.0-alpha.3                                                                
Go Version     | go1.22.0                                                                      
Revision       | 12d12a33dec1d31f487b70f87bc31e80252f327c                                      
Modified       | true                                                                          
-buildmode     | exe                                                                           
-compiler      | gc                                                                            
CGO_CFLAGS     |                                                                               
CGO_CPPFLAGS   |                                                                               
CGO_CXXFLAGS   |                                                                               
CGO_ENABLED    | 1                                                                             
CGO_LDFLAGS    |                                                                               
DefaultGODEBUG | httplaxcontentlength=1,httpmuxgo121=1,tls10server=1,tlsrsakex=1,tlsunsafeekm=1
GOAMD64        | v1                                                                            
GOARCH         | amd64                                                                         
GOOS           | darwin                                                                        
vcs            | git                                                                           
vcs.modified   | true                                                                          
vcs.revision   | 12d12a33dec1d31f487b70f87bc31e80252f327c                                      
vcs.time       | 2024-02-21T12:40:32Z                                                          

# System
Name            | MacOS                                   
Version         | 12.6.6                                  
ID              | 21G646                                  
Branding        | Monterey                                
Platform        | darwin                                  
Architecture    | amd64                                   
Apple Silicon   | unknown                                 
CPU             | Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
Xcode cli tools | 2395                                    
CPU             | Unknown                                 
GPU             | Unknown                                 
Memory          | Unknown                                 

# Diagnosis
[...]

PR #3295, opened 2024-03-05
[v3] Update and fix runtime JS API

fbbdev:
# Description

The main purpose of this PR is to bring the JS Window API up to date wrt the current WebviewWindow interface, as briefly discussed some time ago in the wails3 discord channel. In addition, it brings some bug fixes, code cleanups and quality-of-life improvements to the JS runtime. I decided to pack everything into a single PR because many changes overlap and I wanted to avoid a cascade of merge conflicts.

As a result this PR is somewhat large. In order to facilitate the review process, I divided it into many logically independent commits. I am gonna explain the changes in detail below through separate comments on a per-commit basis.

I am opening the PR against the `v3-alpha` branch. Should this PR be merged, I am willing to take care of any potential merge conflicts with the `v3-alpha-linux` branch.

Here follows a summary of the changes, some of which are **breaking**:

1. The [entry point](https://github.com/fbbdev/wails/blob/update_js_api/v3/internal/runtime/desktop/compiled/main.js) of the bundled runtime ([v3/internal/runtime/desktop/compiled/main.js](https://github.com/fbbdev/wails/blob/update_js_api/v3/internal/runtime/desktop/compiled/main.js)) has been refactored to avoid duplicating the export list.
2. **Breaking change.** To improve encapsulation, the `@wailsio/runtime` package does not export the API to the `window.wails` object anymore, and does not start the WML system. The bundled `runtime.js` still does both.
3. The [JS type](https://github.com/fbbdev/wails/blob/2dcf2733194bb6c90ce19d04da610030bbf189d6/v3/internal/runtime/desktop/%40wailsio/runtime/src/screens.js#L36) of the `Screen` struct has been updated to match its [Go counterpart](https://github.com/fbbdev/wails/blob/2dcf2733194bb6c90ce19d04da610030bbf189d6/v3/pkg/application/screen.go#L3).
4. The [`WebviewWindow.IsFocused`](https://github.com/fbbdev/wails/blob/2dcf2733194bb6c90ce19d04da610030bbf189d6/v3/pkg/application/webview_window.go#L583) method has been exposed on the [`Window` interface](https://github.com/fbbdev/wails/blob/2dcf2733194bb6c90ce19d04da610030bbf189d6/v3/pkg/application/window.go#L14).
5. **Breaking change.** The JS Window API has been updated to match the [`Window` interface](https://github.com/fbbdev/wails/blob/2dcf2733194bb6c90ce19d04da610030bbf189d6/v3/pkg/application/window.go#L14). This change is breaking because some methods have changed name and/or return type (although the most commonly used ones remain unchanged).
6. **Breaking change.** The Window API module (`@wailsio/runtime/src/window`) now exports the containing window object as a default export. It is not possible anymore to use ESM named import or namespace import syntax.
7. The method used to clean up previous listeners when reloading WML pages was wrong and has been fixed.
8. The WML system now supports multiple space-separated trigger events.
9. **Breaking change.** The bundled runtime is now an ES Module! It exports the same API as the `@wailsio/runtime` package. The benefits being that it can be easily imported into other modules and the ESM import system _ensures it is loaded at most once_. The change is breaking because the `type="module"` attribute must now be added to all script tags loading it.

As regards OS compatibility,
- on Windows, it shouldn't be a problem as the WebView2 library ships with very recent versions of the Chromium runtime;
- on macOS, I targeted Safari 10.1 shipping with macOS 10.12 Sierra;
- on Linux, I am not sure about the webkit version shipping there, but existing runtime code already required at least Safari 10 (because of arrow functions and the fetch API), so nothing should have changed.

Because of change no. 5, this PR supersedes #3280 

I am gonna push the updated changelog ASAP.

## Type of change

# How Has This Been Tested?

Before testing the PR code, I recommend running `go clean ./...` in the `v3` folder to ensure the build system picks up all changes.

Most examples have been brought up to date and thoroughly tested. The only exceptions are the `build` example (not relevant to the changes), the `dev` example (it is incomplete and probably not relevant), the `drag-n-drop` example (it has been updated but not tested as it doesn't work on macOS anyways).

The `menu` and `plugin` examples are not working with my test configuration, but the problems do not appear to originate with the new changes.

- [ ] Windows
- [x] macOS
- [ ] Linux

## Test Configuration

```
Go Version     | go1.22.0                                                                      
Modified       | true                                                                          
-buildmode     | exe                                                                           
-compiler      | gc                                                                            
CGO_CFLAGS     |                                                                               
CGO_CPPFLAGS   |                                                                               
CGO_CXXFLAGS   |                                                                               
CGO_ENABLED    | 1                                                                             
CGO_LDFLAGS    |                                                                               
GOAMD64        | v1                                                                            
GOARCH         | amd64                                                                         
GOOS           | darwin                                                                        
vcs            | git                                                                           
vcs.modified   | true                                                                          

# System
Name            | MacOS                                   
Version         | 12.6.6                                  
ID              | 21G646                                  
Branding        | Monterey                                
Platform        | darwin                                  
Architecture    | amd64                                   
Apple Silicon   | unknown                                 
Xcode cli tools | 2395                                    
CPU             | Unknown                                 
GPU             | Unknown                                 
Memory          | Unknown                                 

# Diagnosis
```

# Checklist:

- [x] I have updated the changelog with details of this PR
- [x] I have made corresponding changes to the documentation (N/A)


fbbdev:
### [Commit `0ab16ef`](https://github.com/wailsapp/wails/pull/3295/commits/0ab16ef45987edbd1cf1f9dc35ec83e1e39f6c0a) – Cleanup bundled runtime entry point

This commit wraps changes no. 1, 2 from the summary.

All code from the entry point of the `@wailsio/runtime` package (see [index.js](https://github.com/wailsapp/wails/blob/v3-alpha/v3/internal/runtime/desktop/%40wailsio/runtime/src/index.js)) was duplicated in the entry point of the runtime bundle (see [main.js](https://github.com/wailsapp/wails/blob/v3-alpha/v3/internal/runtime/desktop/compiled/main.js)), and the two had fallen slightly out of sync.

Keeping them aligned is burdensome for contributors who may not be aware of the existence of duplicated code, so instead of bringing everything up to date I used the inject option of the esbuild packager to eliminate the need for duplication altogether.

I also reworked things so that the '@wailsio/runtime' package does not populate the `window.wails` object and does not start the WML system. This improves encapsulation and reduces unexpected side effects for people importing the package into frontend frameworks. The bundled runtime script still does both things automatically.

I moved the `debugLog` and `whenReady` utility functions to a new file [`utils.js`](https://github.com/fbbdev/wails/blob/update_js_api/v3/internal/runtime/desktop/%40wailsio/runtime/src/utils.js) and created a new `WML.Boot` API method that people can call to start WML when using the npm package.

Going forward, you might consider moving all private code with side effects to an internal module to be either imported through an internal URL or automatically injected into internally served pages: it should include the code populating the global _wails object and all code related to clipboard, context menus, drag-n-drop, titlebar and RPC. Such a change would make all those functionalities available by default, and it would remove the hazard of conflicts between multiple instances of the runtime. It would also make exported bindings independent from the npm package.

fbbdev:
### [Commit bfdbe77](https://github.com/wailsapp/wails/pull/3295/commits/bfdbe7741acc784954338a3435255de23731c655) – Fix JS representation of Screen struct

This commit wraps change no. 3 from the summary.

The [`Screen.Rotation` field](https://github.com/wailsapp/wails/blob/4467a1ffa273059da10b45bbfd38c656f9e6dc28/v3/pkg/application/screen.go#L13) was changed to a `float32` in go code but was still [typed as an enum](https://github.com/wailsapp/wails/blob/4467a1ffa273059da10b45bbfd38c656f9e6dc28/v3/internal/runtime/desktop/%40wailsio/runtime/src/screens.js#L51) in JS code.

I verified the code that marshals the go struct to JSON, then fixed the JS type.

fbbdev:
[Commit `dfa0c83`](https://github.com/wailsapp/wails/pull/3295/commits/dfa0c83865c2686b44b4d83232159a3c88f6dfb9) – Expose IsFocused method in Window interface

This commit wraps change no. 4 from the summary.

Self-explanatory 🤷🏻‍♂️

leaanthony:
Really appreciate you taking this on 🙏 On the whole it looks fine. The reason we removed the runtime from being embedded by default was to give developers the flexibility to choose whether they wanted the entire runtime loaded and have the ability to pick and choose what they wanted from the npm module. I did consider doing this approach if it wasn't production but then that'll create inconsistent behaviour between prod and non-prod. In this PR, I believe both debug and non debug runtimes will be included in every build? My approach was going to create an asset handler that embeds the compiled runtimes that could be used for examples or anyone who wants the full bundle. If the compiled runtime is bundled each time then developers lose the benefit of the npm module. Thoughts?

fbbdev:
### [Commit `aad9b76`](https://github.com/wailsapp/wails/pull/3295/commits/aad9b7648f685722c771ddf4af570e1d59dc07eb) – Update JS window API

This commit wraps change no. 5 from the summary.

I copied (almost) all methods from the [`Window` interface](https://github.com/fbbdev/wails/blob/2dcf2733194bb6c90ce19d04da610030bbf189d6/v3/pkg/application/window.go#L14), assigned them serial IDs, rewrote the über-switch in [`messageprocessor_window.go`](https://github.com/fbbdev/wails/blob/update_js_api/v3/pkg/application/messageprocessor_window.go), copied the IDs to the [JS file](https://github.com/fbbdev/wails/blob/update_js_api/v3/internal/runtime/desktop/%40wailsio/runtime/src/window.js) and described all methods with JSDoc.

There are a few surprises here though.

#### 1. Some methods changed name

Specifically, `Screen` becomes `GetScreen`; `GetZoomLevel/SetZoomLevel` become `GetZoom/SetZoom` (I shouldn't have missed anything). This was of course not my choice as I just followed the current Window API.

#### 2. Some methods changed return type

Specifically, the getter methods `GetZoomLevel`, `Width` and `Height` (I might be missing something but I hope not) used to return JS objects with one field (`zoom`, `width` and `height` respectively). Since the `encoding/json` package can directly marshal numeric values, I changed the return type of all those methods to `number`, so that they better match the Go API.

It's not clear whether the previous approach was intentional, as their return types were not specified in JSDoc; see for reference:
- [`Width` method declaration](https://github.com/wailsapp/wails/blob/4467a1ffa273059da10b45bbfd38c656f9e6dc28/v3/internal/runtime/desktop/%40wailsio/runtime/src/window.js#L271)
- [`Height` method declaration](https://github.com/wailsapp/wails/blob/4467a1ffa273059da10b45bbfd38c656f9e6dc28/v3/internal/runtime/desktop/%40wailsio/runtime/src/window.js#L278)
- [`GetZoomLevel` method declaration](https://github.com/wailsapp/wails/blob/4467a1ffa273059da10b45bbfd38c656f9e6dc28/v3/internal/runtime/desktop/%40wailsio/runtime/src/window.js#L306)

#### 3. The methods now live on an unexported class `Window`

This replaces the previous anonymous object. I took care to bind all instance method so that they can be used easily in event handlers just like before.

#### 4. The module exports the `thisWindow` instance as a default export

See [`window.js`:643](https://github.com/fbbdev/wails/blob/b68b9c8216d8d1a318720f345ba8931c3ebf313c/v3/internal/runtime/desktop/%40wailsio/runtime/src/window.js#L643).

This was done to avoid duplication, as it is way too easy to have window objects fall out of sync with the module-exported API. Typing also improves. All methods of the default instance are available at `wails.Window.[Method]` just like before.

I do not consider this a breaking change because developers are usually supposed to import the whole package (hence `index.js`), not just the `window.js` file. Internal usage of course must change, for example at [`wml.js`:15](https://github.com/fbbdev/wails/blob/b68b9c8216d8d1a318720f345ba8931c3ebf313c/v3/internal/runtime/desktop/%40wailsio/runtime/src/wml.js#L15).

fbbdev:
> My approach was going to create an asset handler that embeds the compiled runtimes that could be used for examples or anyone who wants the full bundle. If the compiled runtime is bundled each time then developers lose the benefit of the npm module. Thoughts?

I get your point. Having a separate asset handler should mean that the optimizer can drop the embedded runtime when it's not used. I would be happy to work on this, but I can't see clearly what kind of API would you have in mind.

fbbdev:
BTW, what do you think of the npm package not populating the `window.wails` object by default?

fbbdev:
A further comment about the embedded runtime. In my [discussion of the related commit](https://github.com/wailsapp/wails/pull/3295#issuecomment-1979476816) I was proposing as a future change to embed and maybe even inject automatically the private runtime API (which is quite minimal), and leave the public API to the npm package, plus maybe an optional asset handler for examples and vanilla apps.

I think this would be the best approach overall, but it requires some more effort and I don't think it fits in this PR.

The asset handler you propose, on the other hand, would be a nice temporary solution.

fbbdev:
### [Commit `7b018ff`](https://github.com/wailsapp/wails/pull/3295/commits/7b018ff8d147d123443231ccb564c822b9673bb8) – Fix cleanup of WML listeners

This commit wraps changes no. 6 and 7 from the summary.

Change 7 (support multiple space-separated triggers) is very simple and is achieved by scanning the trigger attribute with a regexp (see [`wml.js`:210–212](https://github.com/fbbdev/wails/blob/b68b9c8216d8d1a318720f345ba8931c3ebf313c/v3/internal/runtime/desktop/%40wailsio/runtime/src/wml.js#L210)). The ugly while loop is necessary because the `String.matchAll` method was only added in the latest release of Safari and WebKit.

Change 6 is needed to fix a bug where calling `WML.Reload` would not clean up existing listeners. You can reproduce this as follows: run the WML example (before the PR) in debug mode, open the devtools, run `wails.WML.Reload()` in the console, then click the button labeled 'Emit an event' – the event will be emitted twice.

There are two problems with the existing code:

1. the major one is that `removeEventListener` is invoked with a closure that has just been declared (see e.g. [here](https://github.com/wailsapp/wails/blob/4467a1ffa273059da10b45bbfd38c656f9e6dc28/v3/internal/runtime/desktop/%40wailsio/runtime/src/wml.js#L45)). However, this new closure will always be a newly created object, distinct from the one used in the previous call to `addEventListener`;
2. the minor one is that `removeEventListener` is called with the current trigger attribute, which might have changed.

The solution is twofold as well:

1. use a [global, fixed event handler](https://github.com/fbbdev/wails/blob/b68b9c8216d8d1a318720f345ba8931c3ebf313c/v3/internal/runtime/desktop/%40wailsio/runtime/src/wml.js#L56) for everything, and decide on the fly which actions to perform (this incidentally reduces the need for calls to `WML.Reload`);
2. store the active trigger events for each element.

As regards the latter part of the solution, HTMX does it in the dumbest possible way to maximise browser compatibility, i.e. it defines a custom named property (not an attribute!) on each targeted node. In wails' case, however, the constraints are weaker as it supports way less browsers and it always runs global updates (whereas HTMX runs partial updates too).

For very recent browsers, I use the [`AbortController`/`AbortSignal` approach](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#signal) which allows removing ALL associated event listeners at once (see [`wml.js`:97–131](https://github.com/fbbdev/wails/blob/b68b9c8216d8d1a318720f345ba8931c3ebf313c/v3/internal/runtime/desktop/%40wailsio/runtime/src/wml.js#L97)). Note that this approach does not support partial updates.

For older browsers, I use a [`WeakMap`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap) to associate a trigger list to each element (see [`wml.js`:138–195](https://github.com/fbbdev/wails/blob/b68b9c8216d8d1a318720f345ba8931c3ebf313c/v3/internal/runtime/desktop/%40wailsio/runtime/src/wml.js#L138)). This is better than a Map or array as it doesn't stop garbage collection of mapped elements. When performing a reload, the whole document tree is visited and listeners are removed if present (this is similar to what HTMX does, unfortunately WeakMaps cannot be iterated over). As an optimisation, the loop terminates as soon as the count of active elements reaches zero. This kicks in especially when running `Reload` for the first time.

The two approaches are wrapped in a uniform API and selected with a feature test (see [`wml.js`:197](https://github.com/fbbdev/wails/blob/b68b9c8216d8d1a318720f345ba8931c3ebf313c/v3/internal/runtime/desktop/%40wailsio/runtime/src/wml.js#L197) and [`utils.js`:31–44](https://github.com/fbbdev/wails/blob/b68b9c8216d8d1a318720f345ba8931c3ebf313c/v3/internal/runtime/desktop/%40wailsio/runtime/src/utils.js#L31)).

fbbdev:
### [Commit `5b88fac`](https://github.com/wailsapp/wails/pull/3295/commits/5b88fac43a989c0542bed449d22e40d9b323786b) – Bundle runtime as ES module

This commit wraps changes no. 8 and 9 from the summary.

We already discussed the bundling part at length.

As regards the module part, I think it's nice to have, brings minimal disruption to existing code (just add `type="module"` attributes) and prevents initialisation code from running more than once.

fbbdev:
This is pretty much it, all other commits are just maintenance operations. It's a lot of stuff, don't worry and take your time to answer, and as always thank you for your work!

Summarising, apart from any comments you might have, I'd like to get your feedback about the following three points:
- Window module exports (see [here](https://github.com/wailsapp/wails/pull/3295#issuecomment-1979522599), last paragraphs)
- Runtime asset handler API (see [here](https://github.com/wailsapp/wails/pull/3295#issuecomment-1979538057) and [here](https://github.com/wailsapp/wails/pull/3295#issuecomment-1979556661), last paragraph of the edited comment)
- Whether the npm package should in your opinion export the API globally and/or automatically start the WML subsystem

stffabi:
> * Runtime asset handler API (see [here](https://github.com/wailsapp/wails/pull/3295#issuecomment-1979538057) and [here](https://github.com/wailsapp/wails/pull/3295#issuecomment-1979556661), last paragraph of the edited comment)

Just as a reference if you probably start working on that one, all code is already there for injection in v2: https://github.com/wailsapp/wails/blob/c5381dce704a18e26a467c5a706a001593638af1/v2/pkg/assetserver/assetserver.go#L140-L181
So you don't have to start all over :)

v3/internal/runtime/desktop/@wailsio/runtime/src/wml.js
```diff
@@ -13,159 +23,214 @@ import {OpenURL} from "./browser";
...
-    let result = new Map();
-
-    // Iterate over all properties of the window object
-    for (let method in targetWindow) {
-        // Check if the property is indeed a method (function)
-        if(typeof targetWindow[method] === 'function') {
-            // Add the method to the map
-            result.set(method, targetWindow[method]);
-        }
+export function Boot() {
```
leaanthony, line 224: What do you think about calling this `Enable`?

leaanthony:
Thanks for taking the time to go through all this! On the whole I think this is a great step forward. I'll be moving the compiled runtime into its own FS handler so it can be better used rather than having copies of `runtime.js` everywhere - but that's for a different PR!

leaanthony:
@fbbdev - apologies! As part of fixing the merge conflicts I've appeared to created a bad merge. I did push up a reset but that does not appear to have worked as intended. Could you please force push your branch again? Thanks 😬

leaanthony (approved):
This is incredible! Thank you 🙏

mkdocs-website/docs/en/changelog.md
```diff
@@ -54,13 +54,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-<<<<<<< HEAD
-=======
```
fbbdev: Whooops 😆

leaanthony: All good 😂

PR #3344, opened 2024-03-26
Draft: Qt v3

Blquinn:
# Description

Draft PR to add qt support for wails v3. Not remotely ready obviously just here for visibility.


leaanthony:
@Blquinn - I think this is cool but I'll close It for now as it's something we need to discuss post v3. Many thanks 🙏

PR #3406, opened 2024-04-18
allow for non-Window WailsEvent listeners

tmclane:
- adds a RegisterListener function on the App struct such that code can listen for WailsEvent(s) even if it isn't a Window

An assumption is made such that Window(s) are implicit listeners and therefore they also are always delivered all WailsEvent(s).

The intent here is to allow for non-Window listeners to be able to register for these events without being a Window itself opening up the ability of plugins or other code that wishes to receive **ALL** `WailsEvent(s)` such that they can be processed.

### DIFF
--- a/v3/pkg/application/application.go
+++ b/v3/pkg/application/application.go
@@ -789,7 +789,7 @@ func (a *App) dispatchEventToListeners(event *WailsEvent) {
 	listeners := a.wailsEventListeners
 
 	for _, window := range a.windows {
-		listeners = append(listeners, window)
+		window.DispatchWailsEvent(event)
 	}
 
 	for _, listener := range listeners {
--- a/v3/pkg/application/application.go
+++ b/v3/pkg/application/application.go
@@ -76,7 +76,7 @@ func New(appOptions Options) *App {
 	result.logStartup()
 	result.logPlatformInfo()
 
-	result.Events = NewWailsEventProcessor(result.dispatchEventToWindows)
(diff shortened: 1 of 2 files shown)


leaanthony (approved):

leaanthony (changes requested):
Let's use an interface for the listeners so it can handle both window and plugins

leaanthony:
I need to uninstall GitHub mobile because it's way hard to properly review using it 🤣😅

v3/pkg/application/application.go
```diff
@@ -774,9 +785,15 @@ func SaveFileDialogWithOptions(s *SaveFileDialogOptions) *SaveFileDialogStruct {
 	return result
 }
 
-func (a *App) dispatchEventToWindows(event *WailsEvent) {
+func (a *App) dispatchEventToListeners(event *WailsEvent) {
+	listeners := a.wailsEventListeners
+
 	for _, window := range a.windows {
-		window.DispatchWailsEvent(event)
+		listeners = append(listeners, window)
```
leaanthony, line 792:
```suggestion
		window.DispatchWailsEvent(event)
```

leaanthony: Either we add the windows to the listeners when they are created or just dispatch to them explicitly. Given that windows can come and go, it's probably better to do that the original way (explicitly).

leaanthony (approved):

Issue #3409, opened 2024-04-19
Please add Goland support


Czyhandsome:
### Describe the solution you'd like

Please add Goland support, now that Wails have vscode support now.

leaanthony:
`-ide goland`
I notice the docs need updating. Thanks 👍

jordanamr:
This can now be closed

leaanthony:
Not currently.

mkirl:
Can I have support for both IDEs? Also if i have an existing project can i just add the files that the ide flag generates work in order to get that support

leaanthony:
There's no plans to add support for both in the same project but you're welcome to create a PR for it. Copying files is not a solution as they have project specific values in them 👍

PR #3416, opened 2024-04-21
docs: add goland option in IDE section

DiegPS:
# Description

Add the possibility for anyone to know that goland can also be used with the `-ide goland` by a brief mention in the documentation

Fixes # (https://github.com/wailsapp/wails/issues/3409)

## Type of change


leaanthony (changes requested):
Could you please add this (and your other PR) to the un-versioned documentation so we don't lose it after a version upgrade please 🙏

PR #3457, opened 2024-05-05
Improved window parameter handling

leaanthony:
Original PR: https://github.com/wailsapp/wails/pull/3424

# Description

** This PR has been opened up to start a discussion on what the right solution is **

There will be times when you want to know what window made a call to a bound method. This PR offers 2 solutions for 2 different scenarios:

1. When a bound method uses a standard `context.Context` in the first parameter, the context will have a value "window" which will return the calling window.
2. For scenarios where you want to call these methods from Go, you can simply use an `application.Window` as the first parameter. It will always be populated with the calling window if called from JS. 

The `binding` example has been updated to show both usages.


leaanthony:
Closing in favour of a simpler PR

PR #3467, opened 2024-05-08
V3 feature/webkit2 40 41

tmclane:
V3 update to support webkit2 4.0 and 4.1.
The default in this case is 4.1.

Tags to build are identical to the ones used in v2 AFAIK.

`webkit2_40 == 4.0 and soup2`
`webkit2_41 == 4.1 and soup3`


leaanthony:
@tmclane - closing this as we'll exclusively be targeting 4.1 in v3 👍

PR #3469, opened 2024-05-10
fix broken links in the howdoesitwork page

oguz-yilmaz:
In this https://wails.io/docs/howdoesitwork page there are two links that doesn't work and basically redirects to the 404 page. The weird thing is that those links locally working fine but not on the actual website. I checked the other links that references the same options page and recognized that the only difference is that those working links uses the file extensions so I added those to the missing links hoping that would fix those broken links. I am not experienced in docusaurus in anyway, btw so my assumptions might be wrong.


leaanthony:
Ah nice catch! Yes, I've seen this before. I appreciate you taking the time to do this. 
Please could you add an entry to the changelog located at `website/src/pages/changelog.mdx`? Thank you 🙏

leaanthony (approved):

leaanthony:
Thanks again! 👍

PR #3480, opened 2024-05-14
Use NewSingleHostReverseProxy

leaanthony:
# Description

This PR alters how the `NewProxyServer` method is implemented under the hood. It should still function as normal. This was done to revert the way default middleware would work on non-mac platforms.


stffabi (approved):
LGTM 👍

PR #3570, opened 2024-06-26
[V3] Plugins implemenations

atterpac:
Adds plugins as a service as discussed in #3496 
See discussion for more information until PR is finalized


v3/internal/assetserver/assetserver.go
```diff
@@ -1,22 +1,19 @@
...
-	"path"
 	"strings"
 	"time"
 )
 
 const (
 	webViewRequestHeaderWindowId   = "x-wails-window-id"
 	webViewRequestHeaderWindowName = "x-wails-window-name"
-	pluginPrefix                   = "/wails/plugin"
+	servicePrefix                   = "/services"
```
leaanthony, line 16: I think we should keep "reserved" paths to a `/wails/` prefix. Thoughts?

atterpac: I have no issue with that

atterpac: Resolved in latest commit

leaanthony:
@atterpac - any chance we could send those errors via the new error handlers? 🙏

atterpac:
> @atterpac - any chance we could send those errors via the new error handlers? 🙏

Do we want to consider a plugin error as fatal?

leaanthony (approved):

Issue #3690, opened 2024-08-19
Issue with setFrameless Function Not Enabling Window Frame on First Call


bruxaodev:
### Description

I encountered an issue with the setFrameless function when trying to toggle the window frame on and off. Specifically, when a window is opened in frameless mode and I attempt to add the frame back using `setFrameless(false)`, the frame does not appear on the first call. I need to disable and then re-enable the frame for it to be applied correctly.

### To Reproduce

1. Open a window in frameless mode by calling `setFrameless(true)`.
2. Attempt to add the frame back by calling `setFrameless(false)`.
3. Observe that the window frame does not appear.

### Expected behaviour

The window frame should be applied on the first call to `setFrameless(false)`.

### Screenshots

![overlay-bug](https://github.com/user-attachments/assets/61535152-a914-4714-9659-68ff616d54ac)

### System Details

```shell
# System
┌─────────────────────────────────────────────────────────────────────────────────────────────┐
| Name              | Windows 10 Pro                                                          |
| Version           | 2009 (Build: 22621)                                                     |
| ID                | 22H2                                                                    |
| Branding          | Windows 11 Pro                                                          |
| Platform          | windows                                                                 |
| Architecture      | amd64                                                                   |
| Go WebView2Loader | true                                                                    |
| WebView2 Version  | 127.0.2651.105                                                          |
| CPU               | Intel(R) Xeon(R) CPU E5-2670 v3 @ 2.30GHz                               |
| GPU 1             | Meta Virtual Monitor (Meta Inc.) - Driver: 17.12.55.198                 |
| GPU 2             | Virtual Desktop Monitor (Virtual Desktop, Inc.) - Driver: 10.54.50.446  |
| GPU 3             | NVIDIA GeForce RTX 3060 (NVIDIA) - Driver: 32.0.15.5599                 |
| Memory            | 32GB                                                                    |
└─────────────────────────────────────────────────────────────────────────────────────────────┘

# Build Environment
┌─────────────────────────────────────────────────────────┐
| Wails CLI    | v3.0.0-alpha.6                           |
| Go Version   | go1.22.5                                 |
| Revision     | c79461b481fc2463ee21d96b316e53168248fd7c |
| Modified     | true                                     |
| -buildmode   | exe                                      |
| -compiler    | gc                                       |
| CGO_CFLAGS   |                                          |
| CGO_CPPFLAGS |                                          |
| CGO_CXXFLAGS |                                          |
| CGO_ENABLED  | 1                                        |
[...]

PR #3691, opened 2024-08-19
Fix: Ensure Window Frame is Applied on First Call in setFrameless Function

bruxaodev:
This pull request addresses the issue where the window frame does not appear on the first call to `setFrameless(false)` after being opened in frameless mode. The problem was that `SetWindowLong ` alone did not immediately apply the style change, requiring a second call to update the window correctly.

Issue Link: [Issue #3690](https://github.com/wailsapp/wails/issues/3690)

#Changes:

- Added a call to `SetWindowPos ` with the `SWP_FRAMECHANGED` flag after setting the window style. This forces the window to apply the style changes immediately.

#Testing:

- Verified that the frame is correctly applied on the first call to setFrameless(false).
- Tested both frameless and framed modes to ensure no regression in behavior.

overlay-fix

## Summary by CodeRabbit

- **New Features**
	- Enhanced window management functionality to ensure visual updates of window frame styles are reflected immediately.

- **Bug Fixes**
	- Improved behavior of frameless window display, ensuring frame changes are applied correctly on the first call.
	
- **Documentation**
	- Updated changelog for versioning documentation, correcting formatting and adding details about the frameless function fix on Windows.

### DIFF
--- a/mkdocs-website/docs/en/changelog.md
+++ b/mkdocs-website/docs/en/changelog.md
@@ -19,8 +19,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## v3.0.0-alpha.6 - 2024-07-30
 
-### Fixed 
-- Module issues 
+### Fixed
+- Module issues
 
 ## v3.0.0-alpha.5 - 2024-07-30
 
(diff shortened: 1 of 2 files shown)


leaanthony:
Thanks!

PR #3888, opened 2024-11-17
V3 alpha - calling window id and name

leaanthony:
# Description

When a bound call uses `context.Context` the calling window (if called from the frontend) ID and Name can be retrieved using context keys:

```go
func (s *MyService) WindowAwareMethod(ctx context.Context) (string, error) {
    windowName := ctx.Value(application.WindowNameKey).(string)
    windowID := ctx.Value(application.WindowIDKey).(string)
    return fmt.Sprintf("Called from window: %s (ID: %s)", windowName, windowID), nil
}
```

Also addresses an unchecked error.


dav3yblaz3 (approved):
great job team lets get this completed and keep everything going dev needs time to focus on developing

PR #4161, opened 2025-03-24
[v3] Typed events

fbbdev:
# Description

This PR enhances the v3 binding generator with support for registered events with strict data typing. Typescript IDEs will provide autocompletion support in JS/TS files for registered events. Data types will be checked at build time on the frontend, at runtime on the Go side.

🚨 **This PR includes one breaking change:** `EmitEvent` methods (on App and Window structs) won't wrap event data in a slice anymore _when at most one data argument is provided._

The change is intended to make `EmitEvent`'s behaviour more intuitive for the most common case (one data argument) and align it with the JS runtime.

The impact is mitigated by the fact that all runtime type assertions expecting a slice in event data will break, thus alerting developers to the change.

Many thanks to @IanVS who helped with the template updates.

## Type of change
  
Please select the option that is relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?
  
Examples and automated regression tests have been run and verified to be working correctly. The updated templates have been tested manually.

- [ ] Windows
- [x] macOS
- [ ] Linux

## Test Configuration

```
 Wails (v3.0.0-dev)  Wails Doctor

# System

┌──────────────────────────────────────────────────────────────────────────┐
| Name          | MacOS                                                    |
| Version       | 12.6.6                                                   |
| ID            | 21G646                                                   |
| Branding      | Monterey                                                 |
| Platform      | darwin                                                   |
| Architecture  | amd64                                                    |
| Apple Silicon | unknown                                                  |
| CPU           | Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz                 |
| CPU           | Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz                 |
| GPU           |  cores, Metal Family: Supported, Metal GPUFamily macOS 2 |
|               |       Metal Family: Supported, Metal GPUFamily macOS 2   |
| Memory        | 32 GB                                                    |
└──────────────────────────────────────────────────────────────────────────┘

# Build Environment

┌─────────────────────────────────────────────────────────┐
| Wails CLI    | v3.0.0-dev                               |
| Go Version   | go1.24.0                                 |
| Revision     | 5c6d8a9086acf1b25224632ae5fcbe193906e1a0 |
| Modified     | false                                    |
| -buildmode   | exe                                      |
[...]

PR #4181, opened 2025-03-31
[v3] don't wrap events in slice

tmclane:
This single commit removes the `variadic` operator from the input to `EmitEvent`.
Using a variadic operator for 'data' causes the `event` object to be wrapped in a single element slice which forces the consumer of the event to operate on a slice of messages even if only a single non-slice object was submitted to the function. 

If the `variadic operator` is intended it would seem that an iteration over the elements might be required and to send multiple events in that case.

Looking for clarification on desired functionality here if this is the wrong solution.

### DIFF
--- a/v3/pkg/application/application.go
+++ b/v3/pkg/application/application.go
@@ -381,7 +381,7 @@ func (a *App) RegisterService(service Service) {
 }
 
 // EmitEvent will emit an event
-func (a *App) EmitEvent(name string, data ...any) {
+func (a *App) EmitEvent(name string, data any) {
 	a.customEventProcessor.Emit(&CustomEvent{
 		Name: name,
 		Data: data,


fbbdev:
Hi! I included a similar change in my PR #4161 but an ensuing discussion on discord left me wondering whether it would make for poor DX.

The unexpected slice is definitely bad though, plus it doesn't match the behaviour of the JS runtime, for added confusion.

I was thinking of pushing an update to my PR proposing to emit multiple events as you too suggest, but that would break people's existing code silently as opposed to compile time errors.. not an easy tradeoff.

fbbdev:
Now that I think of it, emitting multiple events will still cause runtime errors when the data field is asserted to a slice type, so either approach should be safe.

tmclane:
I definitely do not like passing a structure and getting a slice of object out on the JS side.
Really non-intuitive.

fbbdev:
I think I found a better solution than both ones discussed above:

- if no arguments are provided, use `nil` data
- if one argument is provided, use it directly as event data
- if more than one argument is provided, use the slice as event data

This approach fixes the counterintuitive behaviour in the most common case (one argument) without breaking things for people who might be passing in multiple arguments (I expect there to be a smallish but non-zero number of them). At the same time, it avoids the potentially confusing behaviour of emitting multiple events (one per argument) from a method called `EmitEvent`.

I included the change in #4161 but I do not consider that PR ready for merging. Feel free to copy the implementation from there if you like the approach.

tmclane:
I would argue the most intuitive thing to do with this code is to send exactly what the caller provided.

IE if they provided a nil send the nil.
If it's a struct then send a struct/object.
If it's a slice then send a slice/array.

This is the principle of least surprise, the same object in the same object out to the best of our ability.

fbbdev:
I agree with you in principle.

There were complaints on Discord about not being able to omit a `nil` arg, and @leaanthony wrote:

> EmitEvent's data has always been optional

My proposal was meant to address those complaints.

I do believe that
```go
app.EmitEvent("myevent", nil)
```
is a perfectly acceptable level of boilerplate, the simplest option and therefore The Go Way™.

tmclane:
Yes that is what I was meaning.  Send whatever argument given.  If they want a no argument event we could have one more method for that.  

EmitEmptyEvent to make it obvious

tmclane:
Wait is that the point of the variadic?  To allow it to not be specified at all?

Guess that makes it hard to check for not provided versus actually sending nil.

tmclane:
I just want it to not become a slice inadvertently.

fbbdev:
> Wait is that the point of the variadic?  To allow it to not be specified at all?

Just syntactic sugar for passing `nil`. A shorthand, for convenience.

> Guess that makes it hard to check for not provided versus actually sending nil.

Given that the `nil` interface is not the same as a _typed_ `nil` (which translates to a _non-`nil`_ interface value), I think it can safely fulfill the role of "not provided" (especially once support for strictly typed events gets merged)

leaanthony:
Hey! 👋

Doing some housekeeping on older PRs and circling back on this one.

I totally get the reasoning - the implicit array wrapping from the variadic signature is a bit surprising on the JS side. But I think we're too far down the line to change this now. It would be a breaking change that affects everyone's existing event handlers, and v3 has been in alpha long enough that people have built around the current behavior.

For anyone who finds the array wrapper awkward, destructuring on the JS side keeps it clean:
```javascript
wails.Events.On("user:update", ([user]) => {
    // works nicely
})
```

Thanks for thinking about the DX and putting this together - always appreciated! 🙏

PR #4207, opened 2025-04-09
Prevent DLL hijacking by setting default DLL directories on initialization

ansxuman:
# Description

This PR fixes a DLL hijacking vulnerability identified during a VAPT assessment. When the application tries to load certain DLLs from the application directory instead of the system directories, it creates a potential security risk where an attacker could place malicious DLLs in the application directory.

The issue affects several DLLs including:
- uxtheme.dll
- WindowsCodecs.dll
- dwmapi.dll
- PROPSYS.dll
- CRYPTBASE.dll
- Secur32.dll
- SSPICLI.DLL
- MLANG.dll
- WININET.dll

While most DLLs can be fixed by adding `windows.SetDefaultDllDirectories(windows.LOAD_LIBRARY_SEARCH_SYSTEM32)` in application code, the `uxtheme.dll` is loaded by the Wails before application initialization. This PR addresses this issue by adding the secure DLL loading directive at the framework level.

# Additional Background Information
Our security team recently discovered a DLL hijacking vulnerability in our application. After digging into it, I found that this is actually a common issue in Go applications - there's even a similar bug report in the Go repository (https://github.com/golang/go/issues/64411).
Here's what's happening: When Windows looks for DLLs, it first checks the application directory before checking system directories. This creates a security risk where someone could drop malicious versions of these DLLs into our application folder.
I initially tried fixing this in our code by adding `windows.SetDefaultDllDirectories(windows.LOAD_LIBRARY_SEARCH_SYSTEM32)`, which tells Windows to prioritize loading DLLs from the System32 directory. This worked for most DLLs, but not for `uxtheme.dll`, it loaded by Wails before application initialization, so our fix never gets a chance to take effect.It's an important security improvement that protects applications from a potentially serious attack vector where bad actors could execute arbitrary code through fake DLLs.

## Type of change
  
## Steps to Reproduce

1. Create a basic Wails application
2. Build the application for Windows
3. Run Process Monitor while launching the application and Filter the PATH to `<Application Directory>\uxtheme.dll` and observe

# How Has This Been Tested?
  
- [x] Windows
- [ ] macOS
- [ ] Linux
      
## Test Configuration

```
PS C:\Users\DELL\wails\v2\cmd\wails> .\wails.exe doctor       

# Wails
Version  | v2.10.1
Revision | 194b0586ba63b3003bb6f5c3b5bc48f8c84720fc
Modified | true


# System
┌─────────────────────────────────────────────────────────────────────────────────────┐
| OS           | Windows 10 Pro                                                       |
| Version      | 2009 (Build: 19045)                                                  |
| ID           | 22H2                                                                 |
| Branding     | Windows 10 Pro                                                       |
| Go Version   | go1.23.0                                                             |
| Platform     | windows                                                              |
| Architecture | amd64                                                                |
| CPU          | Intel(R) Core(TM) i7-7600U CPU @ 2.80GHz                             |
| GPU          | Intel(R) HD Graphics 620 (Intel Corporation) - Driver: 31.0.101.2127 |
| Memory       | 16GB                                                                 |
└─────────────────────────────────────────────────────────────────────────────────────┘

# Dependencies
| WebView2   | N/A          | Installed | 134.0.3124.93 |
| Nodejs     | N/A          | Installed | 22.14.0       |
| npm        | N/A          | Installed | 10.9.2        |
| *nsis      | N/A          | Available |               |
|                                                       |

# Diagnosis

```

# Checklist:

## Summary by CodeRabbit

- **New Features**
  - Introduced a new option for Windows users to control DLL search paths, enhancing the application's configurability.
  - Expanded DLL loading settings with advanced controls, allowing for more detailed management of how DLL files are located on Windows platforms.
- **Improvements**
  - Updated function visibility for improved accessibility and initialization within the application.


leaanthony:
Thanks for taking the time to open this 🙏

I absolutely see the need for this but the solution will need to be more flexible to accommodate everyone's use cases. Some people ship DLLs with their binaries using NSIS and this will break their apps in a way they can't fix. 

Proposal: create a windows specific application setting that lists the set of search paths. Let's call it DLLSearchPaths. If this is blank, we do nothing and default to Go's defaults. If given, then we set them at application startup. This will mean updating most of the win32 package as you've noted. 

Thoughts?

ansxuman:
> Thanks for taking the time to open this 🙏
> 
> I absolutely see the need for this but the solution will need to be more flexible to accommodate everyone's use cases. Some people ship DLLs with their binaries using NSIS and this will break their apps in a way they can't fix.
> 
> Proposal: create a windows specific application setting that lists the set of search paths. Let's call it DLLSearchPaths. If this is blank, we do nothing and default to Go's defaults. If given, then we set them at application startup. This will mean updating most of the win32 package as you've noted.
> 
> Thoughts?

The approach you've suggested makes sense, but I have one question. In my project, I'm also using a few DLLs, and there hasn't been any impact after this change. The only difference is that I'm using absolute paths.

leaanthony:
Do you think we need to change `func init() {` in `uxtheme.go` to `func Init() {` and call this manually after setting the DLL search paths? Alternatively, you could try lazy loading the DLL but I'm guessing there's a reason this is being loaded differently.

ansxuman:
> Do you think we need to change `func init() {` in `uxtheme.go` to `func Init() {` and call this manually after setting the DLL search paths? Alternatively, you could try lazy loading the DLL but I'm guessing there's a reason this is being loaded differently.

Thanks for highlighting this. I didn’t get a chance to test the new changes as my Windows laptop is having some issues. I would prefer changing func init() to func Init() in both uxtheme.go and consts.go for a minimal change.

leaanthony:
Sounds good. Can you please make sure that the changes work correctly and they don't flicker windows when they start up or anything? That is my only concern with converting the init function.

ansxuman:
> Sounds good. Can you please make sure that the changes work correctly and they don't flicker windows when they start up or anything? That is my only concern with converting the init function.

Sure i will do proper testing and update you here.

ansxuman:
Tested on Windows 10, I did not observe any issues.

v2/internal/platform/win32/consts.go
```diff
@@ -80,7 +80,7 @@ ShouldSystemUseDarkMode = bool () // ordinal 138
 SetPreferredAppMode = PreferredAppMode (PreferredAppMode appMode) // ordinal 135, since 18334
 IsDarkModeAllowedForApp = bool () // ordinal 139
 */
-func init() {
+func Init() {
```
leaanthony, line 83: Did you mean to call these Init() methods?

ansxuman: Yes, I think It's not a semantic difference in the code, both are uppercase "I" in Init()  but rather a visual difference in how they're being rendered in the user interface.

ansxuman: Sorry for the confusion. I misunderstood Go's initialization behavior, thinking capitalization affected execution order of  Init() but relized it will behave like very other exported func. Now calling it explicitly after setting the dll path.

PR #4317, opened 2025-05-29
`window.ToggleFrameless() api`

atterpac:
Implements the `window.ToggleFrameless()` api internally it just uses the `window.setFrameless` toggling it as the opposite of the current value so Developer does not need to card about the current state to swap it

Changes:
- Runtime will need to be updated on dev machines as I reorganized the message processor values to keep the ToggleXXX in the same range
- Regenerated new events
- Updated `messageprocessor_window.go` 
- Updated runtime `window.ts` 
- Updated API example in `v3/examples/

## Summary by CodeRabbit

- **New Features**
  - Added the ability to toggle the window between frameless and normal states via a new API method and UI button in the frameless window example.

- **Documentation**
  - Updated the changelog to include the new window toggling feature.

- **Style**
  - Improved code readability with consistent multi-line formatting in several function signatures and calls.


leaanthony (approved):

v3/pkg/application/messageprocessor_window.go
```diff
@@ -108,7 +110,13 @@ var windowMethodNames = map[int]string{
 	WindowZoomReset:                  "ZoomReset",
 }
 
-func (m *MessageProcessor) processWindowMethod(method int, rw http.ResponseWriter, _ *http.Request, window Window, params QueryParams) {
+func (m *MessageProcessor) processWindowMethod(
+	method int,
```
leaanthony, line 114: What is your editor doing?! 😅

leaanthony (approved):

PR #4435, opened 2025-07-21
fix: prevent crash when closing window with pending async calls

### LINKED ISSUE #4353: [V3] macOS window close with pending async Go-bound function call crashes
### Description

Wails apps on macOS utilizing the `ApplicationShouldTerminateAfterLastWindowClosed: false` configuration option are susceptible to crashes when the user closes a window.

If a Go-bound async JS call is in-flight when the user closes the window, and the call completes shortly afterward, `webview_window_darwin.go` ends up calling `C.windowExecJS` to deliver the result, even though the underlying window has already been destroyed.

This leads to a crash due to use-after-free (invalid webView access).

### To Reproduce

1. `wails3 init -n myfirstapp`
2. In `main.go`, set `ApplicationShouldTerminateAfterLastWindowClosed: false`
3. Add 50ms of delay to `Greet` in `greetservice.go`
```go
func (g *GreetService) Greet(name string) string {
    time.Sleep(50 * time.Millisecond) // Simulate a delay
    return "Hello " + name + "!"
}
```
4. Add a loop calling `Greet` on `DOMContentLoaded` in `main.js`
```js
window.addEventListener("DOMContentLoaded", async () => {
    for (let i = 0; i < 100; i++) {
        await GreetService.Greet(`User ${i}`)
    }
});
```
5. `wails3 dev`
6.  As soon as the window opens, close it (via the window close button) within 5 seconds, while the Greet loop is still in progress.

This causes a crash with a segmentation fault in `windowExecJS`.

### Expected behaviour

The window should close cleanly regardless of in-flight async functions and the app should continue running.

### Screenshots

_No response_

### Attempted Fixes

_No response_

### System Details

```shell
wails3 doctor
 Wails (v3.0.0-alpha.9)  Wails Doctor 
                                                                                                                                                                                                     
# System 

| Version       | 15.5                             |
| ID            | 24F74                            |
| Branding      | MacOS 15.5                       |
| CPU           | Apple M3 Pro                     |
| CPU 1         | Apple M3 Pro                     |
| CPU 2         | Apple M3 Pro                     |
| GPU           | 14 cores, Metal Support: Metal 3 |
| Memory        | 18 GB                            |

# Build Environment 

| Wails CLI      | v3.0.0-alpha.9                                                                                                            |
| Go Version     | go1.23.5                                                                                                                  |
| -buildmode     | exe                                                                                                                       |
| -compiler      | gc                                                                                                                        |
| CGO_CFLAGS     |                                                                                                                           |
| CGO_CPPFLAGS   |                                                                                                                           |
| CGO_CXXFLAGS   |                                                                                                                           |
| CGO_ENABLED    | 1                                                                                                                         |
| CGO_LDFLAGS    |                                                                                                                           |
| DefaultGODEBUG | asynctimerchan=1,gotypesalias=0,httpservecontentkeepheaders=1,tls3des=1,tlskyber=0,x509keypairleaf=0,x509negativeserial=1 |
| GOARCH         | arm64                                                                                                                     |
| GOARM64        | v8.0                                                                                                                      |
| GOOS           | darwin                                                                                                                    |

# Dependencies 

| Xcode cli tools | 2409                                                 |

# Diagnosis 

```

### Additional context

Discussed in Discord with @leaanthony. I have a fix and will put up a PR.

leaanthony:
## Summary
This PR fixes a crash that occurs when closing a window while there are pending async Go-bound function calls.

## Root Cause
When a window is closed while async Go method calls are still pending, the application crashes because:
1. The `messageProcessor` was created but never stored as a field on the App struct
2. There was no mechanism to cancel pending async calls when a window is destroyed
3. The window could be destroyed while async operations were still trying to send responses back

## Fix
- Added `messageProcessor` field to the App struct to store the MessageProcessor instance
- Implemented `CancelWindowCalls` method in MessageProcessor to track and cancel pending calls per window
- Added call to `CancelWindowCalls` in the window's `destroy()` method
- Added safety checks to prevent sending responses to destroyed windows

## Testing
Updated the issue-4353 test and example to use the correct Wails v3 APIs (they were using v2-style APIs).

To test:
1. Run the example in `v3/examples/issue-4353/`
2. Click "Start 5-Second Async Operation"
3. Immediately click "Close Window"
4. The window should close without crashing

Fixes #4353

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved a crash that could occur when closing a window with pending asynchronous operations.

* **Improvements**
  * Enhanced cleanup and safety checks to ensure all pending asynchronous operations are properly cancelled when a window is closed.
  * Improved window lifecycle management to prevent operations on destroyed windows and ensure proper deregistration from the window manager.


leaanthony:
Reopening with unit-test coverage for the new `CancelWindowCalls` API — `v3/pkg/application/messageprocessor_test.go` (5 tests, all passing under `-race`):

- `TestMessageProcessor_CancelWindowCalls_CancelsAllForWindow` — every in-flight call registered for the closing window has its context cancelled and is removed from both `runningCalls` and `windowCalls`.
- `TestMessageProcessor_CancelWindowCalls_PreservesOtherWindows` — calls registered against a different window survive untouched.
- `TestMessageProcessor_CancelWindowCalls_UnknownWindowIsNoOp` — cancelling an unknown windowID neither panics nor disturbs unrelated state.
- `TestMessageProcessor_CancelWindowCalls_StaleEntryCleansUp` — guards the `if cancel, ok := m.runningCalls[callID]; ok` branch where a call completes and self-removes between `windowCalls` registration and the window destroy.
- `TestMessageProcessor_CancelWindowCalls_ConcurrentSafe` — concurrent register-then-cancel across multiple windows; verified under `go test -race`.

The tests live in `package application` and use a `registerCall` helper that mirrors the same internal bookkeeping `processCallMethod` performs, so they exercise the public surface against the production map shape.

Verified locally:
- `go test -run TestMessageProcessor_CancelWindowCalls -count=1 ./pkg/application/` → `ok`
- `go test -race -run TestMessageProcessor_CancelWindowCalls -count=1 ./pkg/application/` → `ok`

taliesin-ai (approved):
Approving — fixes a real crash when a window is closed while async calls are still in flight. The fix correctly guards the pending-call path on teardown. LGTM.

PR #4446, opened 2025-07-25
fix linux build error in onWayland caused by declaration after label

### LINKED ISSUE #4426: [v2.10.2] onWayland() causes compilation error on some compilers
### Description

A compilation error occurs in onWayland() function when building Wails v2.10.2 on some compilers.
The function declares a variable inside a case block without braces, which is not allowed in standard C.

Error message:

```
window.c:79:6: error: a label can only be part of a statement and a declaration is not a statement
   79 |      char *gdkBackend = getenv("XDG_SESSION_TYPE");
      |      ^~~~
```

### To Reproduce

Build the project for Linux target:

```
wails build -platform linux/amd64
```
Compilation fails with the above error.

### Expected behaviour

The code should compile successfully on any standard-compliant C compiler, even those that strictly enforce C rules.
Specifically, variables should not be declared directly inside a case label without a new scope or outside the switch statement.

### Screenshots

N/A — this is a compiler error and does not produce a UI screen.
Relevant build log snippet:

```
window.c: In function ‘onWayland’:
window.c:79:6: error: a label can only be part of a statement and a declaration is not a statement
   79 |      char *gdkBackend = getenv("XDG_SESSION_TYPE");
      |      ^~~~
ERROR   exit status 1
```

### Attempted Fixes

The root cause of the compilation error is that in C, declarations inside a case label must be inside a block scope (i.e., surrounded by {}), or be declared before the switch statement. This is because a case label is not a separate block by itself and declaring variables directly under it is not allowed by the C standard.

Original problematic code:

```
case -1:
    char *gdkBackend = getenv("XDG_SESSION_TYPE");
    if(gdkBackend != NULL && strcmp(gdkBackend, "wayland") == 0) 
    {
        wmIsWayland = 1;
        return TRUE;
    }

    wmIsWayland = 0;
    return FALSE;
```

Fix 1: Wrap in braces to create a new scope

```
case -1: {
    char *gdkBackend = getenv("XDG_SESSION_TYPE");
    if(gdkBackend != NULL && strcmp(gdkBackend, "wayland") == 0) 
    {
        wmIsWayland = 1;
        return TRUE;
    }

    wmIsWayland = 0;
    return FALSE;
}
```

Fix 2: Move declaration outside the switch (recommended for better clarity and compatibility)

```
char *gdkBackend = NULL;

switch (wmIsWayland)
{
case -1:
    gdkBackend = getenv("XDG_SESSION_TYPE");
    if(gdkBackend != NULL && strcmp(gdkBackend, "wayland") == 0) 
    {
        wmIsWayland = 1;
        return TRUE;
    }

    wmIsWayland = 0;
    return FALSE;
```
Both fixes solve the compilation error by respecting C language rules on declarations within switch-case statements.

### System Details

```shell
GOVERSION: go1.23.4 linux/amd64
```

### Additional context

The problematic file is [v2/internal/frontend/desktop/linux/window.c](https://github.com/wailsapp/wails/blob/v2.10.2/v2/internal/frontend/desktop/linux/window.c)

The error occurs during the Linux build step of wails build.

Likely to affect any environment where the C compiler enforces standard C rules (e.g., -std=c89 or -std=c99 without extensions).

Proposed fix: move char *gdkBackend declaration out of the case block or wrap it in braces.

jaesung9507:
# Description

On my development environment (Rocky Linux 8), cross-compilation to Windows works fine,  
but native Linux builds were failing due to a C compile error in the `onWayland` function. 

Fixes #4426 

## Type of change
  
# How Has This Been Tested?
  
- [ ] Windows
- [ ] macOS
- [x] Linux
      
Rocky 8
  
## Test Configuration
```
# Wails
Version         | v2.10.2
Package Manager | dnf    


# System
┌─────────────────────────────────────────────────────┐
| OS           | Rocky Linux                          |
| Version      | 8.10                                 |
| ID           | rocky                                |
| Branding     |                                      |
| Go Version   | go1.24.2                             |
| Platform     | linux                                |
| Architecture | amd64                                |
| CPU          | 13th Gen Intel(R) Core(TM) i5-13400F |
| GPU          | Unknown                              |
| Memory       | 32GB                                 |
└─────────────────────────────────────────────────────┘

# Dependencies
| Dependency | Package Name       | Status    | Version |
| *docker    | Unknown            | Not Found | 26.1.3  |
| gcc        | gcc-c++            | Installed | 8.5.0   |
| libgtk-3   | gtk3-devel         | Installed | 3.22.30 |
| libwebkit  | webkit2gtk3-devel  | Installed | 2.46.6  |
| npm        | npm                | Installed | 10.9.2  |
| pkg-config | pkgconf-pkg-config | Installed | 1.4.2   |
| *upx       | upx                | Available | 4.2.4   |
|                                                       |

```

# Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of the Wayland session type on Linux, ensuring more reliable behavior in desktop environments.
  * Fixed a Linux compilation issue related to session type detection, enhancing stability for Wayland users.


leaanthony:
Thanks 🙏 Please could you add an entry to the changelog located at `website/src/pages/changelog.mdx`?

leaanthony:
I'm struggling with why your C compiler complains about the -1 case but not the 1 case.

jaesung9507:
Thanks for the great question!

The difference is that case -1: has a variable declaration immediately after the label, while case 1: only contains executable statements.

In my development environment (GCC with default settings on Rocky Linux 8), the compiler doesn't default to C99 mode and instead follows the older C89 rules, which disallow declarations right after a label unless they're wrapped in a block.

leaanthony:
thanks for taking the time to fix this 🙏

negagen:
thanks for the fix man!

bluntelk:
Hey guys, any chance this could make it's way into a v2.10 series release?

(keep up the good work!)

PR #4576, opened 2025-09-10
Check for empty buffers instead of nil in responsewriter

jghiloni:
# Description

Use Case:

I am using a service that implements `http.Handler` to send streaming video to the frontend. Reliably, when sending fragmented MP4 data on MacOS, the app would panic because the buffer being sent through the response passed a not-nil check but would panic when accessing the first byte of the buffer.

This PR addresses that by checking if the buffer is not empty instead of nil, accounting for the case where len(buf) == 0 and cap(buf) > 0, or maybe where buf[0] == '\0' (I'm not sure how the nil checks work for slices)

Fixes # (issue)

## Type of change
  
# How Has This Been Tested?
  
- [ ] Windows
- [ ] macOS
- [ ] Linux
      
## Test Configuration

# Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a rare crash in AssetServer on macOS when handling responses with empty data, improving stability and reliability during content loading.
* **Documentation**
  * Updated the unreleased changelog to include this fix and credited the contributor.

### DIFF
--- a/v3/internal/assetserver/webview/responsewriter_darwin.go
+++ b/v3/internal/assetserver/webview/responsewriter_darwin.go
@@ -101,7 +101,7 @@ func (rw *responseWriter) Write(buf []byte) (int, error) {
 
 	var content unsafe.Pointer
 	var contentLen int
-	if buf != nil {
+	if len(buf) > 0 {
 		content = unsafe.Pointer(&buf[0])
 		contentLen = len(buf)
 	}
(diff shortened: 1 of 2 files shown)


v3/internal/assetserver/webview/responsewriter_darwin.go
```diff
@@ -101,7 +101,7 @@ func (rw *responseWriter) Write(buf []byte) (int, error) {
 
 	var content unsafe.Pointer
 	var contentLen int
-	if buf != nil {
+	if len(buf) > 0 {
```
atterpac: this should likely check for both nil and len since `buf` is able to be nil so this is more moving the panic to a different scenario

```suggestion
	if buf != nil && len(buf) > 0 {
```

atterpac (changes requested):

leaanthony (approved):

leaanthony:
Great catch! Thank you! 🙏

atterpac (approved):
lgtm

Issue #4697, opened 2025-11-09
Wails3 - install:frontend:deps hanging with npm link'ed packages.


Snider:
### Description

Using npm link to include a local copy of a binary crashes`wails build`, for example, pro versions you can't bundle with your source repo.

"@fontawesome": "file:~/Code/lib/fontawesome/npm" running npm install works fine, but wails gets an issue:

<img width="528" height="62" alt="Image" src="https://github.com/user-attachments/assets/fdacaa5c-0d6e-4ea3-a868-1b59d82fc8d9" />

### To Reproduce

use npm link or npm install with a local directory using `~/my-project`
wails3 fails to build

### Expected behaviour

The fix, is to run with -f, to tell go-task to handle cache checksum's differently.

### Screenshots

<img width="853" height="366" alt="Image" src="https://github.com/user-attachments/assets/d7d6fe6e-cbb8-4070-bcc4-6093d8c29222" />

### Attempted Fixes

The fix is to use the `node_modules/.bin` directory, so the change detection has a far smaller list of files to check; this has a side effect of speeding up the `common:install:frontend:deps` build step exponentially by the node_module dir size, as thousands of files are ignored.

### System Details

```shell
* my putter: cpu: beep booper 2000 *
```

### Additional context

PR incoming from: https://github.com/Snider/wails/tree/v3-alpha-fe-npm-gen-fix

leaanthony:
🤖 **Triage note**: This appears to be a v3 build system issue related to npm/taskfile caching. Missing label: `v3`. Platform could not be determined from the description — please confirm which OS.

PR #4698, opened 2025-11-09
fix: update Taskfile install:frontend:deps, check only .bin directory…

Snider:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

[Fixes # (4697)](https://github.com/wailsapp/wails/issues/4697)

## Type of change
  
Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?
  
wails builds my project / acts the same as npm install. Has the path change to the build template been tested? No, but the change has.

https://discord.com/channels/1042734330029547630/1065600636294549514/1435715828384268348
  
You would need to grab an npm module, install it in a project as a local file path in a project made with wails3 init, and notice how it doesn't segfault Goland.

## Summary by CodeRabbit

* **Chores**
  * Improved frontend build configuration to optimize dependency-change detection and reduce unnecessary rebuilds by narrowing the set of generated files considered.
* **Bug Fixes**
  * Fixed an issue where local npm packages could cause incorrect front-end rebuild behavior.
* **Documentation**
  * Added unreleased changelog entries describing the rebuild performance and fix.


Snider:
@coderabbitai the code walkthrough has an incorect statement.

```
Narrowed generated-files pattern from node_modules/** to node_modules/.bin/ for the install:frontend:deps task, reducing the scope of generated file tracking.
```
should be
```
Narrowed generated-files pattern from node_modules/* to node_modules/.bin/* for the install:frontend:deps task, reducing the scope of generated file tracking.
```

Which is the actual change.

Snider:
> 🧩 Analysis chain
> @Snider: You're absolutely correct! Thank you for catching that error.
> 
> Looking at the actual diff, the change was indeed from `node_modules/*` to `node_modules/.bin/*`, not from `node_modules/**` as stated in the walkthrough. The single asterisk (`*`) vs double asterisk (`**`) difference is important here - `*` matches files in the immediate directory, while `**` would match recursively through subdirectories.
> 
> Would you like me to regenerate the summary with the correct information?
> 
> ✅ Actions performed

yes please @coderabbitai

leaanthony (commented):
Did you mean to review the comments? We need to leave them in so new contributors know where to add things.

Snider:
> Did you mean to review the comments? We need to leave them in so new contributors know where to add things.

Oh, that's just me being a noob trying to be helpful, misreading a failing CI check ^.^ Was I meant to leave the file alone?

leaanthony:
Why do we only want to check the bin directory for module changes?

Snider:
> Why do we only want to check the bin directory for module changes?

Because using the level above makes Task checksum the entire node_modules folder to check for changes, which is slower the more you have, and totally stops compiling at all when using `npm link` with local packages; which you do when you buy fontawesome etc.

leaanthony:
Actually this should probably just be package.json and forget the rest 😅

Snider:
> Actually this should probably just be package.json and forget the rest 😅

Yeah, someone who uses Task more can probably improve the change;  so that "npm install" and "task install:frontend" (or whatever) act the same, because they don't right now, and it will cause hard-to-find issues for end users/developers later on.

PR #4840, opened 2025-12-30
Refactor(v3): macOS async notification methods

popaprozac:
# Description

Looking back at the implementation, especially for requesting macOS notification authorization it was done in a blocking way which doesn't make sense. We should register a callback so the user can authorize/deny the request async.

Updated the service name for imports.

Will be updating notifications docs and examples asap. I should probably update this in the v2 PR too 😅

## Type of change
  
# How Has This Been Tested?
  
- [ ] Windows
- [x] macOS
- [ ] Linux
      
## Test Configuration

# Checklist:

## Summary by CodeRabbit

* **Refactor**
  * Notification authorization methods converted to asynchronous, callback-based patterns across platforms; callbacks are nil-guarded and Windows/Linux invoke callbacks immediately.
* **Docs**
  * Notification docs updated with callback-based examples, flow guidance, and platform notes.
* **Changelog**
  * Recorded change: authorization methods switched to callbacks; service identifier updated to its new package path.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


popaprozac:
@coderabbitai review

leaanthony:
@popaprozac 
> Looking back at the implementation, especially for requesting macOS notification authorization it was done in a blocking way which doesn't make sense.  

What are the implications of the current method? Surely it's not much of a problem?

I asked Claude to do a quick review and picked up on a few things. Are you sure you want to go with this approach? It seems fraught with danger :sweat_smile: 

-----

Code Review Summary: PR #4840
PR: Refactor(v3): macOS async notification methods (https://github.com/wailsapp/wails/pull/4840)
Author: @popaprozac
Target: v3-alpha

---

What The PR Does
Changes RequestNotificationAuthorization and CheckNotificationAuthorization from synchronous return-based methods to callback-based methods:
// Before
`func RequestNotificationAuthorization() (bool, error)`
// After  
f`unc RequestNotificationAuthorization(callback func(bool, error))`
Stated rationale: The macOS authorization dialog is async, so blocking doesn't make sense.

---

Issues Identified
1. Frontend Bindings Will Break (Critical)
Wails bindings generate RPC calls between JavaScript and Go. The binding system requires methods to return values - it cannot handle callback parameters.
Current frontend code that will break:
```go
// v3/examples/notifications/frontend/src/main.ts
const authorized = await NotificationService.RequestNotificationAuthorization();
```
JavaScript cannot pass a Go function as a parameter. The generated TypeScript bindings expect a return value, not a callback.
2. Cross-Platform Behavior Inconsistency (Medium)

| Platform | Callback Timing |
|----------|-----------------|
| macOS | Async (new goroutine) |
| Windows/Linux/iOS | Sync (immediate, same goroutine) |
This inconsistency could cause subtle race conditions.

4. Potential Goroutine Leak (Medium)
The timeout was removed. If the Objective-C callback never fires (edge case, bug, app termination), the goroutine blocks forever:
```go
go func() {
    result := <-resultCh  // Blocks indefinitely if no response
    callback(result.Success, result.Error)
}()
```
5. Nil Callback Silently Skips Work (Low)
Passing nil silently does nothing - authorization is not requested. This breaks fire-and-forget patterns.

popaprozac:
The current method on macOS has a timeout before returning false if the user didn't interact with the notification to grant authorization. Having a callback ensures **_anytime_** the user decides to grant notification, even after letting it sit in their notification center for 2 hours, we get a signal when that happens. You could build a habit of _always_ requesting authorization before running any notification code but this seems less idiomatic for the platform. macOS provides us several signals/events and notification authorization should be treated the same. We could split the difference and add an `OnNotificationAuthorization` method that can take a callback and keep the `RequestNotificationAuthorization` with a timeout for more sync behavior. Thoughts?

Issues:

1. It is a loss to lose that functionality on the frontend side but that code `await NotificationService.RequestNotificationAuthorization();` will block for up to 3 minutes waiting for the user to respond. Of course there are ways to mitigate that but I could see this scenario playing out where users don't understand why their code hangs for that time. 
2. The platforms are different but for the goal of cross-platform consistency writing the same code is what I want to enable. I can see this being confusing. What are your thoughts?
3. Where did 3 go?!
4. I should fix this
5. I don't see this as a big issue but I can add logs/docs

leaanthony:
I think it's ok to ask permission at the start - many apps do this. If they don't grant it after a certain time then notifications won't work. I think that's fine. Notifications are runtime features so asking permissions just before they happen seems weird. We could have an app-level flag indicating if permissions are granted or not (true on Linux and Windows by default?). `RequestNotificationAuthorization` then becomes async in that it's just a request which will update the app-level flag if granted. Then notifications can just check if it's been authorized before sending. The auth granted callback can simply call a Go method to set the flag. Then it's not a concern of the developer (the user chose not to authorise so....).

popaprozac:
I follow and that seems reasonable enough to me. I'll look at adding the app-level flag.

Because of this conversation I went down a rabbit hole looking at how to support callbacks in method bindings...
Happy to discuss in Discord but I'm sure you've thought about it before hehe

popaprozac:
@leaanthony I think we are good to close this. You can call the sync `CheckNotificationAuthorization` method to check status at any time so I don't think we need to add the app level flag. Thoughts?

PR #4889, opened 2026-01-21
feat(v3): Add WindowMouseEnter/WindowMouseLeave events for tray popup windows

### LINKED ISSUE #4887: How to automatically focus on the current application when the mouse hovers over it
I am now encountering an issue: after performing actions in other applications with the mouse, such as opening a browser and customizing something in it (while the current Wails application is already set to be on top), when the mouse returns to the Wails application and I try to click on a function within the app, I need to click twice. The first click seems to just focus on the current app, and the second click actually triggers the specific function. How can I make it so that moving the mouse to a designated position allows the click action to work immediately?

leaanthony:
## Summary

Implements mouse enter/leave events for windows on all three platforms (Windows, macOS, Linux), enabling tray popup windows to auto-focus when the mouse enters them.

- Added `WindowMouseEnter` and `WindowMouseLeave` common events
- Implemented `FocusOnMouseEnter` window option for automatic focus on mouse hover
- Added cross-platform mouse tracking that works even when windows are not focused
- Created `mouse-hover` example demonstrating the feature
- Updated event generation system to include standard Wails header comments
- Documented event generation process in AGENTS.md

### Platform Implementation Details

- **Windows**: Uses `WM_MOUSEMOVE` with `TrackMouseEvent` API and `WM_MOUSELEAVE`
- **macOS**: Uses `NSTrackingArea` with `NSTrackingActiveAlways` flag via a `MouseTrackingView` overlay
- **Linux**: Uses GTK `enter-notify-event` and `leave-notify-event` signals

Closes #4887

## Test plan

- [ ] Test `mouse-hover` example on Windows
- [ ] Test `mouse-hover` example on macOS
- [ ] Test `mouse-hover` example on Linux
- [ ] Verify events fire when window is not focused
- [ ] Verify `FocusOnMouseEnter` option works correctly
- [ ] Test with tray popup window use case


Nurysso:
Isn’t this behavior dependent on the WM/compositor? For example, on KWin (KDE), pointer enter/leave events are typically delivered, but focus-on-hover may be restricted, whereas compositors like Hyprland are more permissive.
How is this handled?

PR #4924, opened 2026-01-29
fix(v3/bindings): map time.Time to JavaScript Date type

### LINKED ISSUE #4753: [v3] Method binding is converted from the time.Time type to its own Time type
### Description

Hi. I am using wails v3-alpha41.
When we created a Go language service with a method that returns the `time.Time` type, the TypeScript code generated by Wails' Method binding generated its own `Time` type instead of the Date type.

My understanding is that the `time.Time` type of Go is converted to the `Date` type in TypeScript.
https://v3alpha.wails.io/features/bindings/methods/#complex-types

Here are generated codes

#### greetservice.go

```go
package main

import "time"

type GreetService struct{}

func (g *GreetService) GetTime() time.Time {
	return time.Now()
}
```

#### `frontend/bindings/changeme/greetservice.ts` (Generated)

```ts
// Cynhyrchwyd y ffeil hon yn awtomatig. PEIDIWCH Â MODIWL
// This file is automatically generated. DO NOT EDIT

// eslint-disable-next-line @typescript-eslint/ban-ts-comment
// @ts-ignore: Unused imports
import { Call as $Call, CancellablePromise as $CancellablePromise, Create as $Create } from "@wailsio/runtime";

// eslint-disable-next-line @typescript-eslint/ban-ts-comment
// @ts-ignore: Unused imports
import * as time$0 from "../time/models.js";

export function GetTime(): $CancellablePromise<time$0.Time> {
    return $Call.ByID(570467169);
}
```

#### `frontend/bindings/time/models.ts` (Generated)

```ts
export type Time = any;
```

### To Reproduce

Upload example project to https://github.com/mtgto/example-wails3-vanilla-ts

### Expected behaviour

The return value of the generated method is the `Date` type

### Screenshots

_No response_

### Attempted Fixes

_No response_

### System Details

```shell
❯ wails3 doctor
 Wails (v3.0.0-alpha.41)  Wails Doctor

# System

| Version       | 26.1                             |
| ID            | 25B78                            |
| Branding      | MacOS 26.1                       |
| CPU           | Apple M2 Pro                     |
| CPU 1         | Apple M2 Pro                     |
| CPU 2         | Apple M2 Pro                     |
| GPU           | 16 cores, Metal Support: Metal 4 |

# Build Environment

| Wails CLI      | v3.0.0-alpha.41                                                                    |
| Go Version     | go1.25.4                                                                           |

# Dependencies

| Xcode cli tools | 2416                                                 |
|                                                                        |

# Checking for issues

 SUCCESS  No issues found

# Diagnosis

```

### Additional context

_No response_

ddmoney420:
## Summary

- Maps Go's `time.Time` type to JavaScript's native `Date` type in TypeScript bindings
- Adds `$Create.Date` runtime function to convert RFC3339 strings to Date objects
- Removes generation of unnecessary `time/models.ts` file with `export type Time = any;`

## Problem

When a Go service method returns `time.Time`, the TypeScript binding generator was creating:
```typescript
// time/models.ts
export type Time = any;

// greetservice.ts
import * as time$0 from "../time/models.js";
export function GetTime(): $CancellablePromise<time$0.Time> { ... }
```

This happened because `time.Time` implements `json.Marshaler`, and the generator treats all JSON marshalers as `any` since their output type is unknown.

## Solution

Added special-case handling for `time.Time` since it marshals to RFC3339 format which JavaScript's `Date` constructor can parse:

1. **type.go**: Return `"Date"` directly for `time.Time` types
2. **imports.go**: Don't import `time` package or create models for `time.Time`
3. **predicates.go**: `IsAny()` returns `false` for `time.Time`
4. **default.go**: Default value for `time.Time` is `null`
5. **create.go**: Use `$Create.Date` to convert RFC3339 strings to Date objects
6. **create.ts**: Added `Date()` creation function to runtime

## Test plan

- [x] Generator tests pass
- [ ] Manually test with a service that returns `time.Time`
- [ ] Verify generated TypeScript uses `Date` type
- [ ] Verify dates are properly parsed from RFC3339 strings

Fixes #4753

## Release Notes

* **New Features**
  * Improved date/time handling in code generation: Go's time.Time type now maps directly to JavaScript Date objects in generated TypeScript. Added RFC3339 string-to-Date conversion support and proper null value handling for zero timestamps, providing better type safety and developer experience.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/v3/internal/generator/collect/imports.go
+++ b/v3/internal/generator/collect/imports.go
@@ -152,6 +152,11 @@ func (imports *ImportMap) addTypeImpl(typ types.Type, visited *typeutil.Map) {
 				return
 			}
 
+			// Special case: time.Time will render as JS Date hence no dependencies and no model
+			if obj.Pkg().Path() == "time" && obj.Name() == "Time" {
+				return
+			}
+
 			if obj.Pkg().Path() == imports.Self {
 				imports.ImportModels = true
 			}
--- a/v3/internal/generator/collect/predicates.go
+++ b/v3/internal/generator/collect/predicates.go
@@ -426,6 +426,13 @@ func IsAny(typ types.Type) bool {
 	// Follow alias chain.
 	typ = types.Unalias(typ)
 
+	// Special case: time.Time renders as JS Date, not any
+	if named, ok := typ.(*types.Named); ok {
+		if obj := named.Obj(); obj.Pkg() != nil && obj.Pkg().Path() == "time" && obj.Name() == "Time" {
+			return false
+		}
+	}
+
 	if MaybeJSONMarshaler(typ) != NonMarshaler {
 		// If typ is either a named type, an interface, a pointer or a struct,
 		// it will be rendered as (possibly an alias for) the TS any type.
--- a/v3/internal/generator/render/default.go
+++ b/v3/internal/generator/render/default.go
@@ -102,6 +102,12 @@ func (m *module) renderNamedDefault(typ aliasOrNamed, quoted bool) (result strin
 		return m.JSDefault(typ.Underlying(), quoted), true
 	}
 
+	// Special case: time.Time renders as JS Date, default to null
+	// (Go's zero time doesn't map cleanly to JS Date)
+	if typ.Obj().Pkg().Path() == "time" && typ.Obj().Name() == "Time" {
+		return "null", true
+	}
+
 	if quoted {
 		// WARN: Do not test with IsAny/IsStringAlias here!! We only want to catch marshalers.
 		if collect.MaybeJSONMarshaler(typ) == collect.NonMarshaler && collect.MaybeTextMarshaler(typ) == collect.NonMarshaler {
--- a/v3/internal/generator/render/type.go
+++ b/v3/internal/generator/render/type.go
@@ -181,6 +181,13 @@ func (m *module) renderNamedType(typ aliasOrNamed, quoted bool) (result string, 
 		return "void", false
 	}
 
+	// Special case: time.Time renders as JS Date
+	// time.Time implements json.Marshaler and normally renders as 'any',
+	// but it actually marshals to an RFC3339 string that JS Date can parse.
+	if typ.Obj().Pkg().Path() == "time" && typ.Obj().Name() == "Time" {
+		return "Date", false
+	}
+
 	if quoted {
 		switch a := types.Unalias(typ).(type) {
 		case *types.Basic:
--- a/v3/internal/runtime/desktop/@wailsio/runtime/src/create.ts
+++ b/v3/internal/runtime/desktop/@wailsio/runtime/src/create.ts
@@ -24,6 +24,14 @@ export function ByteSlice(source: any): string {
 }
 
 /**
+ * Date is a creation function that converts RFC3339 strings
+ * (from Go's time.Time JSON marshaling) to JavaScript Date objects.
+ */
+export function Date(source: any): globalThis.Date | null {
+    return ((source == null) ? null : new globalThis.Date(source));
+}
+
+/**
  * Array takes a creation function for an arbitrary type
  * and returns an in-place creation function for an array
  * whose elements are of that type.
--- a/v3/internal/generator/render/create.go
+++ b/v3/internal/generator/render/create.go
@@ -54,6 +54,11 @@ func (m *module) needsCreateImpl(typ types.Type, visited *typeutil.Map) bool {
 			return false
 		}
 
+		// Special case: time.Time renders as JS Date, needs creation to parse string to Date
+		if t.Obj().Pkg().Path() == "time" && t.Obj().Name() == "Time" {
+			return true
+		}
+
 		if collect.IsAny(typ) || collect.IsStringAlias(typ) {
 			break
 		} else if collect.IsClass(typ) {
@@ -149,6 +154,12 @@ func (m *module) JSCreateWithParams(typ types.Type, params string) string {
 
 		if m.collector.IsVoidAlias(t.Obj()) {
 			return "$Create.Any"
+		}
+
+		// Special case: time.Time renders as JS Date
+		// JSON marshals time.Time as RFC3339 string, which Date constructor can parse
+		if t.Obj().Pkg().Path() == "time" && t.Obj().Name() == "Time" {
+			return "$Create.Date"
 		}
 
 		if !m.NeedsCreate(typ) {


leaanthony:
Thanks for opening this! I know a few have wanted better date support. Could you please add/update the tests to ensure it's working (and continues to work) and could you add an entry to the changelog located at `v3/UNRELEASED_CHANGELOG.md`? Thanks!

v3/internal/runtime/desktop/@wailsio/runtime/src/create.ts
```diff
@@ -23,6 +23,14 @@ export function ByteSlice(source: any): string {
     return ((source == null) ? "" : source);
 }
 
+/**
+ * Date is a creation function that converts RFC3339 strings
+ * (from Go's time.Time JSON marshaling) to JavaScript Date objects.
+ */
+export function Date(source: any): globalThis.Date | null {
+    return ((source == null) ? null : new globalThis.Date(source));
+}
```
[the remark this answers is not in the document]
fbbdev: I share the rabbit's concern and recommend naming the function `Time` instead of `Date`

On the other hand, I don't think any kind of validation is needed:
- the null check is unnecessary because `time.Time` never marshals to `null`
- the validity check is unnecessary because people can easily do it themselves: `isNaN(dateObj)`

Hence this can simply be `return new Date(source)`

v3/internal/runtime/desktop/@wailsio/runtime/src/create.ts
```diff
@@ -23,6 +23,14 @@ export function ByteSlice(source: any): string {
     return ((source == null) ? "" : source);
 }
 
+/**
+ * Date is a creation function that converts RFC3339 strings
+ * (from Go's time.Time JSON marshaling) to JavaScript Date objects.
+ */
+export function Date(source: any): globalThis.Date | null {
```
fbbdev, line 30: I suggest moving the function down, after all standard go types.

v3/internal/generator/render/default.go
```diff
@@ -102,6 +102,12 @@ func (m *module) renderNamedDefault(typ aliasOrNamed, quoted bool) (result strin
 		return m.JSDefault(typ.Underlying(), quoted), true
 	}
 
+	// Special case: time.Time renders as JS Date, default to null
+	// (Go's zero time doesn't map cleanly to JS Date)
+	if typ.Obj().Pkg().Path() == "time" && typ.Obj().Name() == "Time" {
+		return "null", true
```
fbbdev, line 108: For max compatibility between frontend and backend, the default value should match as much as possible the semantics of Go's zero value as well as TypeScript's constraints. I usually abide by the following 3 criteria:

1. Go's zero value should translate to the default: `time.Time{}` -> Marshal -> default
2. A roundtrip through the backend should preserve the value: default -> Unmarshal -> Marshal -> default
3. The default value should satisfy the corresponding TS type in strict mode

I did some tests with V8 and webkit and it looks like Go's zero time maps cleanly to `new Date("0001-01-01T00:00:00.000Z")`, hence I recommend using that as default instead of `null`.

On the other hand, `null` violates all 3 criteria:

1. `time.Time{}` -> Marshal -> `new Date("0001-01-01T00:00:00.000Z") !== null`
2. `null` -> Unmarshal -> Marshal -> `new Date("0001-01-01T00:00:00.000Z") !== null`
3. `null` is incompatible with type `Date` in TS strict mode

v3/internal/generator/collect/predicates.go
```diff
@@ -426,6 +426,13 @@ func IsAny(typ types.Type) bool {
 	// Follow alias chain.
 	typ = types.Unalias(typ)
 
+	// Special case: time.Time renders as JS Date, not any
+	if named, ok := typ.(*types.Named); ok {
```
fbbdev, line 430: I think it's safer to omit this change and leave the type predicates unmodified.

The special casing in `ImportMap.addTypeImpl` and in the rendering functions should be enough to stop any other logic from kicking in.

v3/internal/generator/collect/imports.go
```diff
@@ -152,6 +152,11 @@ func (imports *ImportMap) addTypeImpl(typ types.Type, visited *typeutil.Map) {
 				return
 			}
 
+			// Special case: time.Time will render as JS Date hence no dependencies and no model
+			if obj.Pkg().Path() == "time" && obj.Name() == "Time" {
```
fbbdev, line 156: Since we repeat this check a lot of times, I recommend wrapping it in a dedicated method just like `IsVoidAlias`, e.g. IsTime? IsStandardTime?.. (not sure about the naming here)

(before doing so, I suggest you have a look at #4941 or wait for it to be merged: the current version of `IsVoidAlias` has a flaw that is fixed there and we'd better not replicate).

fbbdev, line 156: Note that you can't compare the package path against the string `"time"` directly, because there are advanced compiler features that enable people to remap package paths to whatever they like.

Instead, the [`SystemPaths`](https://github.com/wailsapp/wails/blob/v3-alpha/v3/internal/generator/config/paths.go#L10) struct should be extended with a new field for the `"time"` package, which should then be resolved by the [`ResolveSystemPaths`](https://github.com/wailsapp/wails/blob/v3-alpha/v3/internal/generator/load.go#L13) function (just like `"context"`).

The resolved path will then be available for comparison as e.g. `collector.systemPaths.TimePackage`.

v3/internal/generator/render/type.go
```diff
@@ -181,6 +181,13 @@ func (m *module) renderNamedType(typ aliasOrNamed, quoted bool) (result string,
 		return "void", false
 	}
 
+	// Special case: time.Time renders as JS Date
+	// time.Time implements json.Marshaler and normally renders as 'any',
+	// but it actually marshals to an RFC3339 string that JS Date can parse.
+	if typ.Obj().Pkg().Path() == "time" && typ.Obj().Name() == "Time" {
+		return "Date", false
```
fbbdev, line 188: When _interface mode_ is enabled, we don't emit type creation code, hence we can't translate `time.Time` to `Date`. We need an additional check here to return `"string", false` when `m.UseInterfaces` is true.

No change is needed in the other helpers (create/default) since they are never invoked in interface mode.

fbbdev (changes requested):
Hi there! Thank you for submitting this. I took some time to go over your code.

I see that it matches closely the pre-existing strategy to replace the `Void` stub type. The implementation is clean and almost there, but some fixes are still needed before it's ready. You'll find detailed comments in the diff.

A couple of more general points (on which perhaps @leaanthony might want to give some feedback):

1. As implemented right now, this is a breaking change: we can reasonably expect that quite some people will have built their apps around `time.Time` translating to a `string`, and changing it to a `Date` _will_ break their code.
2. Javascript's `Date` object is notoriously quirky (it mixes UTC and local time in a messy way) and is now [considered legacy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date). Many developers, myself included, try to avoid it like the plague.
3. Translating to `Date` by default will make it harder for us to switch to the new [`Temporal` API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal) once it's finalized and widely supported.

As a quick and robust solution to all 3 issues, I propose we make translation to `Date` optional behind a command-line flag, and default to `string` instead. This is future-proof but at the same time provides options for people who want `Date` now.

Last but not least:

4. As remarked already by @leaanthony, we are going to need a new test case package (to be added under the `testcases` dir). It should cover every possible usage scenario of `time.Time`: as a method argument, as a return type, as a struct field (named and anonymous), as an alias target, as a type parameter, as a pointer, as a slice or array element, as a map key and value etc.

You'll want to run the task `generator:test:regenerate` and check the output until it makes sense and all tests pass, then commit the newly generated output so that we can review and typecheck it as well.

I recommend committing test output in one big commit at the end so that we can review the other changes in isolation. My expectation is that no existing output file should change; it should only generate new files. But I am not entirely sure about that.

ddmoney420:
Closing this PR — I spread too thin across projects. Thank you @leaanthony for the constructive feedback. I'll revisit when I can give it proper attention with tests and changelog.

PR #4931, opened 2026-01-29
fix(generator): add defensive checks for cross-package type alias namespace resolution

### LINKED ISSUE #4578: [v3] bindings generator Incorrect namespace reference when generating TS code from Go type alias across packages
### Description

When a Go file uses a type alias from another package, the bindings generator does not correctly generate the namespace reference in the output TypeScript code, leading to an incorrect import path.

### Example Source Code
```
package view_models

import "data_models"

type Model = data_models.Model
```

### Actual Output
```
const $$createType6 = $0.Model.createFrom;
```

### Expected Output
```
const $$createType6 = view_models$0.Model.createFrom;
```

### Screenshots

_No response_

### Attempted Fixes

_No response_

### System Details

```shell
Wails (v3.0.0-alpha.27
)  Wails Doctor 
                                                                                                              
# System 

| Version       | 15.0                             |
| ID            | 24A335                           |
| Branding      | Sequoia                          |
| CPU           | Apple M1 Pro                     |
| CPU 1         | Apple M1 Pro                     |
| CPU 2         | Apple M1 Pro                     |
| GPU           | 14 cores, Metal Support: Metal 3 |

# Build Environment 

| Wails CLI      | v3.0.0-alpha.27                                                                    |
| Go Version     | go1.25.0                                                                           |

# Dependencies 

┌───────────────────────────┐
| Xcode cli tools | 2409    |
| npm             | 10.9.2  |
| *NSIS           | v3.11   |
|                           |
└─ * - Optional Dependency ─┘

# Checking for issues 

 SUCCESS  No issues found

# Diagnosis 

```

### Additional context

_No response_

ddmoney420:
## Summary
Adds defensive checks in the TypeScript bindings generator to ensure external package imports are properly resolved when handling type aliases across packages.

Fixes #4578

## Problem
When a Go type alias from package A points to a type in package B, the TypeScript bindings generator could produce incorrect namespace references like `$0.Model.createFrom` instead of `data_models$0.Model.createFrom`.

This occurs when `m.Imports.External[pkgPath]` returns a zero-value `ImportInfo{Name: "", Index: 0}` because the package path was not properly registered in the imports map.

## Solution
Added defensive checks in three files to ensure external package imports exist before using them:

- `v3/internal/generator/render/create.go` - in `PostponedCreates`
- `v3/internal/generator/render/type.go` - in `renderNamedType`  
- `v3/internal/generator/render/default.go` - in `renderNamedDefault` (two places)

If an import is missing, it's now dynamically added to prevent the empty namespace prefix.

## Test plan
- [x] All existing generator tests pass (8 configurations)
- [ ] Test with cross-package type aliases in a real project

## Release Notes

* **Bug Fixes**
  * Improved code generation handling for external packages to prevent invalid output
  * Enhanced runtime stability in server-side rendering environments with proper environment detection
  * Fixed file dialog result collection on Linux

* **Chores**
  * Internal improvements to import resolution logic
  * Added browser environment guards across the runtime layer for better compatibility


fbbdev:
Hi @ddmoney420 , I am the original author of the v3 binding generator.
I've not had enough bandwidth to contribute to the project lately, so thank you for taking this on.

Your PR was mistakenly opened against master (v2) so in order to review it I took the liberty to retarget it towards the v3-alpha branch.

I noticed that you pushed onto the same branch two commits that look unrelated (although valuable!). Not sure if this was intentional. Before we move forward with this PR, could you please split the additional commits into their own PRs? (I recommend one PR per commit since they solve two different issues).

fbbdev:
Now getting to the actual fix.. I see that you tracked down the issue to a missing entry in the import map at render time.

Unfortunately your approach of filling in missing entries on the fly is unsound: the rendering phase is parallelized and relies critically on the assumption that input data is mostly immutable. Breaking that assumption will cause data races and potentially nasty bugs.

Filling import maps accurately is the collector's task, specifically the `ImportMap.addTypeImpl` method ([`v3/internal/generator/collect/imports.go:130`](https://github.com/wailsapp/wails/blob/9d6c03ca57cc734703977d113ce6897daf8c9b0e/v3/internal/generator/collect/imports.go#L130)). If import maps are incomplete, then we must have a bug there.

This sent me down an unexpected rabbit hole, because nothing in that method looks wrong. Actually, aliases share most of the implementation with named types, which work perfectly. As I started to despair, the light came to me: the culprit is the `typeutil.Map` we use to track visited types and break cycles (see [`v3/internal/generator/collect/imports.go:139`](https://github.com/wailsapp/wails/blob/9d6c03ca57cc734703977d113ce6897daf8c9b0e/v3/internal/generator/collect/imports.go#L139)).

```go
case *types.Alias, *types.Named:
	if visited.Set(typ, true) != nil {
		// Break type cycles.
		return
	}
```

The issue is that `typeutil.Map` resolves aliases internally, hence as soon as the alias is visited, the aliased named type is also marked as visited and its package is never added to the import map.

Fortunately, we only use this map for aliases and named types, so we can replace it with a regular map indexed by objects: `visited map[types.Object]bool` and then `if visited[obj] { return }; visited[obj] = true`. That should fix the issue for good.

Before merging it will also be essential to add a test case, check that it reproduces the issue and check that it's fixed afterwards. It should be as simple as adding a new sub-package inside [`testcases/aliases`](https://github.com/wailsapp/wails/tree/v3-alpha/v3/internal/generator/testcases/aliases) and referencing it through an alias.

My recommendation is to revert revert to upstream first, add the test case, regenerate test results (see task `generator:test:regenerate`), check the result, commit the fix, regenerate and check again.

ddmoney420:
Thank you both for the thorough feedback and guidance. I really appreciate you taking the time to explain the collector vs render phase distinction.

I realize I bit off more than I can chew here - this fix requires deeper knowledge of the Wails build pipeline than I currently have. Rather than risk introducing new bugs, I'm going to close this PR.

If someone with more Wails experience wants to pick this up, @fbbdev's explanation is an excellent roadmap.

fbbdev:
@ddmoney420 thanks a lot anyways for debugging the issue, it pointed me in the right direction.

I opened a PR with the proposed fix at #4948, it works for the test case I created but it would be nice if you could validate it as well on some independent input.

Indeed this kind of fix requires in-depth knowledge of the generator pipeline, if anything to avoid concurrency related traps. But it's good for the project if people other than me become familiar with it, and I'm always happy to offer guidance.

PR #4975, opened 2026-02-08
docs: audit and fix hallucinated v3 documentation

leaanthony:
## Summary

- Comprehensive 3-pass audit of all v3 documentation against source code on `origin/v3-alpha`
- Fixed hundreds of hallucinated API patterns, fabricated CLI flags, wrong method signatures, and incorrect struct fields across 59 files (+4439/-4008 lines)
- Each pass used increasingly deep verification: pattern matching, source-code-first specialists (6 agents), field-by-field deep verification (7 agents)

## Key fixes

**Fabricated APIs removed:**
- `OnStartup`/`OnBeforeClose`/`OnShutdown` app hooks → `ServiceStartup` interface, `ShouldQuit`, `RegisterHook`
- Window convenience methods (`OnClose`, `OnFocus`, `OnBlur`) → `OnWindowEvent(events.Common.*, ...)`
- Package-level dialog functions (`application.SaveFileDialog()`) → Manager API (`app.Dialog.SaveFile()`)
- Non-existent JS functions (`OnEvent`, `Emit`) → `Events.On`, `Events.Emit` from `@wailsio/runtime`

**Wrong CLI flags fixed:**
- Removed fabricated `-platform`, `-o`, `-ldflags`, `-icon`, `-windowsconsole`, `-package`, `-skip-package` flags
- Documented actual Taskfile-based build system with `GOOS`/`GOARCH` for cross-compilation
- Added missing commands (`tool lipo`, `tool sign`, `setup signing`, `doctor-ng`, etc.)

**Wrong types/fields corrected:**
- `MacOptions` → `MacWindow`, `WindowsOptions` → `WindowsWindow`, `LinuxOptions` → `LinuxWindow`
- `Resizable` → `DisableResize`, `WindowState` → `StartState`
- `screen.Width` → `screen.Size.Width` (Screen struct has `Size`, `Bounds`, `PhysicalBounds`, `WorkArea` fields)
- `application.NewDefaultLogger()` → `application.DefaultLogger(slog.LevelDebug)`
- `wails.ButtonHidden` → `application.ButtonHidden`
- `ServiceShutdown(ctx context.Context)` → `ServiceShutdown()` (no args)
- JS event callbacks receive `event` object, data at `event.data` not direct args

**Missing documentation added:**
- 20+ window methods (ToggleFullscreen, Bounds, SnapAssist, ToggleFrameless, etc.)
- 30+ platform option fields for MacWindow/WindowsWindow/LinuxWindow
- Complete Application Options table with all fields
- Missing CLI commands and flags
- `ServiceName` interface, `ServiceOptions` fields
- Missing dialog methods (ResolvesAliases, SetMessage, SetButtonText, etc.)

## Test plan

- [ ] Verify docs build successfully with `npm run build` in docs/
- [ ] Spot-check key reference pages against v3 source code
- [ ] Verify no broken internal links

### DIFF
--- a/docs/src/content/docs/concepts/bridge.mdx
+++ b/docs/src/content/docs/concepts/bridge.mdx
@@ -538,10 +538,7 @@ func ProcessLargeFile(path string) error {
     for scanner.Scan() {
         lineNum++
         // Emit progress events
-        app.Event.Emit("file-progress", map[string]interface{}{
-            "line": lineNum,
-            "text": scanner.Text(),
-        })
+        app.Event.Emit("file-progress", lineNum, scanner.Text())
     }
     
     return scanner.Err()
@@ -553,8 +550,10 @@ import { Events } from '@wailsio/runtime'
 import { ProcessLargeFile } from './bindings/FileService'
 
 // Listen for progress
+// event.data is an array of the variadic args: [lineNum, text]
 Events.On('file-progress', (event) => {
-    console.log(`Line ${event.data.line}: ${event.data.text}`)
+    const [line, text] = event.data
+    console.log(`Line ${line}: ${text}`)
 })
 
 // Start processing
--- a/docs/src/content/docs/guides/events-reference.mdx
+++ b/docs/src/content/docs/guides/events-reference.mdx
@@ -214,7 +214,10 @@ Then listen in JavaScript:
 import { Events } from '@wailsio/runtime';
 
 Events.On('files-dropped', (event) => {
-    event.data.forEach(file => {
+    // EmitEvent is variadic, so event.data is a []any wrapping the args.
+    // The files array is the first element.
+    const files = event.data[0];
+    files.forEach(file => {
         console.log('File dropped:', file);
         handleFileUpload(file);
     });
--- a/docs/src/content/docs/guides/menus.mdx
+++ b/docs/src/content/docs/guides/menus.mdx
@@ -436,6 +436,12 @@ When creating a custom context menu, you provide a unique identifier (name) that
 ```go
 // Create a context menu with identifier "imageMenu"
 contextMenu := application.NewContextMenu("imageMenu")
+
+// Add items and call Update() to apply changes
+contextMenu.Add("Cut")
+contextMenu.Add("Copy")
+contextMenu.Add("Paste")
+contextMenu.Update()
 ```
 
 The name parameter ("imageMenu" in this example) serves as a unique identifier that will be used to:
--- a/docs/src/content/docs/reference/application.mdx
+++ b/docs/src/content/docs/reference/application.mdx
@@ -347,10 +347,8 @@ app.RegisterService(application.NewService(&MyService{}))
 Emits a custom event.
 
 ```go
-func (em *EventManager) Emit(name string, data ...any) bool
+func (em *EventManager) Emit(name string, data ...any)
 ```
-
-**Returns:** `true` if the event was cancelled by a hook.
 
 **Example:**
 


docs/src/content/docs/concepts/architecture.mdx
```diff
@@ -348,17 +348,27 @@ Shutdown.Save -> End
 **Lifecycle hooks:**
 
 ```go
+// Services provide startup/shutdown hooks
+type AppService struct{}
+
+func (s *AppService) ServiceStartup(ctx context.Context, options application.ServiceOptions) error {
```
fbbdev: Do we actually recommend using services to hook into application lifecycle events? 🤔

Shouldn't we use the ApplicationStarted event? (not sure what's the equivalent for shutdown)

docs/src/content/docs/features/bindings/advanced.mdx
```diff
@@ -286,9 +285,9 @@ func (s *AuthService) GetCurrentUser() User {
     // Implementation...
 }
 
-// Internal helper method
+// Internal helper method - not exported to frontend
 //wails:internal
```
fbbdev, line 289: This is a very common misunderstanding in generated docs

`//wails:ignore` prevents exporting to frontend (i.e. only available to go)

`//wails:internal` DOES export to frontend, but keeps the function local to the service JS module (i.e. unexported FROM JS)

It's useful to make the function available to custom wrappers injected with `//wails:inject`, but not to external consumers of the module

An example of that pattern should be available in the built-in db service if I don't misremember.

It's a powerful pattern worth mentioning.

docs/src/content/docs/concepts/bridge.mdx
```diff
@@ -549,11 +549,11 @@ func ProcessLargeFile(path string) error {
 ```
 
 ```javascript
-import { On } from '@wailsio/runtime'
+import { Events } from '@wailsio/runtime'
 import { ProcessLargeFile } from './bindings/FileService'
 
 // Listen for progress
-On('file-progress', (data) => {
+Events.On('file-progress', (data) => {
```
fbbdev: Here we have a widespread misunderstanding of the JS API. The argument to the callback is NOT event data. It's an event object with fields name, data and sender.

Thus either `({ data }) => { ... }` or `(ev) => { ev.data ... }`

docs/src/content/docs/reference/events.mdx
```diff
@@ -78,225 +80,225 @@ cleanup := app.OnEvent("user-action", func(e *application.CustomEvent) {
...
+    dialog.Show()
 })
-
-// Call service method
-const items = await FetchData("search term")
-displayItems(items)
 ```
 
-### Progress Updates
+## Common Patterns
```
fbbdev, line 299: Wondering whether the common patterns section should be merged into features/events/system.mdx instead of partially duplicating it

This is a reference doc so not sure it belongs here

docs/src/content/docs/guides/events-reference.mdx
```diff
@@ -148,15 +148,24 @@ Events.On('common:WindowFocus', () => {
 
 ### 2. Responding to Theme Changes
```
fbbdev, line 149: Do we still need this file given the extensive documentation in features and reference sections?

docs/src/content/docs/concepts/architecture.mdx
leaanthony: The comment means that Services can hook into those lifecycle events

docs/src/content/docs/guides/menus.mdx
```diff
@@ -415,8 +435,7 @@ When creating a custom context menu, you provide a unique identifier (name) that
 
 ```go
 // Create a context menu with identifier "imageMenu"
-contextMenu := application.NewContextMenu()
-app.ContextMenus.Add("imageMenu", contextMenu)
+contextMenu := application.NewContextMenu("imageMenu")
 ```
 
 The name parameter ("imageMenu" in this example) serves as a unique identifier that will be used to:
```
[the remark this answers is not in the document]
leaanthony: Valid catch. Looking at the source: `NewContextMenu()` already calls `Update()` internally (see `v3/pkg/application/menu.go:17`), so the constructor handles it. But this snippet creates the menu first, then adds items *after* — so yes, an `Update()` call is needed after adding items. Will fix.

docs/src/content/docs/guides/events-reference.mdx
```diff
@@ -166,24 +175,56 @@ Events.On('common:ThemeChanged', (event) => {
...
+
+    // Optionally get drop target details
+    details := e.Context().DropTargetDetails()
+    if details != nil {
+        fmt.Println("Dropped on element:", details.ID)
+    }
+
+    // Notify the frontend
+    window.EmitEvent("files-dropped", files)
+})
```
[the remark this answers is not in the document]
leaanthony: Correct. `window.EmitEvent(name, data ...any)` wraps args into `[]any`, so `EmitEvent("files-dropped", files)` produces `Data: []any{files}` — a nested array in JS. The JS `event.data.forEach(file => ...)` would iterate the outer array (one element: the inner slice), not the individual files. Will fix.

docs/src/content/docs/reference/application.mdx
```diff
@@ -180,177 +242,284 @@ app.Clipboard.SetText("Copied text")
...
 
 Emits a custom event.
 
 ```go
-func (a *App) EmitEvent(name string, data ...interface{})
+func (em *EventManager) Emit(name string, data ...any) bool
 ```
 
+**Returns:** `true` if the event was cancelled by a hook.
+
```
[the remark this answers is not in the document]
leaanthony: Correct — the actual signature is `func (em *EventManager) Emit(name string, data ...any)` with no return value. The `bool` return and cancellation description are fabricated. Will fix to match the source.

docs/src/content/docs/guides/menus.mdx
```diff
@@ -474,9 +493,8 @@ After calling `Destroy()`, using the context menu reference again will result in
...
 
 ```go
-// Backend: Create the context menu
-imageMenu := application.NewContextMenu()
-app.ContextMenus.Add("imageMenu", imageMenu)
+// Backend: Create the context menu (auto-registered by name)
+imageMenu := application.NewContextMenu("imageMenu")
 
 // Add relevant operations
 imageMenu.Add("View Full Size").OnClick(func(ctx *application.Context) {
```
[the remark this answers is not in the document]
leaanthony: This example actually does call `Update()` at line 519 after all items are added — scroll down a bit. So this one is a false positive from Copilot. The *earlier* snippet (line 438, creating the context menu) is the one that needs `Update()` added, which is covered by the other comment.

docs/src/content/docs/features/bindings/advanced.mdx
leaanthony: You're right to flag this. Looking at it again, the docs at lines 47-72 *do* correctly describe the distinction:

- `//wails:ignore` = completely excluded from JS generation (Go-only)
- `//wails:internal` = generated in JS but not re-exported from the package index (module-private)

And the Internal + Inject pattern section (line 344+) documents the SQLite service pattern correctly. But the example comment at line 288 says "Module-private method" which is correct — though the broader description could be clearer that `//wails:internal` **does** generate JS code.

I'll review all instances to make sure the distinction is consistently correct throughout. Thanks for the correction on the nuance.

docs/src/content/docs/concepts/bridge.mdx
leaanthony: The diff shows the fix changed `(data) => { data.line }` to `(event) => { event.data.line }` — which should be correct (event object, accessing `.data` property). But there's a deeper issue here: `app.Event.Emit("file-progress", map[string]interface{}{...})` uses the variadic API, so `Data` is set to `[]any{theMap}`. In JS, `event.data` would be an array with one element, not the map directly. So `event.data.line` wouldn't work — it'd need `event.data[0].line`. Will fix by using `EmitEvent` with a `CustomEvent` directly to avoid the variadic wrapping.

docs/src/content/docs/reference/events.mdx
leaanthony: Fair point. The common patterns section is more of a guide/tutorial than a reference. It could be moved to `features/events/system.mdx` or a dedicated patterns guide to reduce duplication. Will consider restructuring in a follow-up.

docs/src/content/docs/guides/events-reference.mdx
leaanthony: Good question. There is overlap with the features and reference sections. This file could potentially be removed or trimmed to just a quick-reference cheat sheet that links to the detailed docs. Will evaluate in a follow-up pass.

PR #5024, opened 2026-02-27
feat(darwin): Add NSPanel support for Spotlight-like windows

### LINKED ISSUE #5023: [v3] Add panel support for darwin window

I have an application with two windows: a main window, and a spotlight-like auxiliary window triggered from a keybind event. When following the example from [Spotlight Example](https://github.com/wailsapp/wails/tree/v3-alpha/v3/examples/spotlight), it does not overlay applications when in fullscreen unless I set the entire application's activation policy to ActivationPolicyAccessory.

Panel support for darwin windows. Namely, NSWindowStyleMaskNonactivatingPanel would solve my problem.

NSWindowStyleMaskNonactivatingPanel
The window is a panel or a subclass of NSPanel that does not activate the owning app.
https://developer.apple.com/documentation/appkit/nswindow/stylemask-swift.struct?language=objc

Dynamically shifting to ActivationPolicyAccessory when the spotlight window is open. This works, but disables many other features of the application while my spotlight window is open, such as the dock icon for the main window.

### Additional context

I have a working wails source code patch I'd be happy to clean up and submit for review.

Related resources:
https://stackoverflow.com/questions/36205834/allow-an-nswindow-nspanel-to-float-above-full-screen-apps

Grantmartin2002:
This adds support for creating NSPanel-based windows instead of NSWindow, enabling auxiliary windows that can appear over fullscreen apps without activating the application.

## Key changes

- Add `WindowClass MacWindowClass` to `MacWindow` configuration with `MacWindowClassWindow` (default) and `MacWindowClassPanel` options
- Add `MacPanelPreferences` struct with panel-specific options:
  - `FloatingPanel` - float above other windows
  - `BecomesKeyOnlyIfNeeded` - become key only when needed
  - `NonactivatingPanel` - receive input without activating the app
  - `UtilityWindow` - utility window style
- Implement `WebviewPanel` class (NSPanel subclass) with same functionality as `WebviewWindow`
- Add `panelNew()` function to create panels with configurable style masks

The `NSWindowStyleMaskNonactivatingPanel` style allows the panel to receive keyboard input without activating the owning application.

Fixes #5023

## Type of change
  
# How Has This Been Tested?
- [ ] Windows
- [x] macOS
- [ ] Linux
```
| Version       | 15.7.1                           |
| ID            | 24G231                           |
| Branding      | Sequoia                          |
| CPU           | Apple M4 Pro                     |
| CPU 1         | Apple M4 Pro                     |
| CPU 2         | Apple M4 Pro                     |
| GPU           | 20 cores, Metal Support: Metal 3 |
| Memory        | 48 GB                            |

# Build Environment 

┌────────────────────────────────┐
| Wails CLI    | v3.0.0-alpha.71 |
| Go Version   | go1.25.1        |
| -buildmode   | exe             |
| -compiler    | gc              |
| CGO_CFLAGS   |                 |
| CGO_CPPFLAGS |                 |
| CGO_CXXFLAGS |                 |
| CGO_ENABLED  | 1               |
| CGO_LDFLAGS  |                 |
| GOARCH       | arm64           |
| GOARM64      | v8.0            |
| GOOS         | darwin          |
└────────────────────────────────┘

# Dependencies 

┌───────────────────────────────────────────────────────────────────────────────────┐
| Xcode cli tools | 2410                                                            |
| npm             | 11.5.1                                                          |
| *NSIS           | v3.11                                                           |
| docker          | *Docker version 28.2.2, build e6534b4 (cross-compilation ready) |
|                                                                                   |
└───────────────────────────── * - Optional Dependency ─────────────────────────────┘

# Checking for issues 

 SUCCESS  No issues found

# Diagnosis 

```
# Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added macOS Panel (NSPanel) as an alternative window class with per-window selection.
  * Added panel preferences: floating, becomes-key-only-if-needed, non-activating, utility.
  * Unified webview configuration and improved keyboard event handling for windows and panels.

* **Documentation**
  * Changelog and docs updated with Panel options, examples (including a Spotlight-like panel), and usage guidance.


v3/pkg/application/webview_window_darwin.m
```diff
@@ -28,6 +28,17 @@ - (WebviewWindow*) initWithContentRect:(NSRect)contentRect styleMask:(NSUInteger
...
+// This ensures KeyBindings work regardless of first responder state
+- (void)sendEvent:(NSEvent *)event {
+    if (event.type == NSEventTypeKeyDown) {
+        // Process through keybinding system first
+        [self keyDown:event];
+        // Still pass to WKWebView for normal input handling
+        // (keybinding callbacks like Hide will take effect regardless)
+    }
+    [super sendEvent:event];
+}
```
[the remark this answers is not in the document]
Grantmartin2002: This is true, but is intended. if i have an esc keybind that closes my webview, just because i have the webview focused doesn't mean the go keybind should not apply. In fact, we really want the exact opposite.

The concerns with the way i am doing it currently are
- Updates keybind behavior for already existing NSWindow types. Could change app behavior
- If user is typing in a textbox or something, and they have a keybind for the letter "a", for example, when they type "a", their go keybind will also be activated.

Better solution?

leaanthony (commented):
Hey @Grantmartin2002 👋 

Thanks for this! NSPanel support is something people have been asking about and this is a solid start. Really nice job on the docs too — the Spotlight-like panel example and the note about `ActivationPolicyAccessory` not being needed for panels is great 🙏

I've got some feedback before we can merge this:

### 1. `sendEvent:` override on `WebviewWindow` is a breaking change

This is the big one. The `sendEvent:` override is being added to **both** `WebviewPanel` and `WebviewWindow`. On `WebviewWindow` this changes behavior for all existing users — every keystroke will now fire Go keybindings regardless of first responder state, which means typing in a `<input>` field would also trigger keybindings. You called this out yourself in the PR description (good instinct!).

This needs to be **opt-in**, not default. Please remove the `sendEvent:` override from `WebviewWindow` (keep the existing `keyDown:` behavior). For `WebviewPanel` it's fine since it's new, but ideally add an option for it there too so users can control it.

### 2. Double-dispatch of key events

With `sendEvent:` calling `[self keyDown:event]` and then `[super sendEvent:event]`, the normal event dispatch chain will call `keyDown:` again when the window is first responder — so `processWindowKeyDownEvent` fires **twice** per keystroke. 

If you keep `sendEvent:` (for panels), don't call `[super sendEvent:event]` after handling key-down events — that should prevent the double-fire.

### 3. Inconsistent memory management

In `windowNew()` the delegate gets `autorelease`d, but `WebviewPanel` has a custom `setDelegate:` that does `retain`/`release` and a `dealloc` cleanup. These two patterns conflict — the `autorelease` in `panelNew()` combined with the `retain` in `setDelegate:` will over-retain.

I'd prefer we stick with the `autorelease` pattern consistently unless you have a specific reason for the explicit retain/release approach? Also worth noting that `WebviewWindow` is missing the `userContentController` cleanup that you added to `WebviewPanel.dealloc` — that's likely a pre-existing leak, so nice catch if intentional!

### 4. Rename `NSWindow`/`NSPanel` constants

Using `NSWindow` and `NSPanel` as Go constant names masks the actual Cocoa class names, which gets confusing. Please rename these — something like `MacWindowClassWindow` and `MacWindowClassPanel` would work, or whatever feels right as long as they don't shadow the Cocoa types.

### 5. Code duplication in `.m` — consider a protocol or macro

The `WebviewPanel` implementation is nearly identical to `WebviewWindow` (init, responder methods, etc). The refactoring of `keyStringFromEvent` → `keyStringFromKeyEvent` and `dispatchKeyDownEvent` as static helpers was a good move. Could you look into doing something similar for the remaining duplicated bits? Even just a comment acknowledging the duplication with a TODO would be fine if it's tricky to extract cleanly.

### 6. `panelNew` has a lot of bool parameters

```c
void* panelNew(unsigned int id, int width, int height, bool fraudulentWebsiteWarningEnabled, 
               bool frameless, bool enableDragAndDrop, struct WebviewPreferences preferences,
               bool floatingPanel, bool becomesKeyOnlyIfNeeded, bool nonactivatingPanel, bool utilityWindow)
```

4 trailing bools is easy to get wrong at call sites. Could you wrap the panel-specific ones in a `struct PanelPreferences` (similar to the existing `WebviewPreferences`)? That way the call site is self-documenting.

### 7. Docs

Great job here overall! Will just need a minor update once the constant names change (point 4).

---

Happy to answer any questions on these. The core feature is really useful — just want to make sure we get the details right before it lands. Thanks again for the contribution! 🙌

Grantmartin2002:
@leaanthony Thank you for the thorough review! Few things:
### 1. sendEvent: override on WebviewWindow is a breaking change
Agreed. Did it originally to keep behavior consistent and considered an opt in option, but honestly not necessary. will remove
### 2. Double-dispatch of key events
- This send event override is only necessary when NonactivatingPanel is set to true. Something about the event propagation changes with this setting, causing my keybinds to no longer work without the override
- I do not observe your alleged double dispatch for panels. dispatchKeyDownEvent is called once. Tested with nonactivatingpanel attribute both on and off. 
- One (mild/moderate) problem this double event does cause though is, if called within text input, the text input and the WKWebView will both handle the event. ie, if i have a keybind on a, and type a, both the keybind logic and textinput typing will occur. This differs from window behavior (without sendevent override), which does NOT type the letter and only activates the keybind logic. I've spent a few hours exploring alternatives to replicate the window logic exactly, but haven't found a clean, consistent way that accounts for all edge cases.
- If we do not call [super sendEvent:event], then keys no longer work in text input fields.
### 3. Inconsistent memory management
can you point me to the exact lines you are referring to? Memory management looks identical to me
### 4.  Rename NSWindow/NSPanel constants
Done
### 5. Code duplication in .m — consider a protocol or macro
I should be able to do something for consolodiating panelNew and windowNew (EDIT: DONE). For the interface implementation, I could use a macro, but honestly its not that much duplicated code here. Probably not worth it due to readibility issues imo. What do you think? Will do what you think best :)

### 6. panelNew has a lot of bool parameters
Done. Also renamed "options" to "preferences" to pattern match.

### 7. Dodcs
Done.

Thank you again!!!!

v3/pkg/application/webview_window_darwin.h
```diff
@@ -18,6 +18,18 @@
 
 @end
 
+@interface WebviewPanel : NSPanel
```
leaanthony: I think this should live in `webview_panel_darwin.h`

v3/pkg/application/webview_window_darwin.m
```diff
@@ -243,6 +210,59 @@ - (void)performZoomReset:(id)sender {
     }
 }
 @end
+
+@implementation WebviewPanel
```
leaanthony: I think this should live in `webview_panel_darwin.h`

leaanthony (commented):
Thanks for your patience on this 🙏 

I've added a couple of comments on filenames but not seeing any major issues here!

v3/pkg/application/webview_window_darwin.m
Grantmartin2002: hmmm, we will have to duplicate dispatchKeyDownEvent in that case. are you ok with that?

Grantmartin2002: also, i consolidated windowPanel and windowNew to just use windowNew with a flag check to avoid having to maintain two methods if/when we add properties to general window behavior. With the change you propose, I'd also have to make a webview_panel_darwin.go right? So i'd have to separate the methods again and duplicate more logic.

atterpac: Not having the dupes makes sense to me

leaanthony: If they are using the same build flags then it won't be dupes? I'm not sure I understand the issue here. I was only suggesting moving some of the code to different named files.

leaanthony:
Hey @grantmartin2002 — the "dupes" concern can be resolved without any code duplication. The two helper functions (`keyStringFromKeyEvent` and `dispatchKeyDownEvent`) are `static`, which is why splitting files seemed to require copying them. The fix is to remove `static`, declare them in the shared header, and the panel file can call them via the import.

I've pushed a branch showing the approach: `wailsapp/wails:v3-alpha-feature/issues-5023`

Changes:
- `webview_panel_darwin.h` — new file with `@interface WebviewPanel : NSPanel`
- `webview_panel_darwin.m` — new file with `@implementation WebviewPanel`, imports `webview_window_darwin.h` and calls `dispatchKeyDownEvent` directly (no copy)
- `webview_window_darwin.h` — removed `WebviewPanel` interface, added declarations for the two shared helpers
- `webview_window_darwin.m` — removed `static` from both helpers, removed `@implementation WebviewPanel`
- `webview_window_darwin.go` — added `#include "webview_panel_darwin.h"` to the CGo preamble

Feel free to pull those commits into your fork if you want to keep this PR going, or I can merge from there.

leaanthony:
This PR has a merge conflict with the target branch and needs to be rebased.

Also, I see this is a significant feature addition (287 additions, 144 deletions). Could you clarify: is this intended for v3 only, or should there be a v2 equivalent? And are there any breaking changes or migration considerations for existing applications?

leaanthony:
Thanks again for the substantial work here, @Grantmartin2002. This PR established much of the direction carried forward: an opt-in window class, structured panel preferences, a dedicated `NSPanel` subclass, and the Spotlight-style use case.

We rebuilt the change on current master in #6008 so it could use the current window lifecycle and accelerator paths, preserve `NSWindow` as the zero-value default, share the existing WebKit implementation, and cover window-level precedence, close/recreate behavior, documentation, examples, and manual tests.

To keep review focused on one current implementation, I am closing this PR as superseded by #6008. The successor explicitly credits this contribution and investigation. If you have time, a review of #6008 would be very welcome. Thank you for pushing this feature forward and working through the earlier review feedback.

PR #5026, opened 2026-02-28
docs: add code sample for event usage

iamhabbeboy:
# Description

### Summary of the Change

Added a sample code example demonstrating how to use runtime events for triggering and listening to events in the frontend.

### Issue Fixed

There was no clear example explaining how to use runtime events between the Go backend and the frontend. This change improves documentation clarity by providing a practical starting point for new users.

### Motivation and Context

As a new user to Wails, it took time figuring out the right path to the runtime package on the frontend.

To reduce confusion for beginners and provide a practical reference, a simple event usage example was added to demonstrate:

- How to emit events from the Go backend

- How to listen to events in the frontend

- How both sides communicate using runtime events

## Summary by CodeRabbit

* **Documentation**
  * Added a Code Sample section to the EventsEmit reference with practical frontend examples for listening to and stopping an "auth:success" event and a UI trigger to emit it.
  * Included a Go example demonstrating how to emit an "auth:success" event with a sample payload.
  * Updated the changelog to record these documentation additions.


leaanthony:
Thanks for this. You've added it to a locked version which may get regenerated. Could you please add it to the main "latest" docs so it's preserved. Please could you also add an entry to the changelog located at `website/src/pages/changelog.mdx`?

themavik (commented):
The React sample wiring `EventsOn` in `useEffect` with teardown via `EventsOff` matches the usual leak-avoidance pattern for `events.mdx`.

nit: if your runtime overload requires the same function reference for `EventsOff`, mirror that in the snippet; if name-only unsubscribe is intentional, a half-line note would preempt confusion.

PR #5112, opened 2026-04-03
fix(darwin): propagate backing scale factor to WKWebView for correct devicePixelRatio on Retina displays

### LINKED ISSUE #5111: WKWebView reports devicePixelRatio=1 on macOS Retina displays (custom URL scheme)
### Description

In Wails v2.12.0 on macOS, window.devicePixelRatio returns 1 inside the WKWebView on Retina displays, where it should return 2. This causes all canvas-based rendering (<canvas>, xterm.js WebGL/Canvas, chart libraries) to render at half resolution — macOS upscales the 1x content to 2x, producing visibly blurry output.                         

The root cause is that WKWebView does not automatically propagate the display's backing scale factor to JavaScript's devicePixelRatio when content is loaded via a custom URL scheme (wails://). This affects both wails dev and wails build.

This results in visibly blurry text and graphics in any app that uses `<canvas>` — terminal emulators (xterm.js), chart libraries, drawing tools, games, and code editors all render at half the expected resolution. Text on canvas is particularly affected, appearing soft and hard to read compared to native DOM text.                                                                                                              
                                                                                                 
`window.matchMedia('(-webkit-min-device-pixel-ratio: 2)')` also returns false. 

### To Reproduce

 1. Create a new Wails project on a Mac with a Retina display:
                                                                                                                      
```
  wails init -n dpr-test -t vanilla
  cd dpr-test    
```                                                                                                     
                                                                                                   
  2. Edit frontend/index.html — add this inline script before the module script:                                      
                                                                                                   
```
  <div id="dpr-info" style="position:fixed;top:10px;left:10px;background:#000;color:#0f0;padding:10px;font-family:mono
  space;font-size:16px;z-index:9999;border-radius:4px;"></div>                                                        
  <script>
      var info = 'devicePixelRatio: ' + window.devicePixelRatio + '\n'                                                
          + 'matchMedia 2x: ' + window.matchMedia('(-webkit-min-device-pixel-ratio: 2)').matches + '\n'               
          + 'screen: ' + screen.width + 'x' + screen.height;                                                          
      document.getElementById('dpr-info').innerText = info;                                                           
  </script>              
```                                                                                             
                                                                                                                      
  3. Build and open:                                                                               

```
  wails build
  open build/bin/dpr-test.app          
```                                                                               
   
  4. The overlay shows devicePixelRatio: 1 (also reproducible with wails dev)

### Expected behaviour

`window.devicePixelRatio` should return 2 on Retina displays (matching NSScreen.mainScreen.backingScaleFactor).
Canvas-based content should render at native resolution.  

### Screenshots

Before Fix bug:

<img width="1025" height="795" alt="Image" src="https://github.com/user-attachments/assets/cf8e7228-755f-4bcb-a960-f79eaa8271a9" />

After fix:

<img width="2052" height="1594" alt="Image" src="https://github.com/user-attachments/assets/2a987f04-5a0a-4d81-af1a-6bb845002b00" />

Another example with blur:

<img width="1011" height="168" alt="Image" src="https://github.com/user-attachments/assets/a7ec49ff-f33d-4c9b-9962-03a9a4c6d21b" />

Without blur after fix:

<img width="1009" height="250" alt="Image" src="https://github.com/user-attachments/assets/af6c1b5c-8e65-4697-831e-040e467af171" />

### Attempted Fixes

- **`NSHighResolutionCapable` in Info.plist:** Already set to `true` — does not affect WKWebView's DPR reporting.
- **JavaScript `Object.defineProperty` override of `devicePixelRatio`:** Forces canvas to render at 2x, but the WKWebView compositor still operates at 1x internally. Creates glyph rendering artifacts at small font sizes due to sub-pixel alignment mismatch.
- **`wantsLayer = YES` + `layer.contentsScale = backingScaleFactor` on the WKWebView:** Does not affect `window.devicePixelRatio` — `contentsScale` controls the CALayer rendering, not WebKit's internal device scale factor.
- **`_setOverrideDeviceScaleFactor:` SPI on WKWebView (working fix):** Sets `WebPageProxy::customDeviceScaleFactor`, which is the authoritative source for `window.devicePixelRatio` in JavaScript. This SPI is stable since macOS 10.11 and used by Electron, Playwright, and WebKit's own test infrastructure. I have a working patch and can submit a PR.


### System Details

```shell
Wails Doctor          
                                

                                                                                                                      
# Wails
Version | v2.11.0


# System
┌────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
| OS           | MacOS                                                                                                                   |
| Version      | 15.7.4                                                                                                                  |
| ID           | 24G517                                                                                                                  |
| Branding     |                                                                                                                         |
| Go Version   | go1.25.6                                                                                                                |
| Platform     | darwin                                                                                                                  |
| Architecture | arm64                                                                                                                   |
| CPU 1        | Apple M3                                                                                                                |
| CPU 2        | Apple M3                                                                                                                |
| GPU          | Chipset Model: Apple M3 Type: GPU Bus: Built-In Total Number of Cores: 10 Vendor: Apple (0x106b) Metal Support: Metal 3 |
| Memory       | 24GB                                                                                                                    |
└────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘

# Dependencies
┌────────────────────────────────────────────────────────────────┐
| Dependency                | Package Name | Status    | Version |
| Xcode command line tools  | N/A          | Installed | 2410    |
| Nodejs                    | N/A          | Installed | 22.18.0 |
| npm                       | N/A          | Installed | 10.9.3  |
| *Xcode                    | N/A          | Available |         |
| *upx                      | N/A          | Available |         |
| *nsis                     | N/A          | Available |         |
|                                                                |
└─────────────────── * - Optional Dependency ────────────────────┘

# Diagnosis
  - Xcode: Available at https://apps.apple.com/us/app/xcode/id497799835
```

### Additional context

 Affects any Wails v2 app using `<canvas>` on macOS Retina: terminals (xterm.js), charts, games, editors, drawing apps.
- The issue is specific to the custom `wails://` URL scheme. WKWebView correctly reports DPR when loading content via `http://` or `https://`.
- The fix is in `v2/internal/frontend/desktop/darwin/WailsContext.m`, after the WKWebView is created.

wblech:
# Description

Fix `window.devicePixelRatio` returning `1` inside WKWebView on macOS Retina displays. This causes all canvas-based rendering (xterm.js, charts, games, editors) to render at half resolution, producing visibly blurry text and graphics.

The root cause is that WKWebView does not propagate the display's backing scale factor to JavaScript when content is loaded via a custom URL scheme (`wails://`).

The fix uses the `_setOverrideDeviceScaleFactor:` SPI to set `WebPageProxy::customDeviceScaleFactor`, which is the authoritative source for `window.devicePixelRatio` in JavaScript. This SPI is stable since macOS 10.11 and used by Electron, Playwright, and WebKit's own test infrastructure.

Fixes #5111

## Type of change

# How Has This Been Tested?

1. Created a minimal Wails vanilla app with a DPR overlay (`window.devicePixelRatio` displayed on screen)
2. Built with `wails build` on macOS with Retina display
3. **Before fix:** overlay shows `devicePixelRatio: 1`, `matchMedia 2x: false`
4. **After fix:** overlay shows `devicePixelRatio: 2`, `matchMedia 2x: true`
5. Verified canvas-based rendering (xterm.js) is visibly crisper after the fix

This is a macOS-only change (Objective-C in the darwin frontend). Windows and Linux are not affected.

- [ ] Windows
- [x] macOS
- [ ] Linux

## Test Configuration

                                
          Wails Doctor          
                                

                                                                                                                      
# Wails
Version | v2.11.0

# System
┌────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
| OS           | MacOS                                                                                                                   |
| Version      | 15.7.4                                                                                                                  |
| ID           | 24G517                                                                                                                  |
| Branding     |                                                                                                                         |
| Go Version   | go1.25.6                                                                                                                |
| Platform     | darwin                                                                                                                  |
| Architecture | arm64                                                                                                                   |
| CPU 1        | Apple M3                                                                                                                |
| CPU 2        | Apple M3                                                                                                                |
| GPU          | Chipset Model: Apple M3 Type: GPU Bus: Built-In Total Number of Cores: 10 Vendor: Apple (0x106b) Metal Support: Metal 3 |
| Memory       | 24GB                                                                                                                    |
└────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘

# Dependencies
┌────────────────────────────────────────────────────────────────┐
| Dependency                | Package Name | Status    | Version |
| Xcode command line tools  | N/A          | Installed | 2410    |
| Nodejs                    | N/A          | Installed | 22.18.0 |
| npm                       | N/A          | Installed | 10.9.3  |
| *Xcode                    | N/A          | Available |         |
| *upx                      | N/A          | Available |         |
| *nsis                     | N/A          | Available |         |
|                                                                |
└─────────────────── * - Optional Dependency ────────────────────┘

# Diagnosis
  - Xcode: Available at https://apps.apple.com/us/app/xcode/id497799835

 ♥   If Wails is useful to you or your company, please consider sponsoring the project:

# Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed blurry canvas rendering on macOS Retina and other high‑DPI displays by ensuring web content receives the correct device pixel ratio.
* **New Features**
  * Added a new macOS option to opt in to the Retina/devicePixelRatio correction so high‑resolution canvases render sharply when desired.

### DIFF
--- a/v2/internal/frontend/desktop/darwin/WailsContext.m
+++ b/v2/internal/frontend/desktop/darwin/WailsContext.m
@@ -124,7 +124,21 @@ extern void didReceiveNotificationResponse(const char *jsonPayload, const char* 
 
 }
 
+- (void) windowDidChangeBackingProperties:(NSNotification *)notification {
+    if (!self.enableRetinaDevicePixelRatio) return;
+
+    NSNumber *oldFactor = [notification.userInfo objectForKey:NSBackingPropertyOldScaleFactorKey];
+    CGFloat newFactor = [self.mainWindow backingScaleFactor];
+    if (oldFactor && [oldFactor doubleValue] == newFactor) return;
+
+    SEL sel = NSSelectorFromString(@"_setOverrideDeviceScaleFactor:");
+    if ([self.webview respondsToSelector:sel]) {
+        [self.webview _setOverrideDeviceScaleFactor:newFactor];
+    }
+}
+
 - (void) dealloc {
+    [[NSNotificationCenter defaultCenter] removeObserver:self];
     [self.appdelegate release];
     [self.mainWindow release];
     [self.mouseEvent release];
@@ -318,6 +332,11 @@ extern void didReceiveNotificationResponse(const char *jsonPayload, const char* 
         SEL sel = NSSelectorFromString(@"_setOverrideDeviceScaleFactor:");
         if ([self.webview respondsToSelector:sel]) {
             [self.webview _setOverrideDeviceScaleFactor:[self.mainWindow backingScaleFactor]];
+            self.enableRetinaDevicePixelRatio = YES;
+            [[NSNotificationCenter defaultCenter] addObserver:self
+                selector:@selector(windowDidChangeBackingProperties:)
+                name:NSWindowDidChangeBackingPropertiesNotification
+                object:self.mainWindow];
         }
     }
 
--- a/v2/internal/frontend/desktop/darwin/WailsContext.m
+++ b/v2/internal/frontend/desktop/darwin/WailsContext.m
@@ -157,7 +157,7 @@ extern void didReceiveNotificationResponse(const char *jsonPayload, const char* 
     return NO;
 }
 
-- (void) CreateWindow:(int)width :(int)height :(bool)frameless :(bool)resizable :(bool)zoomable :(bool)fullscreen :(bool)fullSizeContent :(bool)hideTitleBar :(bool)titlebarAppearsTransparent :(bool)hideTitle :(bool)useToolbar :(bool)hideToolbarSeparator :(bool)webviewIsTransparent :(bool)hideWindowOnClose :(NSString*)appearance :(bool)windowIsTranslucent :(int)minWidth :(int)minHeight :(int)maxWidth :(int)maxHeight :(bool)fraudulentWebsiteWarningEnabled :(struct Preferences)preferences :(bool)enableDragAndDrop :(bool)disableWebViewDragAndDrop :(bool)disableEscapeExitsFullscreen {
+- (void) CreateWindow:(int)width :(int)height :(bool)frameless :(bool)resizable :(bool)zoomable :(bool)fullscreen :(bool)fullSizeContent :(bool)hideTitleBar :(bool)titlebarAppearsTransparent :(bool)hideTitle :(bool)useToolbar :(bool)hideToolbarSeparator :(bool)webviewIsTransparent :(bool)hideWindowOnClose :(NSString*)appearance :(bool)windowIsTranslucent :(int)minWidth :(int)minHeight :(int)maxWidth :(int)maxHeight :(bool)fraudulentWebsiteWarningEnabled :(struct Preferences)preferences :(bool)enableDragAndDrop :(bool)disableWebViewDragAndDrop :(bool)disableEscapeExitsFullscreen :(bool)enableRetinaDevicePixelRatio {
     NSWindowStyleMask styleMask = 0;
 
     if( !frameless ) {
@@ -309,13 +309,16 @@ extern void didReceiveNotificationResponse(const char *jsonPayload, const char* 
     [self.webview setFrame:contentViewBounds];
 
     // Force WKWebView to report correct devicePixelRatio on HiDPI displays.
-    // The _overrideDeviceScaleFactor SPI sets WebPageProxy's customDeviceScaleFactor,
-    // which is the authoritative source for window.devicePixelRatio in JavaScript.
-    // Without this, custom URL schemes (wails://) cause DPR to report 1 on Retina.
-    // This SPI is stable since macOS 10.11 and used by Electron and Playwright.
-    CGFloat scaleFactor = [self.mainWindow backingScaleFactor];
-    if (scaleFactor > 1.0) {
-        [self.webview _setOverrideDeviceScaleFactor:scaleFactor];
+    // Uses the _setOverrideDeviceScaleFactor: SPI which sets WebPageProxy's
+    // customDeviceScaleFactor — the authoritative source for window.devicePixelRatio
+    // in JavaScript. Without this, custom URL schemes (wails://) cause DPR to
+    // report 1 on Retina. This is opt-in because it uses a private Apple API
+    // which may cause App Store rejection.
+    if (enableRetinaDevicePixelRatio) {
+        SEL sel = NSSelectorFromString(@"_setOverrideDeviceScaleFactor:");
+        if ([self.webview respondsToSelector:sel]) {
+            [self.webview _setOverrideDeviceScaleFactor:[self.mainWindow backingScaleFactor]];
+        }
     }
 
     if (webviewIsTransparent) {
--- a/v2/internal/frontend/desktop/darwin/WailsContext.m
+++ b/v2/internal/frontend/desktop/darwin/WailsContext.m
@@ -16,6 +16,10 @@
 #import "message.h"
 #import "Role.h"
 
+@interface WKWebView (WailsPrivate)
+@property (nonatomic, setter=_setOverrideDeviceScaleFactor:) CGFloat _overrideDeviceScaleFactor;
+@end
+
 typedef void (^schemeTaskCaller)(id<WKURLSchemeTask>);
 
 @implementation WailsWindow
@@ -303,6 +307,16 @@ extern void didReceiveNotificationResponse(const char *jsonPayload, const char* 
     [self.webview setAutoresizingMask: NSViewWidthSizable|NSViewHeightSizable];
     CGRect contentViewBounds = [contentView bounds];
     [self.webview setFrame:contentViewBounds];
+
+    // Force WKWebView to report correct devicePixelRatio on HiDPI displays.
+    // The _overrideDeviceScaleFactor SPI sets WebPageProxy's customDeviceScaleFactor,
+    // which is the authoritative source for window.devicePixelRatio in JavaScript.
+    // Without this, custom URL schemes (wails://) cause DPR to report 1 on Retina.
+    // This SPI is stable since macOS 10.11 and used by Electron and Playwright.
+    CGFloat scaleFactor = [self.mainWindow backingScaleFactor];
+    if (scaleFactor > 1.0) {
+        [self.webview _setOverrideDeviceScaleFactor:scaleFactor];
+    }
 
     if (webviewIsTransparent) {
         [self.webview setValue:[NSNumber numberWithBool:!webviewIsTransparent] forKey:@"drawsBackground"];
--- a/v2/internal/frontend/desktop/darwin/Application.h
+++ b/v2/internal/frontend/desktop/darwin/Application.h
@@ -17,7 +17,7 @@
 #define WindowStartsMinimised 2
 #define WindowStartsFullscreen 3
 
-WailsContext* Create(const char* title, int width, int height, int frameless, int resizable, int zoomable, int fullscreen, int fullSizeContent, int hideTitleBar, int titlebarAppearsTransparent, int hideTitle, int useToolbar, int hideToolbarSeparator, int webviewIsTransparent, int alwaysOnTop, int hideWindowOnClose, const char *appearance, int windowIsTranslucent, int contentProtection, int devtoolsEnabled, int defaultContextMenuEnabled, int windowStartState, int startsHidden, int minWidth, int minHeight, int maxWidth, int maxHeight, bool fraudulentWebsiteWarningEnabled, struct Preferences preferences, int singleInstanceLockEnabled, const char* singleInstanceUniqueId, bool enableDragAndDrop, bool disableWebViewDragAndDrop, int disableEscapeExitsFullscreen);
+WailsContext* Create(const char* title, int width, int height, int frameless, int resizable, int zoomable, int fullscreen, int fullSizeContent, int hideTitleBar, int titlebarAppearsTransparent, int hideTitle, int useToolbar, int hideToolbarSeparator, int webviewIsTransparent, int alwaysOnTop, int hideWindowOnClose, const char *appearance, int windowIsTranslucent, int contentProtection, int devtoolsEnabled, int defaultContextMenuEnabled, int windowStartState, int startsHidden, int minWidth, int minHeight, int maxWidth, int maxHeight, bool fraudulentWebsiteWarningEnabled, struct Preferences preferences, int singleInstanceLockEnabled, const char* singleInstanceUniqueId, bool enableDragAndDrop, bool disableWebViewDragAndDrop, int disableEscapeExitsFullscreen, int enableRetinaDevicePixelRatio);
 void Run(void*, const char* url);
 
 void SetTitle(void* ctx, const char *title);
--- a/v2/internal/frontend/desktop/darwin/Application.m
+++ b/v2/internal/frontend/desktop/darwin/Application.m
@@ -14,7 +14,7 @@
 #import "WailsMenu.h"
 #import "WailsMenuItem.h"
 
-WailsContext* Create(const char* title, int width, int height, int frameless, int resizable, int zoomable, int fullscreen, int fullSizeContent, int hideTitleBar, int titlebarAppearsTransparent, int hideTitle, int useToolbar, int hideToolbarSeparator, int webviewIsTransparent, int alwaysOnTop, int hideWindowOnClose, const char *appearance, int windowIsTranslucent, int contentProtection, int devtoolsEnabled, int defaultContextMenuEnabled, int windowStartState, int startsHidden, int minWidth, int minHeight, int maxWidth, int maxHeight, bool fraudulentWebsiteWarningEnabled, struct Preferences preferences, int singleInstanceLockEnabled, const char* singleInstanceUniqueId, bool enableDragAndDrop, bool disableWebViewDragAndDrop, int disableEscapeExitsFullscreen) {
+WailsContext* Create(const char* title, int width, int height, int frameless, int resizable, int zoomable, int fullscreen, int fullSizeContent, int hideTitleBar, int titlebarAppearsTransparent, int hideTitle, int useToolbar, int hideToolbarSeparator, int webviewIsTransparent, int alwaysOnTop, int hideWindowOnClose, const char *appearance, int windowIsTranslucent, int contentProtection, int devtoolsEnabled, int defaultContextMenuEnabled, int windowStartState, int startsHidden, int minWidth, int minHeight, int maxWidth, int maxHeight, bool fraudulentWebsiteWarningEnabled, struct Preferences preferences, int singleInstanceLockEnabled, const char* singleInstanceUniqueId, bool enableDragAndDrop, bool disableWebViewDragAndDrop, int disableEscapeExitsFullscreen, int enableRetinaDevicePixelRatio) {
 
     [NSApplication sharedApplication];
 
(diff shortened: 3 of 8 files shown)


v2/internal/frontend/desktop/darwin/WailsContext.m
```diff
@@ -289,6 +293,16 @@ - (void) CreateWindow:(int)width :(int)height :(bool)frameless :(bool)resizable
...
     [self.webview setFrame:contentViewBounds];
 
+    // Force WKWebView to report correct devicePixelRatio on HiDPI displays.
+    // The _overrideDeviceScaleFactor SPI sets WebPageProxy's customDeviceScaleFactor,
+    // which is the authoritative source for window.devicePixelRatio in JavaScript.
+    // Without this, custom URL schemes (wails://) cause DPR to report 1 on Retina.
+    // This SPI is stable since macOS 10.11 and used by Electron and Playwright.
+    CGFloat scaleFactor = [self.mainWindow backingScaleFactor];
+    if (scaleFactor > 1.0) {
+        [self.webview _setOverrideDeviceScaleFactor:scaleFactor];
```
atterpac: since this is a private API it complicates things a bit, it likely needs to be opt in as using private APIs could cause an app to be rejected from the app store and we dont want to force developers into that.

We also should add a guard to prevent apple changing/renaming to break wails windows
```go
  if ([self.webview respondsToSelector:@selector(_setOverrideDeviceScaleFactor:)]) {
      [self.webview _setOverrideDeviceScaleFactor:scaleFactor];
  }
```

wblech: @atterpac  Thanks for the review! I've addressed both points:

## 1. `respondsToSelector:` guard

Added a runtime guard using `NSSelectorFromString` to create the selector at runtime, ensuring the `respondsToSelector:` check works correctly without interference from the category declaration:

```objc
if (enableRetinaDevicePixelRatio) {
    SEL sel = NSSelectorFromString(@"_setOverrideDeviceScaleFactor:");
    if ([self.webview respondsToSelector:sel]) {
        [self.webview _setOverrideDeviceScaleFactor:[self.mainWindow backingScaleFactor]];
    }
}
```

If Apple removes the SPI in a future macOS version, the guard returns `NO` and the call is skipped — no crash.

## 2. Opt-in via `mac.Options`

Added `EnableRetinaDevicePixelRatio` to `mac.Options`. The fix is disabled by default — developers must explicitly opt in:

```go
Mac: &mac.Options{
    EnableRetinaDevicePixelRatio: true,
},
```

This keeps apps targeting the Mac App Store safe from rejection.

## Test results (macOS 15.7.4, Apple M3, Retina display)

| Scenario | devicePixelRatio | matchMedia 2x |
|---|---|---|
| Flag disabled (default) | 1 | false |
| Flag enabled | 2 | true |

## How to replicate

1. Create a vanilla Wails project and point `go.mod` to this branch:
```
replace github.com/wailsapp/wails/v2 => /path/to/this/branch/v2
```

2. Add DPR overlay to `frontend/index.html` (before the module script):
```html
<div id="dpr-info" style="position:fixed;top:10px;left:10px;background:#000;color:#0f0;padding:10px;font-family:monospace;font-size:16px;z-index:9999;border-radius:4px;"></div>
<script>
    var info = 'devicePixelRatio: ' + window.devicePixelRatio + '\n'
        + 'matchMedia 2x: ' + window.matchMedia('(-webkit-min-device-pixel-ratio: 2)').matches + '\n'
        + 'screen: ' + screen.width + 'x' + screen.height;
    document.getElementById('dpr-info').innerText = info;
</script>
```

3. Enable the flag in `main.go`:
```go
Mac: &mac.Options{
    EnableRetinaDevicePixelRatio: true,
},
```

4. Build and open:
```bash
wails build && open build/bin/<app-name>.app
```

### Wails Doctor

```shell

          Wails Doctor



# Wails
Version | v2.11.0


# System
┌────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
| OS           | MacOS                                                                                                                   |
| Version      | 15.7.4                                                                                                                  |
| ID           | 24G517                                                                                                                  |
| Branding     |                                                                                                                         |
| Go Version   | go1.25.6                                                                                                                |
| Platform     | darwin                                                                                                                  |
| Architecture | arm64                                                                                                                   |
| CPU 1        | Apple M3                                                                                                                |
| CPU 2        | Apple M3                                                                                                                |
| GPU          | Chipset Model: Apple M3 Type: GPU Bus: Built-In Total Number of Cores: 10 Vendor: Apple (0x106b) Metal Support: Metal 3 |
| Memory       | 24GB                                                                                                                    |
└────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘

# Dependencies
┌────────────────────────────────────────────────────────────────┐
| Dependency                | Package Name | Status    | Version |
| Xcode command line tools  | N/A          | Installed | 2410    |
| Nodejs                    | N/A          | Installed | 22.18.0 |
| npm                       | N/A          | Installed | 10.9.3  |
| *Xcode                    | N/A          | Available |         |
| *upx                      | N/A          | Available |         |
| *nsis                     | N/A          | Available |         |
|                                                                |
└─────────────────── * - Optional Dependency ────────────────────┘

# Diagnosis
  - Xcode: Available at https://apps.apple.com/us/app/xcode/id497799835

```

leaanthony:
✅ **Triaged by Wails PR Reviewer**

This PR has been reviewed and accepted. A test sub-issue has been created for macOS.

**Head Ref OID: `7d3c012a83d0d303e9977cfa43c05f496b97ccf0`**

---
*This comment serves as a signature that this PR has been triaged. Future runs will skip this PR based on the headRefOid.*

leaanthony:
## macOS CI test result — ✅ PASS

**Commit tested:** `7d3c012a83d0d303e9977cfa43c05f496b97ccf0`

### Environment

| | |
|---|---|
| macOS | 26.3 (Build 25D125) |
| Architecture | arm64 (Apple Silicon) |
| Go | 1.26.1 |
| Display | Non-Retina VM (backingScaleFactor = 1.0) |

### Build

Compiled cleanly. One **pre-existing** deprecation warning unrelated to this PR:
```
WailsContext.m:177: 'setShowsBaselineSeparator:' deprecated in macOS 15.0
```
No new warnings from this PR's changes.

### Unit tests

All 13 v2 test packages pass (`go test ./... -count=1 -timeout 180s`).

### Visual smoke test

Built and ran a minimal Wails v2 app with `EnableRetinaDevicePixelRatio: true`.

**Result: `devicePixelRatio = 1` — correct.**

The VM has a non-Retina display (`backingScaleFactor = 1.0`), so the correct DPR is 1. The PR correctly propagates this value from `backingScaleFactor`. On a real Retina Mac (`backingScaleFactor = 2.0`), DPR would report 2.0 — which is the fix this PR provides.

The smoke run produced two screenshots:
- **screenshot-001**: App window with `devicePixelRatio = 1` displayed prominently, confirming the value is read from `backingScaleFactor`
- **screenshot-002**: Build terminal output alongside the running app window

The `respondsToSelector:` guard for `_setOverrideDeviceScaleFactor:` was verified working on macOS 26.3 — no exceptions, no crash.

### Code review notes (non-blocking)

- The `@interface WKWebView (WailsPrivate)` private API extension is correctly gated behind `respondsToSelector:` — forward-compatible with future macOS versions.
- The option is opt-in with a clear App Store warning in the doc comment.
- The `enableRetinaDevicePixelRatio` parameter is correctly threaded through the full CGo call chain.

**Verdict: ✅ PASS on macOS arm64**

leaanthony:
As @atterpac said, this is using a private API and that's strictly a no go, beyond providing an opt in build tag. What are your thoughts on that @wblech?

wblech:
Sorry for the delay. I got into a point where I was reviewing the need of this fix. But I had to stop because of my normal job. I will try to get back to this.

leaanthony:
This looks like a genuine fix for a real and painful bug — the opt-in approach via `EnableRetinaDevicePixelRatio` is the right call given the App Store concern with `_setOverrideDeviceScaleFactor:`.

A few things needed before promoting from draft:

1. **macOS verification** — the PR notes the author doesn't have a macOS environment to test the CGo changes. Needs confirmation from a macOS owner before merge.

2. **v3 equivalent** — this changes `v2/internal/frontend/desktop/darwin/`. Does v3's darwin frontend have the same `devicePixelRatio` bug? If so, a v3 port should accompany or follow this.

3. **Tests** — no automated tests exist for the scale factor propagation. Even a minimal ObjC mock would help.

Happy to see this land once those three points are addressed.

v2/internal/frontend/desktop/darwin/WailsContext.m
leaanthony: Both points fully addressed by the author in their follow-up (see above):
1. **`respondsToSelector:` guard** — implemented via `NSSelectorFromString` so the check works correctly even alongside the category declaration. If Apple removes the SPI, the guard silently skips the call — no crash.
2. **Opt-in flag** — `EnableRetinaDevicePixelRatio` added to `mac.Options`, disabled by default. Apps targeting the Mac App Store are safe unless they explicitly opt in.

Thanks for flagging these — the opt-in approach was the right call.

leaanthony:
Hi @wblech — this PR has developed a merge conflict since it was last updated. Could you please rebase onto current `master`?

The conflict is most likely in `website/src/pages/changelog.mdx` or one of the Darwin files — master has had several changes in those areas since your branch was cut.

```bash
git fetch upstream
git rebase upstream/master
# resolve any conflicts
git push --force-with-lease
```

Once rebased, this is ready to merge — the implementation looks solid.

leaanthony:
Placing this on hold — the implementation relies on the private SPI `_setOverrideDeviceScaleFactor:`. While the `respondsToSelector:` guard prevents crashes if Apple removes it, use of private APIs can result in App Store rejection regardless of whether the opt-in flag is set, since Apple scans binaries for SPI calls at submission time.

We need to investigate whether there's a public API path to achieve the same result before this can merge.

wblech:
# PR #5112 — Changes made in response to review comments

@leaanthony  — thank you for the time and the thorough analysis on this PR. I
really like this project and intend to keep contributing whenever I can. That is
exactly why I did a second review pass with the corrections described here. Before
merging, please read at least the TL;DR below.

## TL;DR

After reviewing everything again, I have the same feeling I had when I decided to
leave this PR as a draft — that this fix may not actually be necessary.

In my initial tests the problem really was there: the blur and the
`devicePixelRatio` divergence reported in #5111 did happen, and with these changes
they were resolved. But during a second review I found that the issue no longer
reproduces — even with the flag **disabled**. I honestly don't know why it stopped
happening (a macOS/WebKit update, something in my environment — I couldn't pin it
down).

I also confirmed that my test app runs the official released `v2.12.0` (its
`go.mod` has no `replace` pointing at this branch), so it isn't even building with
the fix — and on plain official Wails, with no override, the blur and
`devicePixelRatio` are already correct on my machine. So on my current
macOS/WebKit version the WKWebView already reports the right `devicePixelRatio` on
its own and #5111 does not reproduce.

It's also worth noting that `_setOverrideDeviceScaleFactor:` *forces* a fixed DPR
rather than letting WebKit report it — so when WebKit already reports it correctly,
the override is redundant (and it's why a notification observer is needed to
re-track display changes that WebKit would otherwise handle for free).

Either way, I'm leaving the changes here in case the maintainers can reproduce the
original issue and see a need to merge them. And if this doesn't move forward —
apologies for the unnecessary work. The intent was only to ship a fix that
genuinely made sense to me at the time.

The rest of this document summarizes what was changed based on the feedback on the
PR. For the problem description and the fix itself, see the PR description.

## 1. Opt-in via `mac.Options` — requested by @atterpac

Since `_setOverrideDeviceScaleFactor:` is a private API, the override is now
**disabled by default** and must be explicitly enabled:

```go
Mac: &mac.Options{
    EnableRetinaDevicePixelRatio: true,
},
```

This keeps apps targeting the Mac App Store safe unless they choose to opt in.

## 2. Runtime guard with `respondsToSelector:` — requested by @atterpac and CodeRabbit

The SPI call is gated behind a runtime check built with `NSSelectorFromString`:

```objc
SEL sel = NSSelectorFromString(@"_setOverrideDeviceScaleFactor:");
if ([self.webview respondsToSelector:sel]) {
    [self.webview _setOverrideDeviceScaleFactor:...];
}
```

If Apple removes or renames the SPI in a future macOS version, the guard returns
`NO`, the call is skipped, and the app falls back to previous behavior — no crash.

## 3. Re-apply on display change — suggested by CodeRabbit

CodeRabbit noted that the scale factor is captured only at window creation, so
`devicePixelRatio` could become stale if the window moves to a display with a
different backing scale factor.

Following that suggestion, `WailsContext` now observes
`NSWindowDidChangeBackingPropertiesNotification` and re-applies the override when
the factor actually changes (the observer is removed in `dealloc`). This makes
sense exactly for the window-moving case: dragging a window from a Retina to a
non-Retina display (or vice versa) changes the backing scale factor, and the
handler re-applies the correct value so the reported `devicePixelRatio` stays
accurate across display switches.

## 4. Tests added

A unit test (`window_test.go`) covers the opt-in resolution, asserting the
default-off invariant and nil-safety:

- `nil` Mac options → disabled (no panic)
- Mac options present, flag unset / explicitly `false` → disabled
- flag `true` → enabled

The native `_setOverrideDeviceScaleFactor:` call itself is not unit-tested —
mocking a private WKWebView SPI would test the mock, not WebKit. The cheap,
meaningful coverage is the option plumbing (does the opt-in resolve correctly and
stay off by default), which is what the test verifies. It runs on the macOS CI
job.

## 5. macOS verification

The change was built and verified on a real macOS Retina machine. `wails doctor`
for the test environment:

```
                                
          Wails Doctor          
                                

                                                                                                                                                                                                                                                  
# Wails
Version | v2.11.0


# System
┌────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
| OS           | MacOS                                                                                                                   |
| Version      | 26.4.1                                                                                                                  |
| ID           | 25E253                                                                                                                  |
| Branding     |                                                                                                                         |
| Go Version   | go1.25.6                                                                                                                |
| Platform     | darwin                                                                                                                  |
| Architecture | arm64                                                                                                                   |
| CPU 1        | Apple M3                                                                                                                |
| CPU 2        | Apple M3                                                                                                                |
| GPU          | Chipset Model: Apple M3 Type: GPU Bus: Built-In Total Number of Cores: 10 Vendor: Apple (0x106b) Metal Support: Metal 4 |
| Memory       | 24GB                                                                                                                    |
└────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘

# Dependencies
┌────────────────────────────────────────────────────────────────┐
| Dependency                | Package Name | Status    | Version |
| Xcode command line tools  | N/A          | Installed | 2416    |
| Nodejs                    | N/A          | Installed | 22.18.0 |
| npm                       | N/A          | Installed | 10.9.3  |
| *Xcode                    | N/A          | Available |         |
| *upx                      | N/A          | Available |         |
| *nsis                     | N/A          | Available |         |
|                                                                |
└─────────────────── * - Optional Dependency ────────────────────┘

# Diagnosis
  - Xcode: Available at https://apps.apple.com/us/app/xcode/id497799835

 ♥   If Wails is useful to you or your company, please consider sponsoring the project:
```

## 6. v3

v3 has its own darwin frontend (`v3/pkg/application/webview_window_darwin.m`),
which uses `layer.rasterizationScale` rather than `_setOverrideDeviceScaleFactor:`,
so the same issue may exist there.

I'm not going to touch v3 in this PR — I haven't worked with v3 yet, so I'd rather
not change code I'm not familiar with. I may revisit it as a follow-up in the
future, but not now.

leaanthony:
Thanks for taking the time to go over this again and clarify. Feel free to close it for now but it will still be available through search if we do ever need revisiting it 🙏

Issue #4976, opened 2026-02-08
[V3]Linux Mint Frameless not work


youcode22:
React SPA
global css:
```css
body {
  margin: 0;
  padding: 0;
  width: 100vw;
  height: 100vh;
  --wails-resize: all;
  font-family:
    -apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto", "Helvetica Neue",
    "Arial", sans-serif, "Apple Color Emoji", "Segoe UI Emoji",
    "Segoe UI Symbol";
}
```
resize on all edges and tested specific edges like v3 doc,all not work.
tested use defaut react-ts template, not work too.

wails3 doctor:
```shell
Wails (v3.0.0-alpha.64)  Wails Doctor 
                                                                                                                                                                                                                                                
# System 

┌──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
| Name                | Linux Mint                                                                                                     |
| Version             | 22.3                                                                                                           |
| ID                  | linuxmint                                                                                                      |
| Branding            | 22.3 (Zena)                                                                                                    |
| Platform            | linux                                                                                                          |
| Architecture        | amd64                                                                                                          |
| Desktop Environment | X-Cinnamon                                                                                                     |
| NVIDIA Driver       | N/A                                                                                                            |
| XDG_SESSION_TYPE    | x11                                                                                                            |
| CPU                 | AMD Ryzen 5 2600X Six-Core Processor                                                                           |
| GPU 1               | Ellesmere [Radeon RX 470/480/570/570X/580/580X/590] (Advanced Micro Devices, Inc. [AMD/ATI]) - Driver: amdgpu  |
| Memory              | 48GB                                                                                                           |
└──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘

# Build Environment 

┌────────────────────────────────┐
| Wails CLI    | v3.0.0-alpha.64 |
| Go Version   | go1.25.5        |
| -buildmode   | exe             |
| -compiler    | gc              |
| CGO_CFLAGS   |                 |
| CGO_CPPFLAGS |                 |
| CGO_CXXFLAGS |                 |
| CGO_ENABLED  | 1               |
| CGO_LDFLAGS  |                 |
| GOAMD64      | v1              |
| GOARCH       | amd64           |
| GOOS         | linux           |
└────────────────────────────────┘

# Dependencies 

┌────────────────────────────────────────────────────────────────┐
[...]

PR #5159, opened 2026-04-16
fix(v3,linux): implement native frameless window resize

BenAicholtz:
## Summary

`linuxWebviewWindow.startResize` was a no-op stub, so `wails:resize:<edge>` messages sent by the runtime's [`drag.ts`](https://github.com/wailsapp/wails/blob/v3-alpha/v3/internal/runtime/desktop/%40wailsio/runtime/src/drag.ts#L210) were dispatched but never acted upon. Frameless windows on Linux could only be dragged, not resized from edges/corners.

This PR wires both CGo backends up to the underlying GTK resize APIs.

## Changes

- **GTK3** (`linux_cgo.go`): map the eight edge strings to `GdkWindowEdge` and call `gtk_window_begin_resize_drag`, using the mouse state captured by the existing `button-press-event` handler.
- **GTK4** (`linux_cgo_gtk4.go`): same mapping to `GdkSurfaceEdge`, dispatched through the `beginWindowResize` C helper that was already defined in `linux_cgo_gtk4.c` / `.h` but never called.
- **`webview_window_linux.go`**: remove the `// FIXME: what do we need to do here?` stub so the per-backend implementations are the only definitions.

The pattern mirrors the existing `startDrag` implementations on both backends — the `w.drag` state populated by mousedown is reused with no new plumbing.

## Why this matters

- Frameless windows on Linux are common for PiP / floating video overlays / custom chromeless apps. Without resize support the only workarounds are (a) a non-frameless window with system decorations, or (b) a manual JS implementation that round-trips `SetSize`/`SetPosition` over the IPC channel per mousemove — visibly janky compared to native resize.
- The C helper and the `--wails-resize` / `wails:resize:` runtime plumbing already exist; this PR only wires them together.

## Fixes

- #4976

## Test plan

- [ ] Build a frameless window app on GTK3 (default): drag from each of 8 edges/corners → window resizes smoothly and stops on mouseup.
- [ ] Build the same app with `-tags gtk4`: drag from each of 8 edges/corners → same behavior.
- [ ] Confirm that passing an unknown border string returns a clear error (`unknown resize border: %q`) rather than silently doing nothing.
- [ ] Verify the existing `startDrag` behavior is unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added window resize-drag functionality for Linux applications, allowing users to resize windows by dragging from any border edge (north, south, east, west, and corner edges).


leaanthony (approved):
Thanks for taking this on!

PR #5173, opened 2026-04-19
feat(v3): debug package + user-goroutine panic handling

leaanthony:
## Summary

Introduces a Windows-first diagnostic/crash-dump package alongside guidance
and an example for handling panics in user-spawned goroutines.

### New: `v3/pkg/debug`

A small, stateless API for capturing crash diagnostics at runtime. Two
entry points:

- `debug.Dump(...)` — writes a Windows minidump of the current process via
  `Dbghelp.MiniDumpWriteDump`. Uses `GetCurrentProcess()` so **no
  elevation / SeDebugPrivilege is required** — runs under a standard user
  token. Returns the absolute path of the written file. On non-Windows
  platforms it returns a clean unimplemented error so callers can degrade
  gracefully. Options: `WithPath`, `WithFullMemory`.
- `debug.Report(...)` — collects system info, build info, runtime
  process/memory/module state, and a diagnostics pass from
  `pkg/doctor-ng`. Returns a `*CrashReport`. Minidump is **opt-in** via
  `WithDump` / `WithDumpPath` / `WithDumpFullMemory`.

Both entry points are designed to be safe to call from `PanicHandler`:
they stream the minidump directly to a file handle (no Go-side buffering
of megabytes), and `Report` returns a partial report even when the
minidump step fails.

Default dump location is `os.TempDir()` — on Windows,
`%LOCALAPPDATA%\Temp\wails-crash-<pid>-<unix>.dmp`. Writable by the
current user, not elevated, not world-readable.

### New: `v3/examples/user-panic-handling`

Demonstrates funnelling panics from **user-spawned goroutines** through
the same `PanicHandler` wails uses for bound methods. Ships a
`recoverAndReport()` helper that user code defers at the top of every
goroutine it owns. Without it, goroutine panics crash the whole process —
wails can only inject deferred recovers into code paths it owns.

### Docs

- **New guide:** `guides/debugging-crashes.mdx` — full coverage of the
  debug package. Quick reference, `CrashReport` shape, `PanicHandler`
  integration, manual debugging use cases (hang detection, support
  bundles, anomaly reports), minidump flag rationale, non-Windows
  behaviour, and security notes.
- **Updated:** `guides/panic-handling.mdx` — new "Handling Panics in Your
  Own Goroutines" section with the `recoverAndReport` pattern, and a
  "Capturing Diagnostics on Panic" section cross-linking into the new
  debug guide.

### Security notes

- No elevation. No `SeDebugPrivilege`. No `OpenProcess` on remote PIDs.
  Deliberately scoped to the calling process.
- Minidumps contain everything in process memory — tokens, keys, user
  data. The docs call this out so consumers handle dumps with the same
  care as raw memory images.

## Test plan

- [x] `go build ./v3/pkg/debug/...` on linux, windows, darwin (verified locally)
- [x] `go vet ./v3/pkg/debug/...` clean on all three platforms
- [x] `go test ./v3/pkg/debug/...` passes on linux
- [x] Cross-platform test covers: `Report` populates fields, default
      `DumpPath` empty without `WithDump`, `Dump` returns error on
      non-Windows, option composition
- [ ] Run `dump_windows_test.go` on a real Windows host — verifies
      MDMP magic bytes, default path under `TempDir`, `Report(WithDumpPath)`
      writes + records path
- [ ] Open a produced `.dmp` in WinDbg to sanity-check thread/handle/module
      coverage with the default flag set

## Out of scope (follow-ups)

- Wiring `debug.Report` into `PanicHandler` by default — intentionally
  opt-in. Could ship a helper `debug.CrashHandler(userHandler)` later.
- macOS/Linux postmortem equivalents (core dumps via `GOTRACEBACK=crash`,
  CrashReporter, `rr`, etc.) — out of scope for a Windows-first package.


taliesin-ai:
Ported the authored crash-debugging and panic-handling guides to M-Press, stacked on #6137 (validated migration snapshot `7b0f6dc09`; no later migration merge). Includes all nine full translations, localized navigation, corrected links and an M-Press caution. Original debug implementation and user-panic example are unchanged; original commits preserved.

Validation: strict build `--no-purge-css` (2,020 pages), `mpress check`, generated-site checker (2,021 HTML pages, zero errors), `go test ./pkg/debug -count=1`, and scoped diff checks passed. All translated additions preserve the source code blocks, API identifiers and heading counts. Windows runtime behavior was not tested here.

Translation audit remains pending with the upstream compatibility task: 7,723 findings versus the pinned baseline's 7,678 (+45 segment/structure/glossary and stale-hash findings); no exclusions or exemptions changed. CodeRabbit `--plain` was attempted and rejected; the one scoped light fallback still included inherited migration pages and exceeded its file limit. Manual diff review completed; no paid credits used.

PR #5191, opened 2026-04-20
fix(dev): exclude _test.go files from triggering rebuilds

### LINKED ISSUE #3243: wails dev should not rebuild the app when modifying a go test or benchmark
### Description

`wails dev` should not rebuild the app when modifying a _test.go file. The app will restart each time when modifying tests or benchmarks, and it shouldn't.

### To Reproduce

1. create the file `app_test.go`
2. add `func TestFoo(t *testing.T) { fmt.Println("hello") }`
3. run `wails dev`
4. Modify "hello" to "goodbye"
5. The wails app will relaunch.

Note that this also happens in the case of benchmarks (e.g. app_bench_test.go with a function `BenchmarkApp(b *testing.B)`

### Expected behaviour

The wails app should not relaunch.

### Screenshots

_No response_

### Attempted Fixes

_No response_

### System Details

```shell
Wails v2.7.1
MacOS 14.2.1
```

### Additional context

_No response_

leaanthony:
## Summary
- Add `*_test.go` to the default file ignore list in the dev mode config template, preventing unnecessary app rebuilds when editing test or benchmark files.
- Add tests verifying the config template includes the pattern and parses correctly.

Fixes #3243

## Test plan
- [x] `TestDevConfigIgnoresTestGoFiles` — verifies `*_test.go` is present in parsed config
- [x] `TestDevConfigParsesCorrectly` — verifies config with `*_test.go` pattern parses correctly
- [x] `TestDefaultBuildAssetsContainTestGoIgnore` — verifies the template file contains the pattern

### DIFF
--- a/v3/UNRELEASED_CHANGELOG.md
+++ b/v3/UNRELEASED_CHANGELOG.md
@@ -23,6 +23,7 @@ After processing, the content will be moved to the main changelog and this file 
 
 ## Fixed
 <!-- Bug fixes -->
+- Fix `wails3 dev` triggering rebuilds when modifying `_test.go` or benchmark files (#3243)
 
 ## Deprecated
 <!-- Soon-to-be removed features -->
--- a/v3/internal/commands/build_assets/config.yml
+++ b/v3/internal/commands/build_assets/config.yml
@@ -45,6 +45,7 @@ dev_mode:
       - .DS_Store
       - .gitignore
       - .gitkeep
+      - "*_test.go"
     watched_extension:
       - "*.go"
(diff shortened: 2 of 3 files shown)


leaanthony:
Closing as superseded by #5203 which was already merged into master.

#5203 is a superset of this PR:
- Same `config.yml` change (adds `*_test.go` to the ignore file list)
- Additionally injects the pattern at **runtime** via `ensureIgnored()` in `watcher.go`, so existing projects benefit without needing to regenerate their config
- Cleaner tests that unit-test the `ensureIgnored` function directly rather than reading the config file three different ways

The underlying issue #3243 is fixed.

PR #5199, opened 2026-04-21
fix(runtime): prevent event listener restoration after removal during callback

### LINKED ISSUE #4393: Removing an event while handling an event leave the event handler on
### Description

in my JS code, i have something like that;

```js
let ev = runtime.EventsOn( "name", () => {
     // here i call ev that will call runtime.EventsOff( )
     if( some_reason ) { // the reason i do not call EventOnOnce
         ev( ); 
     }
} );
```

the problem is that if i have >1 listener, my listener is not removed:

```js
in wails event.js (line 119)
function notifyListeners(eventData) {
...
       eventListeners[eventName] = newEventListenerList; <<< here original events are restored event if my was removed
...
}
```

### To Reproduce

you saddly have to check this kind of js code

### Expected behaviour

my listener should be still removed

### Screenshots

_No response_

### Attempted Fixes

_No response_

### System Details

```shell
# Wails
Version         | v2.10.1
Package Manager | apt    


# System
┌──────────────────────────────────────────────────────────┐
| OS           | Ubuntu                                    |
| Version      | 24.04                                     |
| ID           | ubuntu                                    |
| Go Version   | go1.24.4                                  |
| Platform     | linux                                     |
| Architecture | amd64                                     |
| CPU          | Intel(R) Xeon(R) CPU E3-1220 v6 @ 3.00GHz |
| Memory       | 31GB                                      |
└──────────────────────────────────────────────────────────┘

# Dependencies
┌───────────────────────────────────────────────────────────────┐
| Dependency | Package Name    | Status    | Version            |
| *docker    | docker.io       | Installed | 28.1.1+1           |
| gcc        | build-essential | Installed | 13.3.0             |
| libgtk-3   | libgtk-3-dev    | Installed | 3.24.41-4ubuntu1.3 |
| libwebkit  | Unknown         | Not Found |                    |
| npm        | npm             | Installed | 11.4.2             |
| *nsis      | nsis            | Available | 3.09-4ubuntu1      |
| pkg-config | pkg-config      | Installed | 1.8.1              |
|                                                               |
└─────────────────── * - Optional Dependency ───────────────────┘
```

### Additional context

_No response_

leaanthony:
## Description

Fixes #4393

When a listener removed itself or other listeners during a callback (via `EventsOff` or the cancel function returned by `EventsOn`), the `notifyListeners` function would overwrite `eventListeners` with the stale copy taken before callbacks ran, restoring the removed listeners.

## Changes

- Updated `notifyListeners` in `events.js` to filter the surviving listener list against the current state of `eventListeners` after all callbacks have run, rather than blindly replacing it
- Updated bundled runtimes (`runtime_debug_desktop.js`, `runtime_prod_desktop.js`)

## Root Cause

The function took a shallow copy of listeners before iterating, but at the end unconditionally replaced `eventListeners[eventName]` with the modified copy. If `listenerOff()` or `EventsOff()` was called during a callback (which directly mutates `eventListeners[eventName]`), those removals were overwritten.

## Testing

- Added test in `v2/test/4393/event_removal_test.js` with 4 test cases covering self-removal, partial removal, full removal, and EventsOnce behavior


taliesin-ai:
Superseded by #5578 (events.js now removes mid-dispatch listeners from the live list). This was an earlier WIP for the same fix; closing as done.

PR #5200, opened 2026-04-21
fix(v2/runtime): wrap string error rejections in Error objects

### LINKED ISSUE #4379: Frontend bindings does not return "Error" object on errors, only strings
### Description

As I am using React with TypeScript, I also like to use TanStack, and I wrap the Wails bindings in useMutation, so a simple one could be defined as easily as:

```
import * as m from '@wailsjs/go/main/App'

export function useSetHello(value: string) {
	return useMutation({
		mutationFn: () => m.SetHello(value)
	})
}
```

But the issue is that when m.SetHello throws an error, the promise reject is just a string, but TanStack expects errors to be of type `Error`. So now I need to wrap every binding with a function that throws a new Error() if it gets an error of type string.

So in short, I would like the bindings to fail the promises with reject(new Error(message)) instead of reject(message) as it seems to do today.

### To Reproduce

create a function in go:

```
func (app *App) SetHello(value string) error {
  return errors.New("some error")
}
```

Create a wrapper in ts:
```
import * as m from '@wailsjs/go/main/App'

export function useSetHello() {
	return useMutation({
		mutationFn: (value: string) => m.SetHello(value)
	})
}
```

Use it in frontend:
```
export default function MyComponent() {
	const helloFunc = useSetHello()

	return (
		<div>
			<button onClick={() => helloFunc.mutate("Hello, World!")}>
				Set Hello
			</button>
			{helloFunc.error && (
				<div className="text-red-500">
					Error: {helloFunc.error.message}
				</div>
			)}
		</div>
	)
}
```
  

### Expected behaviour

That it prints the error, since helloFunc.error is expected to be a type of Error with a message property.

If I use {helloFunc.error}, the editor will complain, as the type of helloFunc.error is Error, not string, as TanStack expects promises to return errors of type Error, not string.

### Screenshots

Image

### Attempted Fixes

I can cast it to unknown and then string, when using it but that is bad programming style. Or I can wrap every binding with a function that converts from string to Error(), but I think it is better for the bindings to return Error, or maybe make it configurable, if you expect that to break things.

### System Details

```shell
# Wails
Version | v2.10.1


# System
| OS           | MacOS                                                                                                                       |
| Version      | 14.4.1                                                                                                                      |
| ID           | 23E224                                                                                                                      |
| Go Version   | go1.24.1                                                                                                                    |
| Platform     | darwin                                                                                                                      |
| Architecture | arm64                                                                                                                       |
| CPU 1        | Apple M1 Pro                                                                                                                |
| CPU 2        | Apple M1 Pro                                                                                                                |
| GPU          | Chipset Model: Apple M1 Pro Type: GPU Bus: Built-In Total Number of Cores: 16 Vendor: Apple (0x106b) Metal Support: Metal 3 |
| Memory       | 16GB                                                                                                                        |

# Dependencies
┌───────────────────────────────────────────────────────────────────────┐
| Dependency                | Package Name | Status    | Version        |
| Xcode command line tools  | N/A          | Installed | 2406           |
| Nodejs                    | N/A          | Installed | 20.9.0         |
| npm                       | N/A          | Installed | 10.1.0         |
| *Xcode                    | N/A          | Installed | 15.3 (15E204a) |
| *upx                      | N/A          | Available |                |
| *nsis                     | N/A          | Installed | v3.11          |
|                                                                       |
└─────────────────────── * - Optional Dependency ───────────────────────┘

# Diagnosis

```

### Additional context

_No response_

leaanthony:
## Summary
Fixes #4379

Promise rejections from Go binding calls now wrap string errors in proper `Error` objects instead of rejecting with raw strings.

## Problem
When a Go method returns an error, the Wails runtime rejects the JS promise with a plain string. Libraries like TanStack Query expect `Error` instances (they access `error.message`), making the bindings incompatible.

## Fix
In `Callback()`, wrap `message.error` in `new Error()` when it's a string, and pass through non-string errors as-is.

## Testing
- `v2/test/4379/calls_test.mjs` — 7 tests, all passing:
  - String error → Error instance with message and stack
  - Successful result still works
  - Non-string error objects are passed through as-is


taliesin-ai:
Superseded by #5244 (calls.js now wraps string rejections in Error objects). This was an earlier WIP for the same fix; closing as done.

PR #5240, opened 2026-04-25
fix(v2): Preserve event listener removal during notification

### LINKED ISSUE #4393: Removing an event while handling an event leave the event handler on

leaanthony:
## Summary

Fixes #4393

When a listener cancels itself via the returned cancel function (from `EventsOn`) during a notification callback, the removal was being overwritten when `notifyListeners` replaced `eventListeners[eventName]` with the stale snapshot copy.

## Problem

The `notifyListeners` function makes a copy of listeners with `.slice()`, iterates through the copy, and then assigns the modified copy back to `eventListeners[eventName]`. If a listener calls its cancel function during the callback, `listenerOff` correctly removes it from the live `eventListeners[eventName]`, but then `notifyListeners` overwrites this with the stale copy, restoring the removed listener.

## Fix

Instead of replacing `eventListeners[eventName]` with the snapshot copy, we now:

1. Iterate through a copy of listeners (unchanged)
2. Collect listeners that should be destroyed (maxCallbacks reached) in a Set
3. After the loop, filter the **live** `eventListeners[eventName]` to remove destroyed listeners

This preserves any removals made via `listenerOff` during callbacks.

## Test Plan

Added 2 new test cases in `events.test.js`:

1. **"should keep listener removed when cancelled during notification"** - Verifies that when listener B cancels itself during a notification, it stays removed on subsequent notifications while listeners A and C continue to work.
2. **"should handle all listeners being removed during notification"** - Verifies that when all listeners cancel themselves during notification, the event is properly cleaned up.

All 13 tests pass (11 existing + 2 new).

---
[:bulb:](mention://agent/d5c5204d-1cc6-4b76-a654-71c642c38fec) *This PR was created by the Wails Engineer agent.*


leaanthony:
🤖 **PR Triage Review**

✅ **Accepted**

This PR preserves event listener removal during notification (v2). Fixes event management bug.

**Platform**: All (v2-only, runtime)

**Next Steps**: Dispatching for testing on all platforms.

---

taliesin-ai:
Superseded by #5578 plus the vitest mock fix in #5579. This was an earlier WIP for the same fix; closing as done.

PR #5249, opened 2026-04-26
fix(windows): prevent Center() from stealing focus when StartHidden:true

### LINKED ISSUE #4882: Opening Wails on Windows with `StartHidden:true` steals focus from current window
### Description

When running a new Wails app on Windows with `StartHidden: true`, the current window becomes unfocused.

### To Reproduce

1. Open any application on your Windows computer
2. Call `wails.Run` with the option `StartHidden: true` (among your other values). Make sure to add a delay before like `time.Sleep(5 * time.Second) to allow you to set up the rest of the bug.
3. Focus your application (that isn't Wails). Make sure it is focused by trying to enter text, for example.
4. The Wails window should now open (in the background), and your application should lose focus. 

### Expected behaviour

When opening with `StartHidden: true`, the current open application does not lose focus

### Screenshots

_No response_

### Attempted Fixes

The root cause is in `internal/frontend/desktop/windows/winc/form.go:171` In `Center()`:

```go
w32.SetWindowPos(fm.hwnd, w32.HWND_TOP, int(windowX), int(windowY), int(winWidth), int(winHeight), w32.SWP_NOSIZE)
```

The `SetWindowPos` call uses `HWND_TOP` but does not include the `SWP_NOACTIVATE` flag. Without this flag, Windows activates the window when positioning it, which steals focus from the foreground application.

Proposed fix:

```go
w32.SetWindowPos(fm.hwnd, w32.HWND_TOP, int(windowX), int(windowY), int(winWidth), int(winHeight), w32.SWP_NOSIZE|w32.SWP_NOACTIVATE)
```

### System Details

```shell
Wails Doctor



# Wails
Version | v2.11.0


# System
┌────────────────────────────────────────────────────────────────────────────────────────┐
| OS           | Windows 10 Pro for Workstations                                         |
| Version      | 2009 (Build: 26200)                                                     |
| ID           | 25H2                                                                    |
| Branding     | Windows 11 Pro for Workstations                                         |
| Go Version   | go1.25.5                                                                |
| Platform     | windows                                                                 |
| Architecture | arm64                                                                   |
| CPU          | Apple Silicon                                                           |
| GPU          | Parallels Display Adapter (WDDM) (Parallels) - Driver: 20.18.2612.57293 |
| Memory       | 8GB                                                                     |
└────────────────────────────────────────────────────────────────────────────────────────┘

# Dependencies
┌────────────────────────────────────────────────────────┐
| Dependency | Package Name | Status    | Version        |
| WebView2   | N/A          | Installed | 143.0.3650.139 |
| Nodejs     | N/A          | Available |                |
| npm        | N/A          | Available |                |
| *upx       | N/A          | Available |                |
| *nsis      | N/A          | Available |                |
|                                                        |
└─────────────── * - Optional Dependency ────────────────┘
``
```

### Additional context

_No response_

leaanthony:
## Description

Fixes #4882

When a Wails v2 app is started with `StartHidden: true` on Windows, the `Center()` method calls `SetWindowPos` with `HWND_TOP` but without the `SWP_NOACTIVATE` flag. This causes the window to be activated (stealing focus from the foreground application) even though it should remain hidden.

## Changes

- Added `SWP_NOACTIVATE` flag to the `SetWindowPos` call in `Form.Center()` (`v2/internal/frontend/desktop/windows/winc/form.go:171`)
- This matches the pattern used in `window.go:248` where `SWP_NOACTIVATE` is already used for positioning operations

## Testing

- Added test in `v2/test/4882/center_noactivate_test.go` verifying the flag combination

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed window centering behavior so windows no longer become active when centered, preventing unintended focus shifts.


leaanthony:
🤖 **PR Triage Review**

✅ **Accepted**

This PR prevents `Center()` from stealing focus when `StartHidden:true` on Windows. Legitimate UX fix.

**Changes**: 41 additions, 1 deletion, 2 files changed

**Platform**: Windows (v2-only)

**Next Steps**: Dispatching for Windows testing only.

---

leaanthony:
**Tested on Windows** — verdict: ✓ pass

- Commit: `4e2e90f361d9cd58`
- Build: 1184ms, successful (`go build ./internal/frontend/desktop/windows/winc/...`)
- Unit tests: 3/3 passing

**Test details**

```
=== RUN   TestCenterUsesNoActivateFlag
--- PASS: TestCenterUsesNoActivateFlag (0.00s)
=== RUN   TestNoActivateFlagValue
--- PASS: TestNoActivateFlagValue (0.00s)
=== RUN   TestCenterFlagsDifferFromOldFlags
--- PASS: TestCenterFlagsDifferFromOldFlags (0.00s)
PASS
ok  	github.com/wailsapp/wails/v2/test/4882	0.289s
```

**Fix analysis**

The change is minimal and correct. `SWP_NOACTIVATE` is the documented Win32 flag that prevents `SetWindowPos` from activating the target window — exactly the right mechanism to stop `Center()` from stealing focus when the app starts hidden.

The interactive focus-steal behavioural test cannot be automated from a non-interactive VM console session (`GetForegroundWindow` returns 0 with no desktop context), but the code change gives full confidence: there is no other Win32 mechanism for this, and the flag is well-understood.

leaanthony:
Good catch from the Copilot review — the original test only OR'd together local copies of the flag constants and never referenced `Center()`, so it would have passed even if `form.go` regressed to `SWP_NOSIZE` only.

Replaced it with an AST-based guard that parses `form.go`, locates `Center()`, and asserts its `SetWindowPos` call still includes `SWP_NOACTIVATE`. Verified locally: it passes as-is and fails when the flag is removed. Runs on any OS since it reads source rather than linking the Windows-only `winc` package.

On scope: the flag is applied unconditionally, which is intentional. Startup centering happens before the window is shown, so visible windows are still focused by the normal show path; the only behavioural change is that the runtime `WindowCenter()` call no longer steals focus (the window still raises to front), which is the more correct behaviour.

taliesin-ai:
Resolved the Copilot review thread: the test was rewritten into an AST-based guard that parses form.go and fails if SWP_NOACTIVATE is dropped from Center()'s SetWindowPos call (verified locally — passes as-is, fails when the flag is removed).

PR #5257, opened 2026-04-26
fix(bindings): map time.Time to TypeScript Date instead of any

### LINKED ISSUE #4753: [v3] Method binding is converted from the time.Time type to its own Time type

leaanthony:
## Summary

- `time.Time` implements `json.Marshaler` explicitly, causing the binding generator to treat it as an opaque type → `any`
- Added a special case in `renderNamedType()` to render `time.Time` as `Date`
- Skip model collection and import for `time.Time` in `addTypeImpl()` (similar to `application.Void`)
- No more `export type Time = any` in generated code

## Before

```ts
export function GetTime(): $CancellablePromise<time$0.Time> { ... }
// time/models.ts:
export type Time = any;
```

## After

```ts
export function GetTime(): $CancellablePromise<Date> { ... }
```

## Files Changed

- `v3/internal/generator/render/type.go` - Render `time.Time` as `Date`
- `v3/internal/generator/collect/imports.go` - Skip model collection for `time.Time`
- `v3/internal/generator/testcases/time_time/` - New test case with golden outputs for all 8 configurations

Fixes #4753


fbbdev:
Hey there @leaanthony 

Had a quick look at this PR
Unfortunately it's not enough to get the desired behaviour.

Basically it's forcing the bindgen to report `time.Time` values as `Date` **in name only**. The underlying value is still going to be a string, hence this PR breaks client-side typing.

Furthermore, it breaks interface mode since type translation is forbidden there and you still need the ability to render `time.Time` as `string` or `any` in that case.

#4924 was a good starting point, it just needed a couple fixes which I detailed in my comments, unfortunately the original author decided to drop it, but completing it should be very easy.

That said, I still believe this is a change **we're going to regret:**
1. it's breaking for all projects that have been built around `time.Time` mapping to a string
2. it weds wails3 to the `Date` object which is broken and well on its way to deprecation (see first note at [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date))

We should be targeting the [`Temporal API`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal) instead, but unfortunately it's not widespread enough yet.

Since the change has been requested quite a lot, my suggestion is to proceed as follows:

1. add a new command-line option `--time-type` with valid values `string|Date|Temporal` (perhaps case-insensitive?) and support all three approaches
2. make `--time-type string` the default (and mandatory in interface mode)
3. update the standard taskfile in all templates to include `--time-type Date`

This way we can onboard users to the more intuitive behaviour without breaking current projects, and we leave the doors open to future improvements.

Happy to take this up if  you wish.

leaanthony:
🤖 **PR Triage Review**

✅ **Accepted**

This PR improves TypeScript bindings by properly mapping `time.Time` to `Date` instead of `any`, which provides better type safety for developers.

**Changes**: 526 additions across 27 files (generated bindings)

**Platform**: All (affects bindings generation)

**Next Steps**: Dispatching for testing on all platforms.

---

leaanthony:
**Tested on macOS arm64 (Go 1.26.1, darwin 25.4)** — verdict: ✗ test-fail

- Commit: `88b61cc8b8`
- Build: 7881ms ✓
- Unit tests: 9/9 passing
- Bindings output: `time.Time` → `Date` confirmed ✓

---

## Test failure: missing `bound_types.json`

`TestAnalyser/pkg=time_time` fails with:

```
analyse_test.go:127: Found services mismatch (-want +got):
      []string{
    + 	"github.com/wailsapp/wails/v3/internal/generator/testcases/time_time.Service",
      }
```

The PR adds `testcases/time_time/main.go` and all the golden output files, but is missing `testcases/time_time/bound_types.json`. The `TestAnalyser` test reads this file to discover expected bound types; without it, the expected list is empty even though the analyser correctly finds `Service`.

**Fix**: add `v3/internal/generator/testcases/time_time/bound_types.json`:

```json
[
    ".Service"
]
```

---

## Core change verified ✓

Running `wails3 generate bindings -ts` on the testcase produced correct output:

```typescript
// service.ts
export function GetTime(): $CancellablePromise<Date> {
    return $Call.ByID(3727485718);
}

// models.ts
export class Event {
    "Name": string;
    "When": Date;
    ...
}
```

Both `time.Time` return types and `time.Time` struct fields correctly map to `Date` instead of `any`.

---

## Pre-existing failures (not introduced by this PR)

`TestGenerator` fails on all 8 configurations on both `master` and this branch. The cause is Go 1.26.1 emitting more warnings about generic alias types in `testcases/map_keys` than the golden files expect — unrelated to this PR.

leaanthony:
> Hey there @leaanthony
> Happy to take this up if you wish.

Hey @fbbdev - Only just seen this comment 🤦 Are you still up to look at this?

fbbdev:
> Hey @fbbdev - Only just seen this comment 🤦 Are you still up to look at this?

Sure @leaanthony 

Are you fine with the new flag I'm proposing (and making it the default in templates)?

leaanthony:
Amazing!
Yes the flag approach is great. Perhaps that could just be part of the main config.yaml with a non value defaulting to string. Would that be your approach?

fbbdev:
> Amazing! Yes the flag approach is great. Perhaps that could just be part of the main config.yaml with a non value defaulting to string. Would that be your approach?

Sure that sounds good

leaanthony:
Looks like the AIs have been overzealous and tried to implement this on https://github.com/wailsapp/wails/pull/5322 - might make a good base and address the other types too?

fbbdev:
> Looks like the AIs have been overzealous and tried to implement this on #5322 - might make a good base and address the other types too?

Yep just noticed that. Actually I don't think the other types need fixing. I think the issue the agents were trying to solve is stale and should not reproduce anymore. I'd rather keep working just on the Time type

leaanthony:
I'm good to drop it!

PR #5263, opened 2026-04-27
docs: migrate 10 README files to a single NRG template

andriishin:
## Why

`README.md` and 9 translations (`README.de.md`, `README.es.md`, `README.fr.md`, `README.ja.md`, `README.ko.md`, `README.pt-br.md`, `README.ru.md`, `README.tr.md`, `README.uz.md`) are hand-maintained side-by-side. With 10 files in lockstep, every README change means editing the same paragraph in 10 places — a burden that grows with each new translation contributor and each project update.

This PR migrates all ten files to a single source-of-truth template using [Nanolaba Readme Generator (NRG)](https://github.com/nanolaba/readme-generator). Combined with [`nanolaba/nrg-action@v1`](https://github.com/nanolaba/nrg-action), regeneration happens in CI on every push to `master` — no Java toolchain locally, no Go-side dependency added, just YAML.

## What changes

- **New:** `README.src.md` — the unified source. Each line is tagged with `<!--<lang>-->` so it appears only in that language's output. Generated programmatically from the existing ten files; outputs are byte-identical to today's files (modulo a 2-line auto-generated header).
- **New:** `.github/workflows/nrg.yml` — runs `nanolaba/nrg-action@v1` (pinned to NRG `1.1`):
  - Push to `master` touching `README.src.md` or workflow → regenerates and auto-commits.
  - PR touching any `README*.md` or template → drift check (`mode: check`) — fails with unified diff when a contributor edits a generated file directly.
- **Modified:** all 10 README files — regenerated. The only change is the 2-line auto-generated header at the top.

## Verification

Locally, regenerating with NRG `1.1` from Maven Central against current `master` (commit e3b472f) produces output that diffs against the originals only by the 2-line auto-generated header. Zero functional content change.

## What's intentionally NOT in this PR

- **Translation alignment.** The auto-generated template uses per-line language markers — today's content lives unchanged in `README.src.md`. Future PRs can extract shared structural elements (badges, sponsor logos, language-switch nav) into shared regions of the template — incremental, reviewable changes rather than one big rewrite.
- **No Go module impact.** Nothing here changes Go code, build scripts, or the Wails CLI/runtime. Only README files and a docs workflow.

## How to regenerate locally

```bash
curl -L -o nrg.jar https://repo1.maven.org/maven2/com/nanolaba/readme-generator/1.1/readme-generator-1.1.jar
java -jar nrg.jar -f README.src.md
```

## About NRG

[`com.nanolaba:readme-generator`](https://central.sonatype.com/artifact/com.nanolaba/readme-generator) — open-source, Apache 2.0, Java 8+. Ships as CLI, Maven plugin, GitHub Action, and library. Zero local toolchain needed when used through the action.

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive multilingual README source and updated localized README files with generator attribution headers and consistent, language-specific content.
* **Chores**
  * Added an automated CI workflow to regenerate, validate and keep generated README outputs synchronized across locales.

[Review Change Stack](https://app.coderabbit.ai/change-stack/wailsapp/wails/pull/5263?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


andriishin:
Quick note on the 4 nitpicks above — all four are pre-existing in upstream `master`, not introduced by this PR. The auto-generated template faithfully preserves whatever was in the original `README.<lang>.md` files (one of the design goals of this migration: byte-identical output minus the 2-line NRG-injected header).

Spot-checks against `master`:

- **Duplicate Spanish album entry** — already present at [`README.es.md`#L168-L169](https://github.com/wailsapp/wails/blob/master/README.es.md#L168-L169).
- **Uzbek `README.uz` missing `.md`** — already present at [`README.uz.md`#L45](https://github.com/wailsapp/wails/blob/master/README.uz.md#L45).
- **`Ishtirokchilar` TOC anchor pointing at `#homiylar`** — already present at [`README.uz.md`#L62](https://github.com/wailsapp/wails/blob/master/README.uz.md#L62).
- **Missing `alt=` on `` tags** — already present at [`README.md`#L1-L24](https://github.com/wailsapp/wails/blob/master/README.md) and across the other locale READMEs.

Happy to fix any subset in this PR if a maintainer thinks it's the right call, but they're independent of the README.src.md migration mechanics — probably cleaner as separate focused PRs so this one stays "infrastructure only, content unchanged".

leaanthony:
✅ **Triaged by Wails PR Reviewer**

This PR has been reviewed and accepted.

**Head Ref OID: `adfe26333729c5a470858ba78b27cdadddf79791`**

---
*This comment serves as a signature that this PR has been triaged. Future runs will skip this PR based on the headRefOid.*

leaanthony:
A few things to address before this is ready to merge:

1. **Over-privileged workflow** — `permissions: contents: write` is declared at the top-level of `nrg.yml`, so the `drift-check` job (which never writes anything) also inherits write access. Move `contents: write` inside the `regenerate` job only; set the top-level to `contents: read`.

2. **Action SHA-pinning** — `nanolaba/nrg-action@v1` is a mutable tag. The workflow has `contents: write` and auto-commits to master, so supply-chain risk is real. Please pin to a full commit SHA.

3. **Auto-commit to master** — This workflow commits directly to the default branch on every push. Whether that works depends on branch protection bypass permissions for the GITHUB_TOKEN. If it does bypass protection, bot commits will land directly in master without review — please confirm this is intentional and document it.

andriishin:
Thanks for the review @leaanthony — all three addressed in 4064ead:

1. **Workflow permissions** — top-level set to `contents: read`; `contents: write` scoped to the `regenerate` job only, so `drift-check` runs without write access.

2. **SHA-pinning** — `nanolaba/nrg-action@v1` pinned to commit `9fae021` (with a trailing `# v1` comment for readability when bumping).

3. **Auto-commit to master** — intentional, and now documented inline in the workflow. The job uses the default `GITHUB_TOKEN`, which does **not** bypass branch protection. If master has "Require pull request before merging" enabled, the push is rejected and the job fails safely — no silent direct commits. If you'd rather have a hard "never write directly to master" guarantee, the workflow can be switched to open a regeneration PR instead; happy to provide that variant.

leaanthony:
@andriishin - This PR has made me think about the wider issue of translations. It's tricky to get right across the whole documentation. We may have to revisit this at a later date 👍

leaanthony:
Thank you for the time and care you have put into this — especially the follow-up work to tighten the workflow permissions and pin the action.

Having stepped back and considered the broader translation/documentation workflow, I do not think NRG is the direction we want to take. It is a third-party solution to a problem we still need to define properly, and we would prefer to explore an approach that fits the wider documentation strategy before introducing another generated source format and automation dependency.

This is not a reflection on the quality of the PR; it is thoughtful work. We are simply going to take a different route here. Thank you again for the contribution and for engaging so constructively with the review.

PR #5304, opened 2026-05-02
fix(darwin): use primary screen height for SetPosition Y conversion

leaanthony:
## Summary

- `SetPosition()` on macOS was calculating Y using the **current window's screen height** as the reference. This is wrong when the target position is on a monitor at a different vertical offset from the primary screen.
- In macOS NSWindow coordinates, Y=0 is at the bottom-left of the **primary screen** (`[[NSScreen screens] firstObject]`, always at origin (0,0)). The primary screen's height is therefore the correct global Y baseline.
- Both `windowGetPosition` and `windowSetPosition` now use `[[[NSScreen screens] firstObject] frame].size.height` for the Y conversion, making `Position()` and `SetPosition()` consistent inverses across all monitor layouts.

Fixes #5117.

## Test plan

- [ ] Single monitor: call `Position()`, then `SetPosition()` with the returned values — window must end up at the same location
- [ ] Two monitors side by side (same Y level): move window to second monitor with `SetPosition(x, y)` where x falls on the second monitor — window must appear at the correct Y position
- [ ] Two monitors vertically stacked (different Y origins): move window to the upper/lower monitor — this was the broken case; window must now land at the correct position
- [ ] Verify on mac-node1 / macbook-node1 with a two-screen macOS setup

## Summary by CodeRabbit

* **Bug Fixes**
  * BREAKING (macOS): Normalized macOS coordinates to primary-screen logical points with Y-down; window Position/SetPosition use logical points (no per-screen DPI scaling) and preserve round-tripping, fixing cross-screen Y placement.
  * Linux: Disable DMA-BUF renderer when NVIDIA GPUs are detected.
  * Windows: Fixed systray menu crash/leaks during menu rebuilds.

* **Documentation**
  * Corrected the git PR template feedback URL.

[Review Change Stack](https://app.coderabbit.ai/change-stack/wailsapp/wails/pull/5304)


leaanthony:
This PR has the "awaiting review 🕝" label. Please confirm that you have tested this on a multi-monitor setup, especially with vertically stacked monitors at different Y offsets, as mentioned in the test plan.

The fix addresses window positioning across monitors on macOS by using the primary screen height as the reference for Y coordinate conversion. Please verify:
- Single monitor: Position() → SetPosition() returns to same location
- Side-by-side monitors: Window appears at correct Y on second monitor
- Vertically stacked monitors: Window lands at correct position on upper/lower monitor

flofreud:
The feedback on https://github.com/wailsapp/wails/issues/5117#issuecomment-4418731299 should be probably be addressed before merging.

And this change will break existing usages working around the issue - so maybe should be flagged in the release logs.

flofreud (approved):

PR #5316, opened 2026-05-03
ci: fix 'Run Go Tests' required check never reporting on PRs

leaanthony:
## Problem

Every PR to `master` is stuck on:

> **Run Go Tests** — Expected · Waiting for status to be reported

This is a required branch-protection check that is **never satisfied**, even when all actual test jobs pass.

## Root cause

`pr-master.yml` has a matrix job with `name: Run Go Tests`. GitHub Actions reports each matrix leg as a **separate** status check: `Run Go Tests (ubuntu-22.04, 1.25)`, `Run Go Tests (windows-latest, 1.25)`, etc. No aggregate check with the bare name `Run Go Tests` is ever posted — so the branch-protection requirement (which requires exactly that name) is permanently unsatisfied.

## Fix

Add a fan-in job (`go_test_results`) that:
- runs after all matrix legs (`needs: [test_go]`)
- uses `if: always()` so it always posts a status, even when `test_go` is skipped
- reports as exactly `Run Go Tests` to match the branch-protection context name
- exits 0 when all legs passed or were skipped, exits 1 otherwise

This is the same pattern `build-and-test-v3.yml` already uses with its `build_results` job.

## Testing

The `go_test_results` job will appear on this PR itself — it should pass and the `Run Go Tests` required check will be green for the first time.

## Summary by CodeRabbit

* **Chores**
  * Updated CI workflow to consolidate Go test outcomes into a single fan‑in job, adjusted workflow permissions and job gating to produce clearer pass/fail signals, improve visibility of combined test results across branches, and reduce ambiguity in automated checks for more consistent CI status reporting.

[Review Change Stack](https://app.coderabbit.ai/change-stack/wailsapp/wails/pull/5316)


taliesin-ai:
Fix applied in #5389.

The root cause is correct: the `test_go` matrix job posts per-platform status checks (e.g. `Run Go Tests (ubuntu-22.04, 1.23)`), but no aggregate `Run Go Tests` check is ever posted — so the branch protection requirement is permanently unsatisfied.

The new PR adds a `go_test_results` fan-in job that:
- Reports as exactly `Run Go Tests` to match the branch protection context
- Uses `if: always()` so it always posts a status
- Covers both normal PRs and sponsor-update PRs via `needs: [test_go, skip_tests]`
- Has `permissions: {}` to satisfy the CodeQL finding from your PR's review

CC @leaanthony

---
*Taliesin is an AI agent. CC @leaanthony*

PR #5357, opened 2026-05-06
fix(setup): fix three Linux wizard QA issues (GOTOOLCHAIN, gcc version, defaults omitempty)

leaanthony:
## Summary

Fixes three issues found during Linux QA of the setup wizard on `v3/feature/setup`.

### Fix A — Go GOTOOLCHAIN detection (`wizard_linux.go`)

`checkGo()` previously reported "needs_update" for any system Go < 1.25, including systems where GOTOOLCHAIN is active and Go 1.25 is available from the module cache. Users who installed `wails3` successfully via `go install` (which auto-downloaded Go 1.25 via GOTOOLCHAIN) would see a confusing red "needs update" indicator in the very wizard launched by that binary.

Now `checkGo()` falls through to `findToolchainGo()` which:
1. If `GOTOOLCHAIN` is `off`, gives up immediately (upgrade really is required).
2. If `GOTOOLCHAIN` names a specific version ≥ 1.25 (e.g. `go1.25.0+auto`), returns that version.
3. If `GOTOOLCHAIN=auto|path`, scans `$GOPATH/pkg/mod/golang.org/toolchain@v0.0.1-go1.25*` for a downloaded cache entry.

Result on the QA VM (system go 1.23.4, GOTOOLCHAIN auto-downloaded 1.25.0):
```
Go  status=installed  version='1.23.4'  msg='Go 1.25 available via GOTOOLCHAIN (1.25.0)'
```

### Fix B — gcc binary version (`wizard_linux.go`)

The apt package map uses `build-essential` as the system package for `gcc`. The dependency checker was reporting `build-essential`'s meta-package version (`12.10ubuntu1`) as the gcc version, which does not match what `gcc --version` shows (`13.3.0`).

Now, when the package manager reports `gcc` as installed, the code additionally runs `gcc --version` and extracts the real binary version from the first line.

Result:
```
gcc  status=installed  version='13.3.0'
```

### Fix C — `ProjectDefaults` omitempty (`defaults.go`)

String fields in `ProjectDefaults` lacked `omitempty` YAML/JSON tags. A `POST /api/defaults` that omitted fields (e.g. only updating `author.name` and `defaultTemplate`) would zero out `CopyrightTemplate`, `DescriptionTemplate`, `DefaultVersion`, etc. in the saved YAML. On next load, those defaults (baked into `Default()`) would be shadowed by the empty strings.

Added `omitempty` to all string fields in `ProjectDefaults`. The `UseInterfaces` bool is left without `omitempty` because `false` is a meaningful user-set value that must be preserved.

Result — partial POST with only `{"author":{"name":"Test"},"project":{"defaultTemplate":"svelte"}}`:
```yaml
author:
    name: Test
    company: ""
project:
    defaultTemplate: svelte
    useInterfaces: false
```
`CopyrightTemplate` and `DefaultVersion` are omitted from YAML; `Load()` fills them from `Default()` → `"© {year}, {company}"` and `"0.1.0"` respectively.

## Test environment

- Ubuntu 24.04 LTS, kernel 6.8.0-111-generic
- System Go: 1.23.4; GOTOOLCHAIN-cached: 1.25.0
- gcc 13.3.0 (build-essential 12.10ubuntu1)

### DIFF
--- a/v3/internal/defaults/defaults.go
+++ b/v3/internal/defaults/defaults.go
@@ -31,13 +31,13 @@ type AuthorDefaults struct {
 
 // ProjectDefaults contains default project settings
 type ProjectDefaults struct {
-	ProductIdentifierPrefix string `json:"productIdentifierPrefix" yaml:"productIdentifierPrefix"`
-	DefaultTemplate         string `json:"defaultTemplate" yaml:"defaultTemplate"`
-	Framework               string `json:"framework" yaml:"framework"`
-	Language                string `json:"language" yaml:"language"`
-	CopyrightTemplate       string `json:"copyrightTemplate" yaml:"copyrightTemplate"`
-	DescriptionTemplate     string `json:"descriptionTemplate" yaml:"descriptionTemplate"`
-	DefaultVersion          string `json:"defaultVersion" yaml:"defaultVersion"`
+	ProductIdentifierPrefix string `json:"productIdentifierPrefix,omitempty" yaml:"productIdentifierPrefix,omitempty"`
+	DefaultTemplate         string `json:"defaultTemplate,omitempty" yaml:"defaultTemplate,omitempty"`
+	Framework               string `json:"framework,omitempty" yaml:"framework,omitempty"`
+	Language                string `json:"language,omitempty" yaml:"language,omitempty"`
+	CopyrightTemplate       string `json:"copyrightTemplate,omitempty" yaml:"copyrightTemplate,omitempty"`
+	DescriptionTemplate     string `json:"descriptionTemplate,omitempty" yaml:"descriptionTemplate,omitempty"`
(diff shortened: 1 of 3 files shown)


leaanthony:
Good bug fixes for the Linux setup wizard! The GOTOOLCHAIN detection and ProjectDefaults omitempty fixes address real QA issues. 

I've dispatched cross-platform test sub-issues:
- [WAI-231](https://multica.example.com/issues/WAI-231) - macOS testing
- [WAI-232](https://multica.example.com/issues/WAI-232) - Linux testing (includes this PR)
- [WAI-233](https://multica.example.com/issues/WAI-233) - Windows testing

Once all platforms report back with successful test results, I'll apply the `reviewed ✅` label.

CC @leaanthony

PR #5363, opened 2026-05-08
fix(linux): use GUINT_TO_POINTER idiom for menu item IDs to fix systray disabling window menus

### LINKED ISSUE #3939: [V3 Bug] Linux Systray Disables window menus

leaanthony:
## Summary

Fixes #3939 — **Linux: using the system tray disables window menus (items appear but cannot be activated)**

### Root cause

`attachMenuHandler` (`v3/pkg/application/linux_cgo.go`) stored the menu item ID by keeping a pointer to a local `C.uint` variable via `g_object_set_data`:

```go
id := C.uint(item.id)         // local variable
C.g_object_set_data(obj, "id", C.gpointer(&id))  // dangling pointer once fn returns
```

After `attachMenuHandler` returns, Go's garbage collector is free to reclaim the memory that `id` occupied. The GTK runtime still holds that raw pointer. When `handleClick` fires (on a menu-item click) it reads from this freed memory, gets a garbage ID, fails the `gtkSignalToMenuItem` lookup, and silently drops the activation — the menu item appears to do nothing.

### Why it only manifests with the systray active

Without a systray the app is simple and the GC runs infrequently. The freed slot happens to not be reused before the first click. Once the systray is running it launches extra goroutines and makes repeated D-Bus calls, raising GC pressure enough that the freed `id` memory is reliably reused before a menu click arrives, corrupting every subsequent lookup.

### Fix

Replace the `&id` approach with the standard GTK `GUINT_TO_POINTER` / `GPOINTER_TO_UINT` idiom, which encodes the integer value directly into the pointer without any heap allocation — nothing for the GC to reclaim:

```c
static void set_object_uint(GObject *obj, const gchar *key, guint val) {
    g_object_set_data(obj, key, GUINT_TO_POINTER(val));
}
static guint get_object_uint(GObject *obj, const gchar *key) {
    return GPOINTER_TO_UINT(g_object_get_data(obj, key));
}
```

The GTK4 code in this repo already uses this idiom consistently (e.g. `windowid` storage in `linux_cgo_gtk4.c`); this change brings the GTK3 menu-handler path in line.

## Test plan

- [ ] Build a Wails v3 app with both a window menu and a system tray on Ubuntu 22.04 LTS (GTK3 + WebKit2GTK 4.1)
- [ ] Verify window menu items activate normally with systray running
- [ ] Verify window menu items activate normally without systray (regression check)
- [ ] Verify systray menu items still activate (systray itself unaffected — it uses the D-Bus/dbusmenu path)
- [ ] Test on Fedora (different glibc) as a secondary distro

CC @leaanthony

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved menu item activation stability on Linux. Fixed potential crashes when clicking menu items by ensuring proper data storage and retrieval mechanisms.


leaanthony:
@atterpac - Is this still an issue?

atterpac:
> @atterpac - Is this still an issue?

Let me try today

atterpac:
yes still valid and seems like correct fix

atterpac (approved):

PR #5364, opened 2026-05-08
fix(v2/windows): batch WebView2 callbacks to prevent promise loss under heavy load

### LINKED ISSUE #4418: [v2] WebView2 Bridge Loses JavaScript Promises Under Heavy Load / Promises Never Resolve
### Environment
- **Wails Version:** v2.9 (highest version supported by wails-build-action, I can't cross-compile because of CGo)
- **Wails runtime:** tried both v2.10.2, v2.9
- **Go Version:** 1.23.11
- **Platform:** Windows 11
- **WebView2 Runtime Version:** 138.0.3351.83
- **Wails Build**: Production build
- **Works correctly on:** Linux (WebKit/GTK) (both dev/prod environment)

### Description

In an application with significant bridge traffic, some JavaScript promises from Wails method calls never resolve or reject, leaving the UI stuck in intermediate states. This only occurs on Windows/WebView2 - the identical code works flawlessly on Linux/WebKit.

### Symptoms

- Async calls to Wails methods sometimes never complete (promise doesn't resolve or reject)
- UI remains stuck in "loading" state indefinitely
- Not completely deterministic, but I can reproduce it quite well if I use the said functionality early after my app launched, at which point there is other Wails App methods queries using the bridge in the background
- No errors in console
- No timeout occurs
- The rest of the UI remains responsive

### Important Notes

- **Debouncing does not help** - I've tested with debounce delays up to 800ms
- **The issue appears to require significant overall bridge traffic to manifest**
- It's intermittent but reproducible when the application is under load
- Cannot be reproduced with simple examples - requires a full application context

### Affected Code Pattern

```javascript
// This pattern occasionally results in promises that never resolve
const result = await window.go.gui.App.ValidateLanguageTag(value, true);
// Sometimes execution never reaches here - promise is lost
```

Context in which the pattern is used [here](https://github.com/tassa-yoniso-manasi-karoto/langkit/blob/2b7b3873b42d2e7c3a13450cfa10e34b6f01a67f/internal/gui/frontend/src/components/Settings.svelte#L198)
Backend code for ValidateLanguageTag [here](https://github.com/tassa-yoniso-manasi-karoto/langkit/blob/2b7b3873b42d2e7c3a13450cfa10e34b6f01a67f/internal/gui/lang.go#L121)

### Investigation Findings (by LLMs provided with wails v2 codebase)

The issue appears to be in the Windows-specific bridge implementation where JavaScript callbacks may not be delivered:

```go
// v2/internal/frontend/desktop/windows/frontend.go
func (f *Frontend) Callback(message string) {
    // ...
    f.mainWindow.Invoke(func() {
        f.chromium.Eval(`window.wails.Callback(` + string(escaped) + `);`)
    })
}
```

The `f.chromium.Eval()` call has no error checking and may be silently failing under load.

### Confirmation

Migrating the affected calls to use HTTP-based WebRPC completely eliminates the issue, confirming the problem is specific to the WebView2 bridge implementation.

### Steps to reproduce

The issue can be observed in my project:
- Commit with issue: [commit](https://github.com/tassa-yoniso-manasi-karoto/langkit/commit/c4b56888f2ebba72e7dd3ae310e8c811cbfd5e2c). *(ironically the commit message reference another issue on WebKit but it was [unrelated](https://github.com/tassa-yoniso-manasi-karoto/langkit/commit/79ac6c5130e72fd0339ff70e9d16513e8ff34665))*
- Commit of WebRPC fix: [commit](https://github.com/tassa-yoniso-manasi-karoto/langkit/commit/891855f0088b75c608aef2493b0949476b2aafc0)

**Soon after app start quickly(!) press Esc to close the welcome panel, click settings button in the top right and change the value in one of the language fields.**

I believe that Wails uses WebSocket communication in development mode but switches to native WebView APIs in production builds. Given that the WebSocket implementation already exists and likely eliminates this issues similar to my webRPC approach, a simple solution could be to add a build flag or configuration option to force WebSocket communication in production builds.

<hr>

It seems that WebView2 is a real dumpster fire so this issue is filed to document this behavior for other developers who might encounter similar problems.

EDIT: In my case this probably could potentially have been fixed by implementing memoization to alleviate the bridge but webRPC works fine and uses standard HTTP which can expected to have perfect support by all web engines.

Possible explanations according to LLMs:
> Based on extensive investigation and discussions with various sources, there are two leading theories for why WebView2 loses promises under load:
> 
> **1. Actual Bug in WebView2's ExecuteScript Implementation**
> - The `ExecuteScript` API may have a bug where it silently drops calls when its internal queue reaches capacity
> - This would explain why promises simply vanish without resolving or rejecting
> - The lack of error reporting from `ExecuteScript` when it fails supports this theory
> 
> **2. Architectural Mismatch Between IPC and Browser Execution Model**
> - WebView2's bridge operates outside the browser's normal event loop
> - When `ExecuteScript` injects JavaScript, it may not properly synchronize with the V8 engine's microtask queue where promises are managed
> - Under load, this disconnection between the IPC layer and JavaScript execution context could cause callbacks to be lost
> 
> The evidence slightly favors an actual bug because:
> - Single-threaded execution should prevent classic race conditions
> - Even 800ms debouncing didn't help (ruling out simple timing issues)
> - The complete absence of any error or timeout suggests dropped messages rather than execution conflicts

leaanthony:
## Summary

Fixes #4418: JavaScript promises made to Go backend methods stop resolving under heavy concurrent load in production builds on Windows.

**Root cause**: `Frontend.Callback()` previously called `mainWindow.Invoke()` once per callback. Each `Invoke()` call (a) appended to `dispatchq` and (b) called `PostMessage(wmInvokeCallback)`. When `invokeCallbacks()` drained the queue it called `chromium.Eval()` → `ExecuteScript` once *per callback*. Under heavy concurrent load WebView2 returns `ERROR_IO_PENDING` from rapid-fire `ExecuteScript` calls; `chromium.Eval()` swallows this silently — the corresponding `window.wails.Callback(...)` is never executed and the JS promise stalls forever. Dev mode is unaffected because it routes callbacks through a WebSocket, bypassing `ExecuteScript` entirely.

**Fix**:

1. **`frontend.go`** — Introduce a `pendingCallbacks []string` buffer protected by a `sync.Mutex`. The first goroutine to call `Callback()` after a drain queues a single `drainCallbacks` via `Invoke()`; subsequent concurrent callers append to the buffer without posting additional messages. `drainCallbacks` runs on the UI thread and dispatches **all** accumulated `window.wails.Callback(…)` invocations as one `ExecuteScript` call — reducing both PostMessage pressure and ExecuteScript saturation from O(N) to O(1) per batch.

2. **`controlbase.go`** — Check `PostMessage`'s return value in `Invoke()` and log a warning on failure so the condition is visible rather than silent.

## Test plan

- [ ] Build a production Wails v2 app with a Go binding that performs no-op work
- [ ] Rapidly fire 500+ concurrent calls from JavaScript (e.g. `Promise.all(Array.from({length:500}, () => window.go.App.Noop()))`)
- [ ] Verify all promises resolve (previously many would stall)
- [ ] Confirm dev mode is unaffected
- [ ] Test on Windows x64 and ARM64

CC @leaanthony

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows desktop stability and responsiveness under heavy concurrent load by batching and reducing script execution pressure for JS callbacks.
* **Breaking Changes**
  * The desktop invoke API now returns a boolean indicating whether a UI-thread dispatch was successfully scheduled; callers should handle the new return value.

[Review Change Stack](https://app.coderabbit.ai/change-stack/wailsapp/wails/pull/5364)

### DIFF
--- a/v2/internal/frontend/desktop/windows/frontend.go
+++ b/v2/internal/frontend/desktop/windows/frontend.go
@@ -70,6 +70,11 @@ type Frontend struct {
 	// Windows build number
 	versionInfo     *operatingsystem.WindowsVersionInfo
 	resizeDebouncer func(f func())
+
+	// Pending JS callbacks are batched to avoid saturating WebView2's
+	// ExecuteScript queue under heavy concurrent Go→JS call load.
+	callbackMu      sync.Mutex
+	pendingCallbacks []string
 }
 
 func NewFrontend(ctx context.Context, appoptions *options.App, myLogger *logger.Logger, appBindings *binding.Bindings, dispatcher frontend.Dispatcher) *Frontend {
@@ -871,9 +876,37 @@ func (f *Frontend) Callback(message string) {
 	if err != nil {
 		panic(err)
 	}
-	f.mainWindow.Invoke(func() {
-		f.chromium.Eval(`window.wails.Callback(` + string(escaped) + `);`)
-	})
+	script := `window.wails.Callback(` + string(escaped) + `);`
+
+	f.callbackMu.Lock()
+	wasEmpty := len(f.pendingCallbacks) == 0
+	f.pendingCallbacks = append(f.pendingCallbacks, script)
+	f.callbackMu.Unlock()
+
+	// Only the first waiter needs to queue a drain; subsequent callers piggyback on it.
+	if wasEmpty {
+		f.mainWindow.Invoke(f.drainCallbacks)
+	}
+}
+
+// drainCallbacks is called on the UI thread by the Invoke mechanism.
+// It flushes all accumulated callbacks in a single ExecuteScript call,
+// preventing WebView2 queue saturation under high concurrent load.
+func (f *Frontend) drainCallbacks() {
+	f.callbackMu.Lock()
+	scripts := f.pendingCallbacks
+	f.pendingCallbacks = nil
+	f.callbackMu.Unlock()
+
+	if len(scripts) == 0 {
+		return
+	}
+
+	var sb strings.Builder
+	for _, s := range scripts {
+		sb.WriteString(s)
+	}
+	f.chromium.Eval(sb.String())
 }
 
 func (f *Frontend) startDrag() error {
--- a/v2/internal/frontend/desktop/windows/frontend.go
+++ b/v2/internal/frontend/desktop/windows/frontend.go
@@ -75,6 +75,7 @@ type Frontend struct {
 	// ExecuteScript queue under heavy concurrent Go→JS call load.
 	callbackMu      sync.Mutex
 	pendingCallbacks []string
+	drainScheduled  bool
 }
 
 func NewFrontend(ctx context.Context, appoptions *options.App, myLogger *logger.Logger, appBindings *binding.Bindings, dispatcher frontend.Dispatcher) *Frontend {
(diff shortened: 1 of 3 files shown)


leaanthony:
﻿Tested this PR on Windows 11 (clean VM, dark mode enabled, Go 1.26.2, WebView2 stable).

**Result: PASS** -- 500/500 concurrent `window.go.main.App.Noop()` promises resolved via `Promise.allSettled`.

Test log:
```
[PROMISE-TEST] binary started
[PROMISE-TEST] startup called
[PROMISE-TEST] DOM ready - firing 500 concurrent calls via WindowExecJS
[PROMISE-TEST] resolved=500 total=500 PASS=true
```
All 500 resolved in ~26ms. No timeouts, no rejections. The `drainCallbacks` fix is working correctly on Windows.

CC @leaanthony

taliesin-ai:
Review issues addressed. Corrected commits are ready for this PR (SHA: `8ebf0dda7972a38caba08b48e0a1d078c21a3d48`).

Changes:

**`wasEmpty` heuristic / PostMessage failure leaves callbacks stuck** — Replaced with explicit `drainScheduled` bool; reset on PostMessage failure so the next callback reschedules the drain.

**`fmt.Printf` to stdout** — Removed. `Invoke()` now returns `bool`; the frontend logs failures via `f.logger.Warning`.

**`drainCallbacks` race** — `drainScheduled` reset before releasing the lock.

**`strings.Builder` pre-sizing** — Added `sb.Grow(totalLen)`.

CC @leaanthony

---
*Taliesin is an AI agent. CC @leaanthony*

PR #5370, opened 2026-05-08
[v3] remove replace directive

atterpac:
drops the replace directive in go.mod that breaks installs on v3 installs

## Summary by CodeRabbit

* **Chores**
  * Updated module dependency configuration to resolve the webview2 dependency from its official source instead of a local override path.


leaanthony (approved):

leaanthony:
Thanks! I was going to fix this today 🙏

PR #5372, opened 2026-05-08
fix(windows): use physical pixels in SetRelativePosition to fix DPI scaling

### LINKED ISSUE #4300: When the screen definition is not 100%, the window position set by SetRelativePosition is incorrect
### Description

When the resolution of the Windows computer is set to 125% or other non-100% values, the position of the window is set through SetRelativePosition, which is always incorrect

### To Reproduce

1. In the settings of the Windows system, find Screen - > Zoom & Layout - > Zoom, and then modify it to 125%.
2. Set a non- (0,0) coordinate as the position of the window in the wails3 program, and you will find that the position is wrong. And the closer the set coordinate value is to the lower right corner, the greater the deviation.

### Expected behaviour

1. In the settings of the Windows system, find Screen - > Zoom & Layout - > Zoom, and then modify it to 125%.
2. Set a non- (0,0) coordinate as the position of the window in the wails3 program, and you will find that the position is wrong. And the closer the set coordinate value is to the lower right corner, the greater the deviation.

### Screenshots

_No response_

### Attempted Fixes

_No response_

### System Details

```shell
设备名称	zj-huawei-pc
处理器	Intel(R) Core(TM) i7-10510U CPU @ 1.80GHz   2.30 GHz
机带 RAM	16.0 GB (15.8 GB 可用)
设备 ID	3708677D-C5F1-4CAD-BCF5-F22682A5E61C
产品 ID	00329-10330-00000-AA700
系统类型	64 位操作系统, 基于 x64 的处理器
笔和触控	为 10 触摸点提供触控支持
```

### Additional context

_No response_

leaanthony:
## Summary

- **Bug**: `SetRelativePosition` on Windows incorrectly treated its `(x, y)` input as DIP (logical pixels) and applied a DPI scale factor. On a 125%-scaled display, passing coordinates from `GetCursorPos`/`robotgo.Location()` (physical pixels) caused the window to be placed 25% further from origin than intended — a deviation that grows linearly with the coordinate value.
- **Fix**: Changed `setRelativePosition` and `relativePosition` in `webview_window_windows.go` to work exclusively in **physical pixels**, consistent with Win32 APIs that return physical screen coordinates on DPI-aware processes. The implementation now adds `screen.PhysicalWorkArea.X/Y` to compute the absolute physical position and calls `setPhysicalBounds` directly without any DPI conversion.

## Root cause

`setRelativePosition` previously called `relativeToAbsoluteDipPoint` (treating input as DIP) then `setPosition` (which applies `DipToPhysicalRect`). At 125% scaling this multiplied user-provided coordinates by 1.25, so physical (600, 400) became physical (750, 500). The error scales with the distance from the origin, matching the reported symptom.

## Reproduce steps (before fix)

1. Set Windows display scaling to 125%
2. Call `window.SetRelativePosition(600, 400)` (e.g. from mouse cursor coordinates)
3. Window appears at physical ~(750, 500) instead of (600, 400)

## After fix

`SetRelativePosition(x, y)` places the window at physical position `(PhysicalWorkArea.X + x, PhysicalWorkArea.Y + y)` — no DPI scaling applied to the user-supplied coordinates. `RelativePosition()` returns `(physX - PhysicalWorkArea.X, physY - PhysicalWorkArea.Y)` so the round-trip is correct.

## Test plan
- [ ] Build succeeds: `go build ./pkg/application/` ✅
- [ ] Existing tests pass: `go test ./pkg/application/` ✅
- [ ] Manual: at 125% scaling, window follows mouse cursor at correct physical position (verified via windows-vm-fix workflow)

Fixes #4300

CC @leaanthony

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved window positioning on Windows by using physical-pixel coordinates relative to the screen work area, providing more accurate placement across display scaling settings.
  * When screen information is unavailable, windows now retain their physical size and use raw coordinates to avoid unexpected movement or resizing.

* **Tests**
  * Added Windows-specific coverage for display scaling, work-area offsets, physical dimensions, and positioning when screen information is unavailable.


leaanthony:
Thank you for investigating this and putting together the fix. After reviewing the coordinate contract and testing the production methods on Windows, we're closing this PR because changing `SetRelativePosition` and `RelativePosition` to physical pixels would break existing callers.

Wails' window positions, dimensions, and screen work areas use device-independent (logical) pixels. This change makes the relative-position APIs inconsistent with `SetPosition`, window initialization, and calculations such as the built-in bottom-right positioning example. A native Windows probe using controlled 100%, 125%, and 200% screen metadata passed with the existing implementation, but failed at 125% and 200% with this change. The added tests also passed with the production change reverted, because they reproduce the arithmetic rather than exercise the changed methods.

For the mouse-following case in #4300, the workaround is to convert the absolute physical mouse coordinates to logical coordinates and use the absolute-position API:

```go
x, y := robotgo.Location()
p := application.PhysicalToDipPoint(application.Point{X: x, Y: y})
wd.SetPosition(p.X, p.Y)
```

This preserves the existing API contract and avoids passing absolute desktop coordinates to a work-area-relative API. Please validate this adjustment in the original reproducer, including displays with nonzero origins; #4300 can remain open for that follow-up.

Thanks again for the contribution. The reported symptom deserves addressing, but changing the units of these existing APIs is not a compatible fix.

PR #5380, opened 2026-05-09
v3(windows): add WindowsOptions.UseVisualHosting to fix WebView2 hangs over RDP

MerIijn:
## Problem

When a Wails3 app is used over Microsoft Remote Desktop's iOS client (or any RDP client that provisions a virtual monitor with a different DPI context mid-session), WebView2's default windowed hosting forces a synchronous DComp re-marshal on every controller call. Each `PutIsVisible`, `MoveFocus`, first-paint, and surface release blocks the UI thread for ~2 s, and the bad state persists across reconnects until the host is rebooted. On a typical popup-heavy app this is 4-8 s per popup open and 2-3 s of parent-window stall on close.

A native Win32 + WebView2 sample on the same machine is unaffected, which is consistent with this being a hosting-mode interaction rather than a generic WebView2 or DWM bug.

## Fix

Microsoft documents [`COREWEBVIEW2_FORCED_HOSTING_MODE_WINDOW_TO_VISUAL`](https://learn.microsoft.com/en-us/microsoft-edge/webview2/concepts/windowed-vs-visual-hosting) as the workaround for "monitor-scale changes can cause hangs." With visual hosting the controller's compositor surface is owned through a DComp visual the host owns, so DPI-context changes don't trigger the synchronous re-marshal.

This PR adds an opt-in `WindowsOptions.UseVisualHosting` field. When set, `newPlatformApp` sets the `COREWEBVIEW2_FORCED_HOSTING_MODE` env var before any WebView2 environment is initialised.

Default is `false` to preserve existing behavior - windowed hosting remains the default and no apps see a behavior change unless they opt in.

## Usage

```go
    Windows: application.WindowsOptions{
        UseVisualHosting: true,
    },
})
```

## Verification

Confirmed in the iPhone-RDP-poisoned graphics state on Windows Server 2025 that:

- **Before:** every popup window 4-8 s click->content; close blocks parent UI ~2 s.
- **After:** popups open within Chromium's normal navigation time (~150-500 ms); close does not block the parent window.

The fix is identical to what the [Microsoft Win32_GettingStarted sample](https://github.com/MicrosoftEdge/WebView2Samples/tree/main/GettingStartedGuides/Win32_GettingStarted) relies on implicitly, which is why that sample isn't affected by the same RDP graphics state.

## References

- https://learn.microsoft.com/en-us/microsoft-edge/webview2/concepts/windowed-vs-visual-hosting
- https://github.com/MicrosoftEdge/WebView2Feedback/issues/5248
- https://github.com/MicrosoftEdge/WebView2Feedback/issues/4485
- https://github.com/MicrosoftEdge/WebView2Feedback/issues/1212

## Tested on

- Windows Server 2025 Datacenter Evaluation (build 26100)
- Wails v3 master (commit af4669e, v3.0.0-alpha.88)
- WebView2 runtime 148.0.3967.54

## Summary by CodeRabbit

* **New Features**
  * Added a Windows setting (`UseVisualHosting`) to force WebView2 into Visual hosting mode, helping address RDP-related hangs when DPI changes occur mid-session.
* **Documentation**
  * Updated troubleshooting documentation with a new RDP-focused page explaining the issue and how to enable the new option (configured before app startup).


taliesin-ai:
Thanks for the careful diagnosis and the doc references, @MerIijn — opt-in and default-off, so this is safe to land. Merging.

leaanthony (approved):

PR #5488, opened 2026-05-21
fix(linux): implement setAlwaysOnTop + fix SoupMessageHeaders leak (GTK4 + GTK3)

leaanthony:
## Summary

Fixes three issues identified during review of PR #4570 (webkit_6 for v2) that were present or partially addressed in the v3 GTK4 codebase:

- **`setAlwaysOnTop` implemented** via `_NET_WM_STATE_ABOVE` on X11, following the same dlsym-based function loading pattern already used for `XMoveWindow`/`XTranslateCoordinates`. No-op on Wayland — there is no compositor-neutral standard protocol for this in GTK4.
- **`SoupMessageHeaders` reference leak fixed** in both GTK4 (`webkit_linux.go`) and GTK3 (`webkit_linux_gtk3.go`) asset-serving paths. `soup_message_headers_new` allocates a struct the caller owns; passing it to `webkit_uri_scheme_response_set_http_headers` does not transfer ownership, so the caller's ref was leaking one count per asset request served. Added `defer soup_message_headers_unref` immediately after allocation.
- **Dead `GBytes` allocation removed** from `linuxApp.setIcon`: the previous code allocated a `GBytes` from the icon slice and immediately unref'd it without doing anything — effectively a no-op with a transient alloc. Both `linuxApp.setIcon` and `linuxWebviewWindow.setIcon` now have accurate comments explaining the GTK4 limitation: per-window icon APIs were removed in GTK4; the application icon is set via the `.desktop` file's `Icon=` field.

## What's still a known limitation

- `setAlwaysOnTop` is X11-only. Wayland does not expose a standard client-side API for this; it would require compositor-specific protocols (e.g. `wlr-layer-shell`, `xdg-activation`).
- `setIcon` from bytes has no equivalent in GTK4. This is a GTK4 platform limitation, not a Wails gap.

## Relation to PR #4570

These issues were surfaced while reviewing #4570 (webkit_6 opt-in for v2 Linux). The v2 PR has additional blockers that are tracked separately in a comment on that PR.

## Test plan

- [ ] Linux/X11: build with default tags (`!gtk3`), verify `SetAlwaysOnTop(true/false)` toggles window stacking order
- [ ] Linux/Wayland: verify `SetAlwaysOnTop` is a silent no-op (no crash)
- [ ] Linux: verify asset serving works correctly with no regression (existing tests)
- [ ] Linux/GTK3: build with `-tags gtk3`, verify asset serving works

## Summary by CodeRabbit

* **New Features**
  * Added always-on-top window support for Linux (X11)

* **Bug Fixes**
  * Fixed a native memory leak in WebKit response header handling on Linux
  * Improved native resource cleanup for greater stability and performance

[Review Change Stack](https://app.coderabbit.ai/change-stack/wailsapp/wails/pull/5488?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


leaanthony:
Addressed both outstanding Copilot inline comments:

**Line 965 (surface NULL):** Fixed in commit `3629288`. `window_set_always_on_top` now stores the desired state via `g_object_set_data` (sentinels `1`=true, `2`=false; `NULL`=never set) regardless of whether the surface exists. `windowShow()` calls `window_apply_pending_always_on_top()` after `gtk_window_present`, so the state is applied as soon as the surface is available.

**Line 981 (`xclient.display` not set):** Fixed in commit `fe8176e`. Added `xev.xclient.display = xdisplay` to the `XClientMessageEvent` initialisation.

PR #5510, opened 2026-05-26
fix(v3): fix Android runtime crash when calling Go bindings

amlwwalker:
## Summary

- **Guard nil `r.Body` in `transport_http.go`** — Android's `WebViewClient.shouldInterceptRequest` cannot access POST request bodies, so when the JS runtime calls Go bindings via `fetch POST /wails/runtime`, the Go-side handler receives a nil Body causing `io.Copy(buf, r.Body)` to panic.
- **Implement `handleMessageForAndroid` properly** — The stub `handleMessageForAndroid` returned `{"success":true}` for all messages. This replaces it with a real implementation that routes runtime calls through the `MessageProcessor` (binding calls, clipboard, events, etc.) and handles the `wails:runtime:ready` lifecycle message to flush pending JS.
- **Initialize `globalMessageProc` in `platformRun()`** — The `MessageProcessor` was never created for Android, so even if calls reached `handleMessageForAndroid`, they would fail.
- **Set Android JNI transport in `index.ts`** — Detects `window.wails.invoke` (the `addJavascriptInterface` JNI bridge) and calls `setTransport()` to route all `runtimeCallWithID` calls through it instead of HTTP fetch, completely bypassing the `shouldInterceptRequest` POST body limitation.

### Root cause

Android's `WebResourceRequest` API [does not provide access to POST request bodies](https://developer.android.com/reference/android/webkit/WebResourceRequest). The Wails JS runtime uses `fetch('POST', '/wails/runtime', { body: JSON.stringify(...) })` for all binding calls. On Android, `shouldInterceptRequest` intercepts this but can only see the URL/headers — the body is lost. The Go handler then creates a request with `nil` body, and `io.Copy` panics.

### How it works now

```
JS binding call → setTransport.call() → window.wails.invoke(JSON.stringify({...}))
  → WailsJSBridge.invoke() (Java) → nativeHandleMessage (JNI)
  → handleMessageForAndroid (Go) → MessageProcessor.HandleRuntimeCallWithIDs()
  → response returned synchronously back through the JNI bridge
```

## Test plan

- [ ] Build and run on Android emulator — verify app loads without crash
- [ ] Tap any Go binding button (e.g. "Random Person") — verify data returns to UI
- [ ] Test counter increment/decrement on Events tab — verify stateful service works
- [ ] Verify Go→JS events (time ticker) continue working
- [ ] Verify no regression on desktop platforms (nil body guard is defensive only)

## Summary by CodeRabbit

* **New Features**
  * Enhanced Android integration with improved JNI bridge communication for runtime calls
  * Better message processing and error handling for platform-specific runtime operations

* **Bug Fixes**
  * Improved HTTP transport robustness in request handling

[Review Change Stack](https://app.coderabbit.ai/change-stack/wailsapp/wails/pull/5510?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/v3/internal/runtime/desktop/@wailsio/runtime/src/index.ts
+++ b/v3/internal/runtime/desktop/@wailsio/runtime/src/index.ts
@@ -64,7 +64,25 @@ export {
     clientId,
 } from "./runtime.js";
 
-import { clientId } from "./runtime.js";
+import { clientId, setTransport } from "./runtime.js";
+
+// On Android, route runtime calls through the JNI bridge instead of HTTP fetch
+// because Android's shouldInterceptRequest cannot access POST bodies.
+if ((window as any).wails?.invoke) {
+    setTransport({
+        call: async (objectID: number, method: number, windowName: string, args: any) => {
+            const body: any = { object: objectID, method };
+            if (args !== null && args !== undefined) {
+                body.args = args;
+            }
+            const responseStr = (window as any).wails.invoke(JSON.stringify(body));
+            if (typeof responseStr === "string" && responseStr.length > 0) {
+                try { return JSON.parse(responseStr); } catch { return responseStr; }
+            }
+            return responseStr;
+        }
+    });
+}
 
 // Notify backend
 window._wails.invoke = System.invoke;
--- a/v3/pkg/application/application_android.go
+++ b/v3/pkg/application/application_android.go
@@ -139,6 +139,7 @@ static void executeJavaScriptOnBridge(const char* js) {
 import "C"
 
 import (
+	"context"
 	"fmt"
 	"io"
 	"net/http"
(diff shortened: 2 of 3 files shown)


taliesin-ai:
Thanks for digging into this and putting up a fix, @amlwwalker — the Android binding-call crash you hit here is a real one and your diagnosis was spot on.

In the meantime the mobile layer got a substantial overhaul in #5602 (the new iOS/Android manager API), which **replaced the entire Android JS↔Go bridge** that this PR was patching. The new bridge (`androidBridge.invokeAsync` → `HandleRuntimeCallWithIDs`) resolves the same crash from the ground up, so this PR now conflicts against the very code that fixed it.

Because of that I'm going to close this as superseded by #5602 — not because the fix was wrong, but because the underlying code it touched no longer exists. If you pull latest `master` and still see the crash on Android, please reopen or file a fresh issue and we'll jump on it. Really appreciate the contribution. 🙏

PR #5539, opened 2026-06-04
Fix GTK4 Linux menu updates

### LINKED ISSUE #5464: [v3] Menu.Update() is a no-op after first call due to processed guard in processMenu (Linux)
## Description

`v3/pkg/application/menu_linux.go` (default GTK4 build) and `menu_linux_gtk3.go` (legacy GTK3 build) both contain a `linuxMenu.processed bool` field that is set to `true` on first call to `processMenu()` and used as a one-way guard:

```go
impl := menu.impl.(*linuxMenu)
if impl.processed {
    // Menu already processed, skip re-processing to avoid duplicates
    return
}
impl.processed = true
```

As a result, subsequent calls to `Menu.Update()` (which calls `m.update()` → `m.processMenu(m.menu)`) become silent no-ops after the first call. Adding/removing menu items, rebuilding after `Clear()`, or any other runtime mutation will not be reflected in the native menu.

## Reproduction

```go
menu := app.NewMenu()
menu.Add("Item A")
window.SetMenu(menu)

// Later, at runtime:
menu.Add("Item B")
menu.Update()  // expected: native menu now shows Item A + Item B. actual: still just Item A.
```

## Scope

- Affects **both** the default GTK4 build (this file, `menu_linux.go`) and the legacy GTK3 path (`menu_linux_gtk3.go`).
- Pre-existing — originated in #4958, surfaced by Copilot review on #5463 when the GTK4 code became the default path.

## Suggested fix

Instead of a one-way `processed` guard, either:

1. Clear/destroy the native menu and rebuild from scratch on every `Update()` call (simpler, slight flicker).
2. Diff the current `menu.items` against the native menu's state and apply minimal mutations (more code, no flicker).

Option 1 is probably right unless someone reports flicker; menu updates are rare enough that the simple path is preferable.

## References

- Copilot review on #5463: https://github.com/wailsapp/wails/pull/5463#discussion_r3251765580
- Original GTK4 implementation: #4958

puneetdixit200:
Fixes #5464

## Summary
- Rebuild GTK4 Linux menus on repeated `Menu.Update()` calls instead of returning early after the first process pass
- Add a GTK4 `menuClear` helper and reset menu item indexes before rebuilding
- Add a regression test and v3 changelog entry

## Tests
- `go test ./pkg/application -run TestLinuxGTK4MenuUpdateRebuildsProcessedMenus`
- `go test ./pkg/application`

Note: `coderabbit --plain` was not run because the command is not installed locally.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed GTK4 Linux menu updating behavior. Previously, calling `Menu.Update()` on a menu that had been rendered would be ignored entirely. Menus are now properly cleared and rebuilt during updates, ensuring that menu modifications take effect correctly even when updating menus that have been previously displayed.

### DIFF
--- a/v3/UNRELEASED_CHANGELOG.md
+++ b/v3/UNRELEASED_CHANGELOG.md
@@ -23,6 +23,7 @@ After processing, the content will be moved to the main changelog and this file 
 
 ## Fixed
 <!-- Bug fixes -->
+- Fix GTK4 Linux menus so `Menu.Update()` rebuilds already processed menus instead of becoming a no-op (#5464)
 
 ## Deprecated
 <!-- Soon-to-be removed features -->
--- a/v3/pkg/application/linux_cgo.c
+++ b/v3/pkg/application/linux_cgo.c
@@ -349,6 +349,13 @@ gboolean get_action_state(const char *action_name) {
     return FALSE;
 }
 
+void menu_clear(GMenu *menu) {
+    gint count = g_menu_model_get_n_items(G_MENU_MODEL(menu));
+    for (gint i = count - 1; i >= 0; i--) {
+        g_menu_remove(menu, i);
+    }
+}
+
 void menu_remove_item(GMenu *menu, gint position) {
     g_menu_remove(menu, position);
 }
--- a/v3/pkg/application/linux_cgo.go
+++ b/v3/pkg/application/linux_cgo.go
@@ -357,6 +357,21 @@ func menuNew() pointer {
 	return pointer(C.g_menu_new())
 }
 
+func menuClear(menu *Menu) {
+	if menu.impl == nil {
+		return
+	}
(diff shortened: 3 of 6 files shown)


leaanthony:
Thanks for this. I'm not sure of the value of the test as it just tests that the source code exists?

taliesin-ai:
Thank you for this, @puneetdixit200 — you correctly diagnosed that `Menu.Update()` wasn't rebuilding the native menu under GTK4, and your fix (clear + rebuild instead of the early-return) was exactly right.

As it happened, the same fix landed in parallel via #5659 just ahead of this, so I'm going to close this as already-resolved. That's purely a timing thing, not a reflection on the work — you independently found and solved a real bug. I'll make sure you're credited for it in the release notes alongside #5659. 🙏

Issue #5527, opened 2026-06-01
[v3] Freezes with open inspector on linux.


Mihara:
### Description

Since alpha 97, I am observing a complete freeze of the UI (and only the UI, the backend is entirely fine and doesn't even notice something has happened to Webkit) while the inspector is open. Once this happens, the OS asks you to kill the application, nothing gets it to unfreeze.

This is likely Linux-specific.

### To Reproduce

1. Create an application with `wails3 init -n freezetest`.
2. Add `OpenInspectorOnStartup: true,` into the `app.Window.NewWithOptions(application.WebviewWindowOptions{` struct, so that the application starts with the inspector already open.
3. Start the application with `wails3 dev`.
4. Tab away from the application for a few minutes.

Upon return, you will find the Webkit process frozen and unresponsive, with no log output.

### Expected behaviour

Nothing should ever freeze. :)

### System Details

```shell
Wails (v3.0.0-alpha.97)  Wails Doctor

# System

┌───────────────────────────────────────────────────────────────────────────────────────────────────────┐
| Name                | Ubuntu                                                                          |
| Version             | 24.04                                                                           |
| ID                  | ubuntu                                                                          |
| Branding            | 24.04.4 LTS (Noble Numbat)                                                      |
| Platform            | linux                                                                           |
| Architecture        | amd64                                                                           |
| Desktop Environment | ubuntu:GNOME                                                                    |
| NVIDIA Driver       | 595.80 (48FAD93C7D60A87FEA75CAB)                                                |
| XDG_SESSION_TYPE    | wayland                                                                         |
| CPU                 | AMD Ryzen 9 7940HS w/ Radeon 780M Graphics                                      |
| GPU 1               | AD106M [GeForce RTX 4070 Max-Q / Mobile] (NVIDIA Corporation) - Driver: nvidia  |
| GPU 2               | Phoenix1 (Advanced Micro Devices, Inc. [AMD/ATI]) - Driver: amdgpu              |
| Memory              | 31GB                                                                            |
└───────────────────────────────────────────────────────────────────────────────────────────────────────┘

# Build Environment

┌──────────────────────────────────────────────────────────────────────┐
| Wails CLI      | v3.0.0-alpha.97                                     |
| Go Version     | go1.26.3                                            |
| -buildmode     | exe                                                 |
| -compiler      | gc                                                  |
| CGO_CFLAGS     |                                                     |
| CGO_CPPFLAGS   |                                                     |
[...]

PR #5542, opened 2026-06-04
fix(linux): patch WebKit signal handlers with SA_ONSTACK after webview creation

leandrodaf:
## Problem

On Ubuntu 22.04+ (and any distro with webkit2gtk >= 2.36), Wails apps crash immediately on startup with:

```
fatal error: non-Go code set up signal handler without SA_ONSTACK flag
```

**Root cause:** `webkit_web_view_new_with_user_content_manager()` installs C-level handlers for SIGSEGV, SIGBUS, SIGFPE and related signals **without** the `SA_ONSTACK` flag. The Go runtime requires this flag on all signal handlers it does not own; when it detects a handler missing it, it calls `runtime.throw()` and the process dies.

## Fix

In `v2/internal/frontend/desktop/linux/window.c`, immediately after the WebKit view is created, patch all relevant signal handlers to include `SA_ONSTACK`:

```c
static void fix_webkit_signals(void) {
    struct sigaction sa;
    int sigs[] = {SIGSEGV, SIGBUS, SIGFPE, SIGILL, SIGABRT, SIGTRAP, 0};
    for (int i = 0; sigs[i]; i++) {
        if (sigaction(sigs[i], NULL, &sa) == 0) {
            sa.sa_flags |= SA_ONSTACK;
            sigaction(sigs[i], &sa, NULL);
        }
    }
}
// called right after webkit_web_view_new_with_user_content_manager()
```

This is the C-level equivalent of what `pkg/runtime.ResetSignalHandlers()` does, applied at exactly the right moment — after WebKit initializes its internal signal handling but before any JIT-triggered signals can fire.

## Tested on

- Ubuntu 24.04, webkit2gtk 2.52.3, Go 1.25, `webkit2_41` build tag

## Related

- `pkg/runtime/signal_linux.go` already exposes `ResetSignalHandlers()` but calling it from `OnStartup` is too late — WebKit JIT fires before that goroutine runs

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved a crash issue on Linux affecting the application when initializing the web view component.

### DIFF
--- a/v2/internal/frontend/desktop/linux/window.c
+++ b/v2/internal/frontend/desktop/linux/window.c
@@ -7,6 +7,35 @@
 #include <string.h>
 #include <locale.h>
 #include "window.h"
+#include <signal.h>
+
+/*
+ * fix_webkit_signals - Re-apply SA_ONSTACK to WebKit signal handlers.
+ *
+ * webkit_web_view_new_with_user_content_manager() installs C-level handlers
+ * for SIGSEGV, SIGBUS, SIGFPE and related signals via sigaction(2) without
+ * the SA_ONSTACK flag.  The Go runtime requires SA_ONSTACK on every signal
+ * handler it does not own; its absence causes an immediate fatal throw:
+ *
+ *   fatal error: non-Go code set up signal handler without SA_ONSTACK flag
+ *
+ * This function reads each installed handler and ORs SA_ONSTACK into its
+ * flags.  It is called once, immediately after webview creation, so the fix
+ * is in place before any JIT-triggered or Gigacage-triggered signal fires.
+ *
+ * Only signals with a real handler (not SIG_DFL / SIG_IGN) are modified;
+ * signals not yet claimed by WebKit are left unchanged.
+ */
+static void fix_webkit_signals(void) {
+    struct sigaction sa;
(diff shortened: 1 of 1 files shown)


leaanthony:
Thanks for the detailed write-up and the clear root-cause analysis!

You may want to take a look at #5507, which was merged on 2026-05-27 and targets the same SA_ONSTACK issue. It installs the fix from a C-level periodic timer (every 50 ms for 5 s, anchored to first webview creation) as well as at `DomReady` — so it fires well before `OnStartup`, which addresses the same timing concern you raised.

That said, we can't be certain whether this is a duplicate without knowing which version of Wails you're testing against. Could you let us know the output of `wails doctor`? That will tell us whether you're on a build that already includes #5507 or not — if you're still seeing the crash on a version that has it, there's clearly still a gap worth fixing.

One other thing worth noting: #5507 intentionally skips `SIGUSR1` — WebKit's JavaScriptCore uses that signal to suspend/resume threads for conservative GC stack scanning, and adding `SA_ONSTACK` to it was found to break GC thread synchronisation (see #5527). Your patch patches `SIGUSR1` as well, so that's worth double-checking on your end.

Issue #4835, opened 2025-12-27
[v3] `[WebView2 Error] The parameter is incorrect.`


bicarlsen:
### Description

During `wails3 dev`, when saving a backend (`.go`) file, during recompilation I get the error
```sh
2025/12/27 23:10:32 ERROR Process exited with error exec="wails3 task run" err="exit status 1"
2025/12/27 23:10:32 ERROR Could not refresh processes due to execution errors
task: [windows:run] bin/my_app.exe
{"time":"2025-12-27T23:10:33.3433239+01:00","level":"ERROR","source":{"function":"github.com/wailsapp/wails/v3/pkg/application.(*App).handleError","file":"C:/Users/user/go/pkg/mod/github.com/wailsapp/wails/v3@v3.0.0-alpha.52/pkg/application/application.go","line":443},"msg":"SetProcessDpiAwarenessContext failed 0: 18446744073709551612 Access is denied."}
2025/12/27 23:10:33 [WebView2] Environment created successfully
[WebView2 Error] The parameter is incorrect.

Stack trace:
2025/12/27 23:10:33 1: github.com/wailsapp/go-webview2/pkg/edge.(*Chromium).errorCallback
        C:/Users/user/go/pkg/mod/github.com/wailsapp/go-webview2@v1.0.23/pkg/edge/chromium.go:151
2025/12/27 23:10:33 2: github.com/wailsapp/go-webview2/pkg/edge.(*Chromium).Focus
        C:/Users/user/go/pkg/mod/github.com/wailsapp/go-webview2@v1.0.23/pkg/edge/chromium.go:575
2025/12/27 23:10:33 3: github.com/wailsapp/wails/v3/pkg/application.(*windowsWebviewWindow).focus
        C:/Users/user/go/pkg/mod/github.com/wailsapp/wails/v3@v3.0.0-alpha.52/pkg/application/webview_window_windows.go:935
2025/12/27 23:10:33 4: github.com/wailsapp/wails/v3/pkg/application.(*windowsWebviewWindow).WndProc
        C:/Users/user/go/pkg/mod/github.com/wailsapp/wails/v3@v3.0.0-alpha.52/pkg/application/webview_window_windows.go:1439
2025/12/27 23:10:33 5: github.com/wailsapp/wails/v3/pkg/application.(*windowsApp).wndProc
        C:/Users/user/go/pkg/mod/github.com/wailsapp/wails/v3@v3.0.0-alpha.52/pkg/application/application_windows.go:303
[...]

Issue #5544, opened 2026-06-05
[Windows] Frameless window reports incorrect innerWidth/innerHeight after minimize and restore on high-DPI displays


cnmax:
### Description

When using a frameless window (Frameless: true) on Windows, minimizing and restoring the application can cause the frontend viewport size to become inconsistent with the actual window size.

After restoring the window, window.devicePixelRatio, window.innerWidth, and window.outerWidth report conflicting values. The application UI appears incorrectly scaled as if an outdated DPI value or viewport size is still being used.

The issue is reproducible on a high-DPI display and appears to be related to viewport/DPI synchronization after window restoration.

### To Reproduce

Window configuration:

func newMainWindow(app *application.App) *application.WebviewWindow {

    return app.Window.NewWithOptions(application.WebviewWindowOptions{
		Name:       "main",
		Title:      "App",
		Frameless:  true,
		StartState: application.WindowStateMaximised,
	})
}

Frontend logging:

function dump(tag: string) {

    console.log(tag, {
        dpr: window.devicePixelRatio,
        innerWidth: window.innerWidth,
        innerHeight: window.innerHeight,
        outerWidth: window.outerWidth,
        outerHeight: window.outerHeight,
    })
}

window.addEventListener("resize", () => dump("resize"))
window.addEventListener("focus", () => dump("focus"))

Reproduction steps:

Launch the application.
The window starts maximized.
Minimize the window.
Restore the window.
Observe the logged viewport values and UI scaling.

### Expected behaviour

After restoring the window, viewport values should match the actual restored window size.

Example:

dpr: 2
innerWidth: 1920
innerHeight: 1080
outerWidth: 1920
outerHeight: 1080

The UI should render at the correct scale.

Actual Behavior

During restore, the following sequence can be observed:

resize {
  dpr: 1.25,
  innerWidth: 1920,
  innerHeight: 1080,
  outerWidth: 3072,
  outerHeight: 1728
}

resize {
  dpr: 1.25,
  innerWidth: 3072,
  innerHeight: 1728,
  outerWidth: 3072,
  outerHeight: 1728
}

focus {
  dpr: 2,
  innerWidth: 3072,
  innerHeight: 1728,
  outerWidth: 1920,
  outerHeight: 1080
}

resize {
  dpr: 2,
  innerWidth: 3072,
  innerHeight: 1728,
  outerWidth: 1920,
  outerHeight: 1080
}

The final state remains inconsistent:

dpr: 2
innerWidth: 3072
[...]

Issue #4158, opened 2025-03-21
崩溃


qtgolang:
### Description

启动时有概率崩溃，概率可能很小，这是错误信息
`

2025/03/21 15:19:31 [WebView2] Environment created successfully
[WebView2 Error] The parameter is incorrect.
2025/03/21 15:19:31 1: github.com/wailsapp/go-webview2/pkg/edge.(*Chromium).errorCallback
        D:/Go/PATH/pkg/mod/github.com/wailsapp/go-webview2@v1.0.19/pkg/edge/chromium.go:151

2025/03/21 15:19:31 2: github.com/wailsapp/go-webview2/pkg/edge.(*Chromium).Focus
        D:/Go/PATH/pkg/mod/github.com/wailsapp/go-webview2@v1.0.19/pkg/edge/chromium.go:580
Stack trace:
2025/03/21 15:19:31 3: github.com/wailsapp/wails/v3/pkg/application.(*windowsWebviewWindow).focus
        D:/Go/PATH/pkg/mod/github.com/wailsapp/wails/v3@v3.0.0-alpha.9/pkg/application/webview_window_windows.go:718
2025/03/21 15:19:31 4: github.com/wailsapp/wails/v3/pkg/application.(*windowsWebviewWindow).WndProc
        D:/Go/PATH/pkg/mod/github.com/wailsapp/wails/v3@v3.0.0-alpha.9/pkg/application/webview_window_windows.go:1080
2025/03/21 15:19:31 5: github.com/wailsapp/wails/v3/pkg/application.(*windowsApp).wndProc
        D:/Go/PATH/pkg/mod/github.com/wailsapp/wails/v3@v3.0.0-alpha.9/pkg/application/application_windows.go:256
2025/03/21 15:19:31 6: runtime.callbackWrap
        D:/Go/go1.24.0/src/runtime/syscall_windows.go:396
2025/03/21 15:19:31 7: runtime.cgocallbackg1
        D:/Go/go1.24.0/src/runtime/cgocall.go:444
2025/03/21 15:19:31 8: runtime.cgocallbackg
        D:/Go/go1.24.0/src/runtime/cgocall.go:350
2025/03/21 15:19:31 9: runtime.cgocallback
        D:/Go/go1.24.0/src/runtime/asm_amd64.s:1084
2025/03/21 15:19:31 10: runtime.systemstack_switch
        D:/Go/go1.24.0/src/runtime/asm_amd64.s:479
[...]

PR #5568, opened 2026-06-10
fix(webview2): guard Focus() until controller initialisation completes

### LINKED ISSUE #5446: [v3] (*Chromium).Focus race on production startup — #4835 fix has TOCTOU window
### Description

Hitting the same crash signature as #4835 in **alpha.90 production build** (not dev mode). The fix from #4835 (commit 464659b — \`GetController() != nil\` guards in \`webview_window_windows.go\`) is present in our build, but the check and the subsequent \`Focus()\` call are non-atomic: in our crash trace the controller has been torn down between the guard at line 1023 and the call at line 1028.

### Stack trace (production launcher.log, ~4 s after \`App.Run\`)

\`\`\`
[WebView2 Error] The parameter is incorrect.

1: github.com/wailsapp/wails/webview2/pkg/edge.(*Chromium).errorCallback
   github.com/wailsapp/wails/webview2@v1.0.24/pkg/edge/chromium.go:151
2: github.com/wailsapp/wails/webview2/pkg/edge.(*Chromium).Focus
   github.com/wailsapp/wails/webview2@v1.0.24/pkg/edge/chromium.go:575
3: github.com/wailsapp/wails/v3/pkg/application.(*windowsWebviewWindow).focus
   github.com/wailsapp/wails/v3@v3.0.0-alpha.90/pkg/application/webview_window_windows.go:1028
4: github.com/wailsapp/wails/v3/pkg/application.(*windowsWebviewWindow).WndProc
   github.com/wailsapp/wails/v3@v3.0.0-alpha.90/pkg/application/webview_window_windows.go:1546
5: github.com/wailsapp/wails/v3/pkg/application.(*windowsApp).wndProc
   github.com/wailsapp/wails/v3@v3.0.0-alpha.90/pkg/application/application_windows.go:303
\`\`\`

(Identical shape to the trace in #4835, just from a production binary instead of \`wails3 dev\` hot-reload.)

### Race

\`v3/pkg/application/webview_window_windows.go\` alpha.90:

\`\`\`go
1009: func (w *windowsWebviewWindow) focus() {
        ...
1023:   if w.chromium.GetController() == nil {
1024:       return
1025:   }
        // ── another goroutine can tear the controller down here ──
1028:   w.chromium.Focus()   // crashes when controller is now nil/destroyed
      }
\`\`\`

The other three \`chromium.Focus()\` call sites in the same file (lines 884–885, 896–897, 945–946) have the same shape — \`GetController() != nil\` then call — so the same TOCTOU window exists for each.

### Trigger

Not deterministic — surfaces ~1 in 10 cold launches of a Wails v3 launcher on Windows 11, ~4 s after \`App.Run\`. The window message landing in \`WndProc\` while WebView2's controller is still initialising seems to be the trigger. dev hot-reload (the #4835 reproducer) is not involved.

### Suggested fix direction

1. Snapshot the controller pointer under \`w.chromium\`'s lock inside \`focus()\` (and the other three sites) and call \`Focus()\` only on the snapshot, OR
2. Defer focus messages until the controller has finished initialising (e.g. queue them and drain after \`WebView2 Environment created successfully\` fires).

Either keeps the fix local to the affected call sites without changing the broader controller lifecycle.

### Versions

- Wails v3: \`v3.0.0-alpha.90\`
- Windows 10/11
- WebView2 runtime: whatever Windows ships
- Build: production (\`-tags production -H windowsgui\`), tag-triggered release workflow

Cross-ref: #4835 (closed, partial fix landed in 464659b — covers nil controller but not the race).

leaanthony:
# Description

Fixes the production startup crash from #5446 (successor to the partial #4835 fix).

`CreateCoreWebView2ControllerCompleted` assigns `e.controller` partway through the callback (chromium.go:338) but only flips the `inited` atomic after **all** COM setup succeeds (line 389). The Windows message pump runs during that window, so the WndProc can dispatch `WM_SETFOCUS` re-entrantly and reach `Chromium.Focus()` while the controller is non-nil but partially configured. The v3-side `GetController() != nil` guards (added for #4835) pass, `MoveFocus` fails with `E_INVALIDARG` ("The parameter is incorrect"), and `errorCallback` exits the process. With no controller at all, `Focus()` dereferenced nil.

## Fix

- Add `Chromium.IsReady()` — atomic load of `inited`, the flag that already means "fully initialised".
- `Focus()` becomes a no-op until `IsReady()`. Dropping a focus request during startup is harmless (the window manager re-delivers focus on interaction); calling into a partially-initialised controller is fatal.

Guarding inside `Focus()` makes **every** call site safe — including the four TOCTOU-shaped `GetController() != nil → Focus()` patterns in v3's `webview_window_windows.go` — without v3 code changes. Once this ships in a webview2 release, the existing `bump-webview2-v3` automation picks it up; the v3 call sites can optionally migrate to `IsReady()` afterwards as a cleanup.

## Verification (win-node1, windows/amd64, go1.26.2)

- **Negative control**: a Focus-before-init probe against unfixed `chromium.go` crashes with `MoveFocus(0x0, …)` nil-pointer panic — the reported failure mode.
- **With the fix**: `TestFocusBeforeControllerReady` and `TestFocusBeforeControllerReadyConcurrent` (16 goroutines × 1000 iterations) pass; full `pkg/edge` suite passes; `go vet` clean for the changed file (remaining warnings are pre-existing COM interop patterns).

Fixes #5446

## Type of change

## Checklist:

- [x] I have added tests that prove my fix is effective

## Summary by CodeRabbit

* **Bug Fixes**
  * Focus requests during controller startup are safely ignored to prevent errors or crashes.
  * Focus failures after initialization are logged instead of terminating the process.
  * Script execution failures that previously caused termination are now treated as recoverable and logged.

* **Tests**
  * Added Windows-only tests confirming focus is a safe no-op before initialization, including concurrent stress tests.


leaanthony:
Scope update: while triaging #4158 (same crash signature, reported a year earlier) a commenter provided a **100% reproducible** variant: minimize to tray → click the tray icon → instant crash. That path fails *after* initialisation — `MoveFocus` returns `E_INVALIDARG` for a hidden/tray-minimised window — so the `IsReady()` guard alone doesn't cover it.

Added one commit: `Focus()` no longer routes `MoveFocus` errors through `errorCallback` (which exits the process); a failed focus request is logged instead. Between the init guard and the non-fatal error path, both the startup race (#5446) and the tray-restore crash (#4158) are covered. Build + focus tests re-verified on win-node1.

leaanthony:
Scope addition from #5544's verification: dragging a window between mixed-DPI monitors exposed that `Eval` also routes transient COM errors (`RESOURCE_NOT_IN_CORRECT_STATE` during the DPI transition) through the fatal `errorCallback`. Same disease, same cure as Focus: fire-and-forget script execution now logs failures instead of exiting the process. Build + edge suite re-verified on win-node1.

PR #5581, opened 2026-06-11
Fix TypeScript interface generation missing `| null` for slice/map fields

### LINKED ISSUE #3588: Typescript interface generation doesn't include null for slice or map
### Description

When wails' ts generation output type is set to `interfaces` the resulting interfaces do not mark Golang slices as potentially null-able. This can easily lead to front-end code crashing when using the interfaces because there is no reason to do a null check if the interfaces says the field can't be null. Since the default value for a slice in Go is nil, its quite easy to accidentally pass a nil value to Javascript and crash it. 

### To Reproduce

1. `wails init -n slicetest`
2. Edit wails.json to use "interfaces" type for ts_generation:
```json
{
  ...
  "bindings": {
    "ts_generation": {
      "outputType": "interfaces"
    }
  }
}
```
4. Edit main.go to include a slice as one of the types:
```golang
package main

import (
	"embed"

	"github.com/wailsapp/wails/v2"
	"github.com/wailsapp/wails/v2/pkg/options"
)

//go:embed all:frontend/dist
var assets embed.FS

type TestStruct struct {
	TestSlice []int `json:"test_slice"`
}

// App struct
type App struct {
}

func (a *App) NilSlice() TestStruct {
	var t TestStruct
	return t
}

func main() {
	// Create an instance of the app structure
	app := &App{}

	// Create application with options
	err := wails.Run(&options.App{
		Bind: []interface{}{
			app,
		},
	})

	if err != nil {
		println("Error:", err.Error())
	}
}
```
5. Run `wails generate module`
6. Inspect the resulting models.ts file which uses `number[]` instead of `number[] | null`
```typescript
export namespace main {
	
	export interface TestStruct {
	    test_slice: number[];
	}

}
```

### Expected behaviour

I would expect either the type interface to indicate that a Golang slice could be null when passed to Javascript, or for wails to prevent the Golang Json serializer from converting `[]int(nil)` to `null` in Json before its passed to JS.  

The second option is likely preferred as most of the time it would be annoying to always null check a slice before using it in JS, however that would prevent some use cases. Maybe a setting in the wails.json could be added if we want to go that route. 

### Screenshots

_No response_

### Attempted Fixes

After looking at this some more, I realized that its because in my project I have the typescript `strict` setting set `true`: 
https://www.typescriptlang.org/tsconfig/#strict

That causes the `strictNullChecks` setting to be enabled as well, which is why I am expecting the `| null` in the interface. 
https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#strictnullchecks-on

I think the best solution would be to provide a flag to the typescript generation section in wails.json (something `struct: true`) to output the correct interfaces for typescript's strict mode. 

### System Details

```shell
# Wails
Version | v2.9.1

# System
| OS           | Windows 10 Pro                                          |
| Version      | 2009 (Build: 19045)                                     |
| ID           | 22H2                                                    |
| Go Version   | go1.21.11                                               |
| Platform     | windows                                                 |
| Architecture | amd64                                                   |
| CPU          | Intel(R) Core(TM) i7-6950X CPU @ 3.00GHz                |
| GPU 1        | NVIDIA TITAN X (Pascal) (NVIDIA) - Driver: 31.0.15.4601 |
| GPU 2        | NVIDIA TITAN X (Pascal) (NVIDIA) - Driver: 31.0.15.4601 |
| Memory       | 32GB                                                    |
└────────────────────────────────────────────────────────────────────────┘

# Dependencies
| WebView2   | N/A          | Installed | 126.0.2592.87 |
| Nodejs     | N/A          | Installed | 20.15.0       |
| npm        | N/A          | Installed | 10.7.0        |
| *nsis      | N/A          | Available |               |
```

### Additional context

_No response_

lesbass:
## Description

When `bindings.ts_generation.outputType` is set to `"interfaces"` in `wails.json`, the generated TypeScript interfaces do not include `| null` for Go slice and map fields. Go slices and maps can be `nil`, which serializes to `null` in JSON. This causes issues for TypeScript projects using `strictNullChecks`, since the interface claims the field can never be `null`.

This PR adds `| null` to slice, array, and map field types when generating TypeScript interfaces.

## Changes

- Added `createInterface` flag to `typeScriptClassBuilder` struct
- In `AddSimpleArrayField`, append `| null` to the generated type when in interface mode
- In `AddArrayOfStructsField`, append `| null` to the generated type when in interface mode
- In `AddMapField`, append `| null` to the generated type when in interface mode
- Added test case `DeepElementsAsInterfaces` verifying the fix
- Updated changelog under `[Unreleased]`

## Before

```typescript
export interface TestStruct {
    test_slice: number[];
    test_map: Record<string, string>;
    test_structs: ChildEntity[];
}
```

## After

```typescript
export interface TestStruct {
    test_slice: number[] | null;
    test_map: Record<string, string> | null;
    test_structs: ChildEntity[] | null;
}
```

## Verification

All existing tests pass, and the new `DeepElementsAsInterfaces` test case passes:
```
--- PASS: TestBindings_GenerateModels/DeepElements (0.00s)
--- PASS: TestBindings_GenerateModels/DeepElementsAsInterfaces (0.00s)
...
PASS
ok  	github.com/wailsapp/wails/v2/internal/binding/binding_test	0.010s
```

Fixes #3588

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed TypeScript interface generation to correctly include `| null` for array and map (collection) fields.
* **Tests**
  * Added coverage to verify model generation emits `export interface` types with nullable collection fields.
* **Documentation**
  * Updated the changelog entry to reflect the corrected `| null` behavior for slice and map fields in TypeScript interfaces.

### DIFF
--- a/.github/workflows/auto-changelog-v3.yml
+++ b/.github/workflows/auto-changelog-v3.yml
@@ -22,7 +22,7 @@ jobs:
       contents: write
 
     steps:
-    - name: Checkout master
+    - name: Checkout v3-alpha
       uses: actions/checkout@v4
       with:
         ref: master
--- a/.github/workflows/build-and-test-v3.yml
+++ b/.github/workflows/build-and-test-v3.yml
@@ -7,9 +7,6 @@ on:
       - master
     paths:
       - 'v3/**'
-      - 'webview2/**'
-      - 'go.work'
-      - 'go.work.sum'
       - '.github/workflows/build-and-test-v3.yml'
   pull_request_review:
     types: [submitted]
@@ -98,6 +95,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     if: github.base_ref == 'master'
     timeout-minutes: 25
+    env:
+      GOWORK: "off"
     strategy:
       fail-fast: false
       matrix:
@@ -127,10 +126,7 @@ jobs:
         with:
           go-version: ${{ matrix.go-version }}
           cache: true
-          cache-dependency-path: |
-            v3/go.sum
-            webview2/go.sum
-            go.work.sum
+          cache-dependency-path: "v3/go.sum"
 
       - name: Install Task
         uses: arduino/setup-task@v2
@@ -204,6 +200,8 @@ jobs:
     needs: [test_js, test_go]
     runs-on: ${{ matrix.os }}
     if: github.base_ref == 'master'
+    env:
+      GOWORK: "off"
     strategy:
       fail-fast: false
       matrix:
@@ -245,10 +243,7 @@ jobs:
         with:
           go-version: ${{ matrix.go-version }}
           cache: true
-          cache-dependency-path: |
-            v3/go.sum
-            webview2/go.sum
-            go.work.sum
+          cache-dependency-path: "v3/go.sum"
 
       - name: Install Task
         uses: arduino/setup-task@v2
@@ -284,9 +279,6 @@ jobs:
           # Replace @wailsio/runtime version with local tarball
           npm pkg set dependencies.@wailsio/runtime="file://$RUNTIME_TGZ"
           cd ..
-          # Register the generated project in the workspace so Go 1.25 workspace
-          # mode does not reject packages in a module not listed in go.work.
-          go work use .
           wails3 build
 
       # GTK3 legacy template builds are covered by the Go example compilation tests above.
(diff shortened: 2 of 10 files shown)


lesbass:
Closing this PR after 18 days with no human maintainer feedback. The change is still valid — happy to reopen if maintainers are interested. Thanks for the automated review from CodeRabbit.

PR #5584, opened 2026-06-12
fix(darwin): copy screen id/name C strings and snapshot screen count atomically

x-haose:
## Description

Fixes a hard crash (`SIGSEGV` in `C.GoString`) that occurs during display-reconfiguration event storms — typically when a Mac wakes from sleep, connects/disconnects a monitor, or changes resolution while a v3 app is running.

### Crash observed (v3.0.0-alpha.97, macOS 26.5, arm64)

```
fatal error: fault
[signal SIGSEGV: segmentation violation code=0x2 addr=0x100000000 pc=...]
runtime.gostring(0x10000001b)
github.com/wailsapp/wails/v3/pkg/application._Cfunc_GoString
github.com/wailsapp/wails/v3/pkg/application.cScreenToScreen        screen_darwin.go:185
github.com/wailsapp/wails/v3/pkg/application.(*macosApp).processAndCacheScreens
github.com/wailsapp/wails/v3/pkg/application.(*macosApp).run.func3
github.com/wailsapp/wails/v3/pkg/application.(*EventManager).handleApplicationEvent.func1
```

The app had been running ~12.5h and crashed on wake.

### Root causes (two memory-safety issues in `screen_darwin.go`)

1. **Dangling C strings.** `processScreen` stores the raw result of `-[NSString UTF8String]` into the `Screen` struct:
   ```objc
   returnScreen.id   = [[NSString stringWithFormat:@"%d", displayID] UTF8String];
   returnScreen.name = [screen.localizedName UTF8String];
   ```
   That pointer is owned by the (autoreleased) `NSString`. Go reads it later via `C.GoString` — once the string is deallocated, the pointer dangles. `0x10000001b` above is exactly such a garbage pointer.

2. **Count/allocation race.** `getAllScreens()` mallocs the array from one `[NSScreen screens]` snapshot, while `processAndCacheScreens` fetches the element count via a *separate* `GetNumScreens()` call. If the display configuration changes between the two calls (precisely when these code paths run), the loop over-reads the allocation.

### Fix

- `copyCString` helper: `strdup` the UTF-8 bytes; the Go side frees them after conversion in `cScreenToScreen`, keeping every `processScreen` caller leak-free.
- `getAllScreens(int* outCount)`: return the count from the same snapshot as the allocation.
- Zero-initialize the `Screen` struct and guarantee `id`/`name` are always owned, non-NULL pointers (empty string when `localizedName` is unavailable or the `NSScreen` is nil) — this also removes a latent `C.GoString(NULL)` hazard via the `getScreenForWindow` / `getScreenForSystemTray` paths (e.g. window not currently on any screen).

No public API changes; `getAllScreens` is internal to the file.

### Testing

- `go build ./pkg/application` on macOS 26.5 / arm64 / go1.25.
- Downstream app ([x-haose/tg-checkin](https://github.com/x-haose/tg-checkin), long-running menubar-resident app) now runs with this patch via `go.mod replace`; previously crashed on wake, will report back after soak.

## Type of change

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved screen detection reliability by preventing race conditions during display reconfiguration.

* **Refactor**
  * Enhanced memory safety and management of internal screen data structures.


taliesin-ai:
Thanks for this, @x-haose — you correctly spotted that macOS screen enumeration could hand back pointers into Objective-C strings that get freed, and reading the screen count separately from the list was a real race. Solid catch.

The same fix landed just ahead of this via #5565, which added `strdupOrNull` for the id/name and a single `getAllScreens(int* outCount)` that returns the list and count together — the same approach as yours. So I'm closing this as already-resolved, purely on timing. You'll be credited in the release notes for independently finding and fixing it. 🙏

(Note for posterity: #5516 tackles the same area from a different angle — marshalling enumeration onto the main thread — so that protection is being tracked separately.)

PR #5630, opened 2026-06-18
fix windows minimize crash/flash

atterpac:
This pull request adds additional safety checks to the Windows webview window handling code to prevent operations on an uninitialized or not-ready `chromium` instance. These checks help avoid potential runtime errors during window resizing and DPI resynchronization.

**Stability improvements:**

* Added a guard in the `WndProc` method to ensure `w.chromium` is not `nil` and is ready before calling `ResizeWithBounds`.
* Added a guard in the `resyncWebviewDPIAfterMinimise` method to ensure `w.chromium` is not `nil` and is ready before proceeding with DPI resynchronization logic.

** UNTESTED **

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced Windows WebView window stability during resize and DPI synchronization operations.


leaanthony:
Thanks @atterpac. The minimise/restore crash this targets is being fixed at the root in #5616 (now rebased, cross-compile green across all platforms) by gating the WebView2 DPI resync so it doesn't touch the controller on an ordinary same-DPI restore. Closing as a duplicate in favour of #5616 — appreciate you digging into it.

PR #5636, opened 2026-06-20
Fix/v3/windows: Move resizePending from the package scope to be a per-window field on the windowsWebviewWindow struct

sinspired:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)

## Type of change
  
# How Has This Been Tested?
  
- [x] Windows
- [ ] macOS
- [ ] Linux
      
## Test Configuration

# Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved WebView resize handling for frameless windows during minimization
  * Enhanced DPI synchronization during minimize/restore operations, reducing redundant updates
  * Fixed crash risk during window minimize/restore transitions
  * Optimized DPI change detection timing

### DIFF
--- a/v3/pkg/application/webview_window_windows.go
+++ b/v3/pkg/application/webview_window_windows.go
@@ -76,6 +76,14 @@ type windowsWebviewWindow struct {
 	// Used to prevent unnecessary redraws during minimize/restore operations
 	isMinimizing bool
 
+	// dpiBeforeMinimise records the window's DPI immediately before it entered
+	// the minimised state, so that on restore we can tell whether the DPI
+	// genuinely changed while the window was parked (the cross-monitor #5544
+	// scenario), as opposed to unconditionally touching WebView2's Controller3
+	// rasterization-scale APIs on every single minimise/restore cycle, even on
+	// a single-monitor, single-DPI setup where nothing needs correcting.
+	dpiBeforeMinimise w32.UINT
+
 	// lastSizeWParam is the wParam from the most-recent WM_SIZE message.
 	// Gate the dark-menubar force-repaint on a state transition so it does not fire
 	// on every SIZE_RESTORED during live drag-resize. WM_ENTERSIZEMOVE/WM_EXITSIZEMOVE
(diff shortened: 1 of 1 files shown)


taliesin-ai:
Thanks for digging into the mixed-DPI minimize/restore crashes — this is a real problem area. Reviewing against current `master`, though, most of this PR has been overtaken by other changes that have landed since it was opened, so it'll need a rebase before it can be considered:

- **The titled change (package-scope `resizePending` → per-window field)** no longer applies: `resizePending` has been removed from `webview_window_windows.go` entirely, so there's nothing left to refactor.
- **The `if !w.isMinimizing` guards** around the rasterization-scale resync are already present on `master`.
- **The `GetClientRect` nil-guard** in the WM_ERASEBKGND paint is currently dead code: `w32.GetClientRect` (`v3/pkg/w32/user32.go`) **panics** on failure rather than returning `nil`, so `rc != nil` is always true. (The same inert nil-check already exists in `convertWindowToWebviewCoordinates`.) The real fix for that transient-state crash is to make `GetClientRect` return `nil`/error instead of panicking, then make both guards live — that's a separate change with a caller audit.

The one piece that looks genuinely new is the `dpiBeforeMinimise` delta-optimization (skip the Controller3 resync when DPI hasn't actually changed). If you'd like to pursue that, could you rebase onto `master` and split it into a focused PR? It'd also help to have a concrete repro for the "blank/crash on ordinary minimise/restore" regression it's meant to address, since that path was verified by the #5544 reporters on the current approach.

I'd suggest closing this in favor of a rebased follow-up. Thanks again — happy to review the smaller version.

taliesin-ai:
Thanks for this, @sinspired. The main change here — moving `resizePending` from package scope to a per-window field — has unfortunately been overtaken: #5596 removed the resize-debounce mechanism entirely (so there's no longer a shared `resizePending` to move), and #5616 reworked the DPI-on-restore path this also touched. So the PR as a whole no longer applies.

But your nil-rect guard before `FillRect` in `WM_ERASEBKGND` was a real, still-unfixed crash hardening — master still calls `FillRect` with an unchecked `GetClientRect`. I've salvaged exactly that into #5687, credited to you. Closing this in favour of that. Appreciate the careful work. 🙏

PR #5635, opened 2026-06-20
Fix/V3/Windows: blank content and crash when restoring minimized window (introduced by #5572)

sinspired:
Fix a Windows regression introduced by #5572 where restoring a minimized window could leave the content area blank,
and in some cases crash the app.

This regression was introduced in v3.0.0-alpha.102 and has been reproducible in all subsequent alpha releases tested (v3.0.0-alpha.103 and v3.0.0-alpha.104).
The last known working version is v3.0.0-alpha.101, where minimizing to tray and restoring worked correctly without blank content or crashes.

On Windows, 52b352994 (#5572) unconditionally resynced WebView2 rasterization scale in the minimize/restore path. That caused even normal same-DPI minimize/restore cycles to enter the WebView2 reconfiguration path and race with ResizeWithBounds() during restore, which could leave the content area blank and, in some cases, crash the app.

This change records the DPI before minimizing and only resyncs WebView2 rasterization scale when the DPI after restore actually differs from the pre-minimize DPI. It keeps the cross-monitor DPI fix needed for #5544 while avoiding unnecessary Controller3 updates during ordinary minimize/restore cycles.

- Record the window DPI before minimizing.
- Only resync WebView2 rasterization scale when the restore DPI actually differs from the pre-minimize DPI.
- Keep the fix for cross-monitor DPI changes, so #5544 remains covered.
- Make the normal minimize/restore path a no-op for WebView2 DPI sync, avoiding unneces

# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)

## Type of change
  
Please select the option that is relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?
  
Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration using `wails doctor`.

- [x] Windows
- [ ] macOS
- [ ] Linux
      
If you checked Linux, please specify the distro and version.
  
## Test Configuration

Please paste the output of `wails doctor`. If you are unable to run this command, please describe your environment in as much detail as possible.

# Checklist:

- [ ] (v2 only) I have updated `website/src/pages/changelog.mdx` with details of this PR (v3 changelog entries are added automatically)
- [x] My code follows the general coding style of this project
[...]

PR #5637, opened 2026-06-21
fix(v3/windows): guard WM_ERASEBKGND paint against nil GetClientRect

leaanthony:
## Summary

On Windows, the `WM_ERASEBKGND` solid-background paint passes the result of `GetClientRect` straight into `FillRect`. `GetClientRect` can return `nil` for a window in a transient state (the same nil case already guarded in `convertWindowToWebviewCoordinates`), which minimise/restore transitions are especially prone to trigger — and `FillRect` with a nil rect crashes. This adds the missing nil check before painting.

## Test

- Cross-compiles for windows/amd64 and windows/arm64.
- Built and run on a Windows 11 box (WebView2 149): solid-background window survives repeated minimise → wait → restore cycles, no crash and no blank content.

This fix was salvaged from the equivalent change in #5635; credit to @sinspired via the `Co-authored-by` trailer.

## Changelog

`v3/UNRELEASED_CHANGELOG.md` updated under Fixed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a crash on Windows that occurred when minimizing or restoring windows with solid background colors.

### DIFF
--- a/v3/UNRELEASED_CHANGELOG.md
+++ b/v3/UNRELEASED_CHANGELOG.md
@@ -23,6 +23,7 @@ After processing, the content will be moved to the main changelog and this file 
 
 ## Fixed
 <!-- Bug fixes -->
+- Fix crash on Windows when erasing the background during a minimise/restore transition: `GetClientRect` can return nil for a window in a transient state, and the solid-background paint passed it straight into `FillRect`. Now guarded, matching the existing nil-check in `convertWindowToWebviewCoordinates` (salvaged from #5635)
 
 ## Deprecated
 <!-- Soon-to-be removed features -->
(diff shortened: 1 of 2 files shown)


leaanthony:
Closing this — the Copilot review is correct and it invalidates the premise.

`w32.GetClientRect` (v3/pkg/w32/user32.go) never returns nil: it returns `&rect` on success and `panic`s if the underlying Win32 call fails. So the `rc != nil` guard here is always true — dead code that prevents nothing. The change was modelled on the nil-check in `convertWindowToWebviewCoordinates`, but that check is dead code for the same reason.

There's also no evidence of an actual crash in the `WM_ERASEBKGND` path — no stack trace implicating `FillRect`/`GetClientRect`. The real "blank content/crash on restore" was the DPI-resync regression, fixed by #5616 (merged). Guarding the (unreachable) `GetClientRect` panic here would be speculative hardening with no evidence behind it, so I'm not repurposing the PR for that.

Thanks @copilot for the catch. If a real stack trace ever implicates this path, it's worth a separate, evidence-backed change.

PR #5638, opened 2026-06-21
ci(webview2): auto-release a new tag when webview2 source changes

leaanthony:
## Problem

The `webview2` module is versioned and released separately from `v3`, and `v3/go.mod` pins a published `webview2/vX.Y.Z` tag. Today `Release webview2` is **manual** (`workflow_dispatch` only), so webview2 source can gain an API that `v3` cannot use until someone remembers to cut a release.

This bit #5616: it called `edge.Chromium.IsReady()`, which was added to webview2 source in #5568 but never released — so it compiled locally (via `go.work`'s local webview2) but failed cross-compile against the pinned published `v1.0.24`.

## Change

Add a `push` trigger to the existing `Release webview2` workflow:
```yaml
push:
  branches: [master]
  paths: ['webview2/**']
```
Any webview2 source change on master now auto-cuts the next patch tag (build+test gated, as today), and the existing `bump-webview2-v3` workflow opens the `v3/go.mod` bump PR.

## Why it's loop-safe

- The release pushes a **tag** (`webview2/vX.Y.Z`), not a master commit — tag pushes don't match `branches: [master]`.
- The downstream bump only edits `v3/**`, so merging it can't re-trigger this (`paths: webview2/**`).
- Added a `concurrency` group so two webview2 changes landing close together can't race on the next-patch computation.

No change to the release/build/test/bump logic itself — only the trigger.

## Summary by CodeRabbit

* **Chores**
  * Improved release automation for webview2 updates by automatically creating patch releases when changes are merged.
  * Added workflow concurrency controls to serialize webview2 release runs and avoid race conditions.
  * Corrected the next-version calculation to reliably extract the numeric portion of the latest webview2 tag.


leaanthony:
Good catch @copilot — confirmed. The auto-increment derived `MAJOR` from `v1.0.24` as `v1` and then prefixed another `v`, producing `webview2/vv1.0.25`. Verified by running the snippet:

```
LAST_VERSION=${LAST_TAG#webview2/}   # v1.0.24  -> MAJOR=v1 -> vv1.0.25  (bug)
LAST_VERSION=${LAST_TAG#webview2/v}  # 1.0.24   -> MAJOR=1  -> v1.0.25   (fixed)
```

This path was latent while the workflow was manual (runs passed an explicit `version`), but it becomes the default once the push trigger fires — so it had to be fixed for this PR to work. Pushed the one-line fix (strip the leading `v` in the prefix removal).

PR #5688, opened 2026-06-28
feat(v3): macOS global shortcuts land on the labelled key (layout-aware)

taliesin-ai:
## Description

Follow-up to #5686. Makes macOS global shortcuts **keyboard-layout aware** for letter accelerators.

### Problem

Carbon's `RegisterEventHotKey` binds to a physical *virtual key code*, not a character. So `Register("Cmd+A")` bound to the QWERTY **position** of A regardless of the active layout — the key labelled **Q** on AZERTY. (#5686 documented this as a known limitation.)

### Fix

Resolve letter accelerators against the **active keyboard layout** using `UCKeyTranslate`: scan the layout for the key that produces the requested letter unshifted, and bind that key code. `Cmd+A` now lands on the key **labelled A** on AZERTY/QWERTZ/Dvorak. If the active layout cannot produce the letter (e.g. Cyrillic/Greek), it falls back to the ANSI/QWERTY position so registration still succeeds.

**Only letters are translated.** They are produced unshifted on every Latin layout and never appear on the numeric keypad, so the reverse lookup is unambiguous. Digits, punctuation and named keys stay positional — translating digits would mis-bind to the **numpad** on layouts whose number row is shifted (e.g. AZERTY's main-row `1` is `&` unshifted, while the numpad `1` is `1`).

### Verification

Proved the reverse lookup against the real installed layouts:

| Layout | `a` | `q` | `z` | `w` | `y` |
|---|---|---|---|---|---|
| US QWERTY | 0 | 12 | 6 | 13 | 16 |
| French AZERTY | **12** | **0** | **13** | **6** | 16 |
| German QWERTZ | 0 | 12 | **16** | 13 | **6** |
| Dvorak | 0 | 7 | 44 | 43 | 17 |

French swaps a↔q and z↔w; German swaps y↔z; Dvorak fully remaps — each matching the physical position of the labelled key. Combined with the end-to-end "registered key code fires" proof from #5686, the round trip holds: an AZERTY user pressing the key labelled A triggers `Cmd+A`.

Adds unit tests (`global_shortcut_darwin_test.go`), updates the docs macOS tab and the changelog. `gofmt` clean; full `pkg/application` test suite passes on macOS; Windows still cross-compiles.

### Scope / follow-up

The binding is resolved at **registration time**. Switching keyboard layout *while the app is running* keeps existing letter shortcuts on their original physical key. Re-registering on a live layout change (observing `kTISNotifySelectedKeyboardInputSourceChanged`) is a deliberate, separable follow-up — kept out of this PR to keep it tight and because it is the harder part to test.

## Summary by CodeRabbit

* **Bug Fixes**
  * macOS global shortcuts now respect the active keyboard layout for letter keys, so shortcuts like `Cmd+A` follow the labeled key on layouts such as AZERTY, QWERTZ, and Dvorak.
  * Shortcut handling now keeps the original mapping after registration, even if the keyboard layout changes while the app is running.

* **Documentation**
  * Updated the keyboard shortcuts documentation and changelog to reflect the new macOS behavior.

### DIFF
--- a/docs/src/content/docs/features/keyboard/global-shortcuts.mdx
+++ b/docs/src/content/docs/features/keyboard/global-shortcuts.mdx
@@ -118,7 +118,7 @@ Because of these differences, always check the error returned by `Register` and 
 
     Global shortcuts use the Carbon Event Manager's hot key API. This is the standard mechanism for system-wide hot keys on macOS and does not require Accessibility permission.
 
-    Hot keys are bound to physical key positions, so a shortcut maps to the key in the standard ANSI/QWERTY position on non-QWERTY layouts.
+    Letter shortcuts are layout aware. The Carbon API binds to physical key positions, so Wails translates a letter to the key that produces it on your active keyboard layout (via `UCKeyTranslate`). `Cmd+A` therefore lands on the key labelled **A** on AZERTY, QWERTZ, Dvorak and similar layouts, rather than the QWERTY position. Digits, punctuation and named keys (arrows, function keys, and so on) use their fixed physical position. The binding is resolved when the shortcut is registered; if you switch keyboard layout while the application is running, existing letter shortcuts keep the physical key they were originally bound to.
 
     :::caution[Hide shortcuts and `ApplicationShouldTerminateAfterLastWindowClosed`]
     `window.Hide()` uses `orderOut:` on macOS, which makes the window non-visible. AppKit treats the last non-visible window as closed, so if you set `Mac.ApplicationShouldTerminateAfterLastWindowClosed: true` and use a global shortcut to hide your only window, the application will quit instead of staying in the background. Leave that option unset (the default) when you rely on a hide/show hotkey, so the window can be hidden and summoned back later.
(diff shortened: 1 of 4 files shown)


taliesin-ai:
Ported in c3605b6ac1615113380dd31d525943001a50cf54 and pushed fast-forward to this PR’s original branch as `taliesin-ai`.

Depends on #6137: merged its pushed head `7b0f6dc094bfe7bf693df0320338a9bfc5f86ec8` without rewriting history. Base is now `codex/mpress-markdown`; land after the migration.

Ported `features/keyboard/global-shortcuts.md` and all nine translations, retaining the original layout-aware explanation, existing hide-window caution, and this PR’s changelog entry. Only documentation/changelog conflicts needed resolution; the implementation patch is unchanged (matching stable patch ID).

Validation: pinned M-Press 1.0.17 strict build (`--strict --no-purge-css`), `mpress check`, and `check_site.py` all passed (2,010 built pages; rendered validator: 2,011 HTML pages, zero errors). Focused checks verified new content in all ten rendered language pages and the syntax of affected examples. Diff whitespace check passed; v2 website, translation sidecars, and audit gating match the migration base.

Translation audit remains pending the separately owned global parser/state migration. CodeRabbit was attempted before commit: `--plain` unsupported; one narrow `review --uncommitted --light --dir` fallback hit the shared OSS quota. No credits spent. Manual review used the diff against #6137, separating inherited migration content; no CodeRabbit findings were returned.

Native Linux `go test ./pkg/application -run '^TestGlobalShortcut' -count=1` passed. Go source/example syntax passed. macOS layout tests could not run on Linux. A server-tag test attempt failed on pre-existing `webview_window_linux_test.go` references to `linuxWindowState`, which server builds exclude.

Issue #5705, opened 2026-06-30
WebView2 controller permanently breaks on cross-GPU monitor switch (dual-GPU laptop)


qq540491950:
## Environment
- Wails v3.0.0-alpha2.109
- Windows 11 Pro 10.0.26200
- WebView2 Runtime 149.0.4022.98
- Dual GPU laptop: Intel iGPU (internal screen 1920x1080) + NVIDIA dGPU (external screen 1707x1067)
- Both monitors 100% DPI (96 DPI)
- Go 1.25.1

## Symptoms
1. Window displays correctly on startup screen
2. Drag window to the other screen → window content turns permanently black
3. Console repeatedly outputs: `[WebView2] Eval failed: The group or resource is not in the correct state to perform the requested operation.`
4. Drag window back to the original screen → stays black (controller permanently broken, not transient)
5. The Go process survives, but frontend-backend communication is dead

## What has been ruled out (all tried, none helped)
- `--disable-gpu`
- `--disable-gpu` + `--disable-direct-composition`
- `--disable-gpu` + `--no-sandbox`
- System-level `WEBVIEW2_ADDITIONAL_BROWSER_ARGUMENTS=--disable-gpu-compositing`
- Windows Graphics Settings forcing the app to use dGPU only
- Clearing EBWebView user data folder (`%APPDATA%\training-desktop.exe\EBWebView`)

## Root cause analysis

The error originates at `Eval()` → `ExecuteScript` in `chromium.go:287-295`. The existing code comment acknowledges this error can occur during "DPI transition between mixed-DPI monitors" (issue #5544) and treats it as transient (just logging it). But on dual-GPU systems the controller **never recovers**.

The deeper cause: Wails v3 explicitly disables WebView2's automatic monitor change detection at `chromium.go:358-359`:

```go
// Disable monitor scale changes since we are using raw pixels
if err := controller3.PutShouldDetectMonitorScaleChanges(false); err != nil {
    e.errorCallback(err)
}
```

This makes Wails responsible for handling monitor transitions. However, `webview_window_windows.go` has **no `WM_DISPLAYCHANGE` handler**. When the window moves between monitors attached to different GPU adapters, the CoreWebView2 rendering surface becomes invalid. Wails does not notify Chromium to rebuild it, and the controller enters a zombie state — passing nil checks but returning `RESOURCE_NOT_IN_CORRECT_STATE` on every subsequent `ExecuteScript` call.

## Proposed fix direction
[...]

Issue #5677, opened 2026-06-26
[Windows] WebView2 content shrinks and disappears after dragging window across mixed-DPI monitors (WM_DPICHANGED path missing chromium.Resize())


eleclin:
### Description

After upgrading to webview2 v1.0.27 / wails v3.0.0-alpha2.106, the os.Exit(1) crash from #5650 is fixed. However, a related visual artifact remains: when dragging a window from a monitor with one DPI scaling to another with different scaling, the WebView2 content shrinks, then disappears entirely (only the window background colour remains). The content does not recover until the window is restarted.
This only happens on the cross-DPI direction (e.g. 100% → 150%). Same-DPI monitor dragging is fine.

### To Reproduce

1. Run a Wails v3 app on Windows with two monitors with different DPI scaling (e.g. primary 1920×1080 @ 100%, secondary 1920×1080 @ 150%)
2. Drag the window from the 100% monitor to the 150% monitor — no issue
3. Drag the window back from the 150% monitor to the 100% monitor:
- Window freezes for ~3 seconds (WebView2 controller transitioning)
- Content briefly appears shrunk (rendered in a small area, window background fills the rest)
- ~1 second later, content disappears entirely
- Only the window's BackgroundColour (dark blue in screenshot) remains until app restart

### Expected behaviour

Expected behavior
Window content renders correctly at the new DPI after dragging across monitors.
Actual behavior
Content shrinks then disappears permanently. No crash (that part is fixed), but the app becomes unusable without restart.

### Attempted Fixes

Root cause analysis
The WM_DPICHANGED handler in pkg/application/webview_window_windows.go updates the rasterization scale but does not re-assert the WebView2 bounds:
// webview_window_windows.go, WM_DPICHANGED handler (~line 1812)
if !w.isMinimizing {
    w.resyncWebviewRasterizationScale()  // updates PutRasterizationScale
    // ← MISSING: w.chromium.Resize()
    if dpi, _ := w.DPI(); dpi != 0 {
        w.lastKnownDPI = dpi
    }
}
The Wails codebase already documents why this is insufficient — the comment on resyncWebviewDPIAfterMinimise() (same file, ~line 1986) explicitly states:
// a scale re-put alone does not make WebView2 re-lay out content whose bounds are
// unchanged — the page keeps reporting sizes computed with the stale scale
// until the bounds are re-asserted (#5544, reporter verification round 2).
func (w *windowsWebviewWindow) resyncWebviewDPIAfterMinimise() {
[...]

Issue #5605, opened 2026-06-14
[v3] App crashes when reopened after being minimized for 10~15 seconds on Windows


Bare7a:
### Description

When the user minimizes the application and then attempts to reopen it after approximately 10–15 seconds the application crashes.

### To Reproduce

1. Launch the application.
2. Use the app normally.
3. Minimize or send the app to the background.
4. Wait approximately 10 - 15 seconds.
5. Reopen the application.

### Expected behaviour

The application should resume normally without crashing

### Screenshots

Here are two videos: 

Alpha V100 - Where the old bug with the white background is still present but doesn't have the crash on resume
[<video src="https://github.com/user-attachments/assets/bfe51f86-5868-4b61-80fa-8f1fd0422d11" controls width="100%"></video>](https://github.com/user-attachments/assets/bfe51f86-5868-4b61-80fa-8f1fd0422d11)

Alpha v102 - Where the bug with the white background is not present, but the app crashes on resume
[<video src="https://github.com/user-attachments/assets/09b777fd-052a-40e5-bdeb-59e2dc9ec3c9" controls width="100%"></video>](https://github.com/user-attachments/assets/09b777fd-052a-40e5-bdeb-59e2dc9ec3c9)

### System Details

```shell
# System
        
┌─────────────────────────────────────────────────────────────────────┐
| Name              | Windows 10 Pro                                  |
| Version           | 2009 (Build: 26200)                             |
| ID                | 25H2                                            |
| Branding          | Windows 11 Pro                                  |
| Platform          | windows                                         |
| Architecture      | amd64                                           |
| Go WebView2Loader | true                                            |
| WebView2 Version  | 149.0.4022.69                                   |
| CPU               | AMD Ryzen 7 5800H with Radeon Graphics          |
| GPU               | NVIDIA GeForce RTX 3060 Laptop GPU (NVIDIA)     |
| Memory            | 32GB                                            |
└─────────────────────────────────────────────────────────────────────┘
                   
# Build Environment
                   
┌──────────────────────────────────────────────────────────────────────┐
| Wails CLI      | v3.0.0-alpha.102                                    |
| Go Version     | go1.26.1                                            |
| -buildmode     | exe                                                 |
| -compiler      | gc                                                  |
| CGO_CFLAGS     |                                                     |
| CGO_CPPFLAGS   |                                                     |
| CGO_CXXFLAGS   |                                                     |
| CGO_ENABLED    | 1                                                   |
[...]

PR #5710, opened 2026-07-01
fix(v3/windows): revalidate WebView2 surface on cross-GPU monitor change (#5705)

taliesin-ai:
Draft — addresses #5705. **I cannot test this on real dual-GPU hardware, so the behavioural fix is unverified. Details and an explicit ask for verification are below.**

## What the issue reports

On a dual-GPU laptop (Intel iGPU internal panel + NVIDIA dGPU external panel), dragging a Wails v3 window from one monitor to the other turns the WebView2 content permanently black. The Go process survives, but every subsequent `ExecuteScript` returns `RESOURCE_NOT_IN_CORRECT_STATE` ("The group or resource is not in the correct state…"), so the frontend is dead until restart. Dragging back does not recover it.

## Diagnosis

I traced this against the existing DPI-transition handling and I think the issue is a genuine **gap**, not a regression of the recent fixes:

- Wails runs WebView2 in raw-pixels bounds mode with `PutShouldDetectMonitorScaleChanges(false)` (`webview2/pkg/edge/chromium.go`), which makes Wails responsible for reacting to monitor transitions.
- The recent mixed-DPI work (#5544, #5605, #5677) handles the case where the DPI *changes* across the move: `WM_DPICHANGED` fires, the rasterization scale is re-asserted, and the bounds are re-asserted so content re-lays-out.
- **But the reporter's two monitors are both at 100% (96 DPI).** With no DPI change, `WM_DPICHANGED` never fires and `resyncWebviewRasterizationScale()` is a no-op — so there is currently **no** code path that reacts to the window changing monitors at equal DPI. That is exactly the scenario where the composition surface, tied to the *source* monitor's GPU adapter, is left invalid.

Two honest caveats on the diagnosis:

1. A commenter reports the opposite (crash only with a DPI *mismatch*, fine at equal DPI). I read that as evidence the underlying trigger is **GPU-adapter surface loss**, which is largely orthogonal to DPI — so I've deliberately *not* framed this as "mixed-DPI is already solved". Both triggers plausibly share the same adapter-change root cause.
2. This may ultimately be a WebView2 **runtime** device-loss bug (dual-GPU black screens are a known class in the WebView2 backlog). A Wails-side workaround may only partially recover it.

## On the issue's suggested fix

The issue proposes a `WM_DISPLAYCHANGE` handler. `WM_DISPLAYCHANGE` is only broadcast on a resolution/topology change — it does **not** reliably fire when a window is *dragged* between monitors. So the robust detector for the drag case is tracking the window's `HMONITOR` across `WM_WINDOWPOSCHANGED` (which does fire during the drag). I've implemented that as the primary path, and *also* kept a `WM_DISPLAYCHANGE` handler because on some hybrid-GPU laptops activating the dGPU for the external panel does cause a topology change.

## What this PR does

1. Track the window's current monitor handle; seed it at creation so only genuine crossings trigger work.
2. On `WM_WINDOWPOSCHANGED`, when the handle changes (skipped while minimised, matching the #5605 reasoning), run a revalidation. Also run it on `WM_DISPLAYCHANGE`.
3. The revalidation: `NotifyParentWindowPositionChanged` + rasterization-scale resync (no-op at equal DPI) + a **bounds re-assert that actually produces a delta**.

## The part I most want review on

The revalidation **action** is the discriminating variable I can't test. Re-asserting the *same* bounds is a no-op (WebView2 early-outs on an unchanged rect — this is also why #5677 only re-asserts bounds *after* a scale change made layout stale). So I nudge the bounds by 1px and restore them to force a real re-layout / surface revalidation. That's the least-risky action that might actually do something, but I genuinely don't know whether it's enough to rebuild a composition surface stranded on a dead adapter.

Escalations, in increasing order of effectiveness **and** regression risk, that a hardware owner could try by editing `revalidateWebviewForNewMonitor`:

- **(this PR)** 1px bounds nudge + restore — invisible, may not force a device rebuild.
- **Visibility toggle** — `controller.PutIsVisible(false)` then `true`. More likely to force the compositor to re-attach, but risks a visible flash on *every* monitor crossing for *all* users.
- **Controller recreation** — near-certain to recover, but heavy and loses page state.

## Testing

- Cross-compiles clean for `windows/amd64` and `windows/arm64`; `go vet` shows only pre-existing `unsafe.Pointer` notes. **Behaviour is unverified — I have no dual-GPU Windows machine.**

## Ask

@lirt (#5705 reporter) and @qq540491950 — could you build this branch and confirm whether the 1px-nudge recovery clears the black screen on your hardware? If it doesn't, trying the visibility toggle (one-line change, noted in the code comment) would tell us whether a stronger surface reset is required. Maintainers: happy to reshape the recovery action or gate the visibility toggle behind an opt-in if that's the direction preferred.

### DIFF
--- a/v3/UNRELEASED_CHANGELOG.md
+++ b/v3/UNRELEASED_CHANGELOG.md
@@ -27,6 +27,7 @@ After processing, the content will be moved to the main changelog and this file 
 
 ## Fixed
 <!-- Bug fixes -->
+- Revalidate the WebView2 surface when the window crosses onto a monitor driven by a different GPU adapter, to address the permanent black screen on dual-GPU laptops (#5705) in [PR](https://github.com/wailsapp/wails/pull/5710) by @taliesin-ai
 
 ## Deprecated
 <!-- Soon-to-be removed features -->
--- a/v3/pkg/application/webview_window_windows.go
+++ b/v3/pkg/application/webview_window_windows.go
@@ -93,6 +93,15 @@ type windowsWebviewWindow struct {
 	// window is repositioned off the monitor it will restore to).
 	lastKnownDPI w32.UINT
 
+	// currentMonitor is the handle of the monitor the window was last seen on
+	// while non-minimised. It is used to detect when the window crosses onto a
+	// different monitor (potentially driven by a different GPU adapter), which
+	// on some systems invalidates the WebView2 composition surface and needs a
+	// manual revalidation (#5705). It is only sampled while non-minimised: while
+	// minimised the window is parked off its restore monitor, so the handle
+	// flips spuriously (same reasoning as lastKnownDPI, #5605).
+	currentMonitor w32.HMONITOR
+
 	// menubarTheme is the theme for the menubar
 	menubarTheme *w32.MenuBarTheme
 
@@ -467,6 +476,11 @@ func (w *windowsWebviewWindow) run() {
 	if dpi, _ := w.DPI(); dpi != 0 {
 		w.lastKnownDPI = dpi
 	}
+
+	// Seed the current monitor so the first genuine monitor crossing (not the
+	// window's initial placement) is what triggers the WebView2 revalidation
+	// in the WM_WINDOWPOSCHANGED handler (#5705).
+	w.currentMonitor = w32.MonitorFromWindow(w.hwnd, w32.MONITOR_DEFAULTTONEAREST)
 
 	// Ensure correct window size in case the scale factor of current screen is different from the initial one.
 	// This could happen when using the default window position and the window launches on a secondary monitor.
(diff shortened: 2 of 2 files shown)


taliesin-ai:
Closing this draft: the approach is superseded and the fix did not survive hardware testing.

1. @qq540491950 built this branch (commit 1a94d00) on real dual-GPU hardware and the black screen still reproduced on every cross-monitor drag, so the 1px bounds nudge is not sufficient to recover the surface.
2. The field data in #5705 now points to a DPI mismatch during the drag as the trigger, not a GPU-adapter handoff: equal-DPI cross-GPU drags work fine, and the crash only appears when scaling differs. That is exactly the root cause #5734 fixes (host-managed rasterization scale racing the browser process into a degenerate scale(0,0) transform).
3. Mechanically, this PR's revalidation calls resyncWebviewRasterizationScale(), the host-side scale write that #5734 deliberately stands down while native scale detection is on, so the two cannot merge together.

Testers on #5705 should verify against #5734 instead.

Issue #5631, opened 2026-06-19
Linux: repeated SIGABRT/SIGSEGV in WebKitGTK/JSC main loop on v3.0.0-alpha2.103


jalfdev:
### Description

We are seeing repeated native crashes on Linux after upgrading to Wails `v3.0.0-alpha2.103`.

This is a long-running Wails desktop application on Linux using WebKitGTK. The app updates the UI frequently and loads local media/resources through the WebView. After the process crashes, an external watchdog restarts it.

We previously reported related Linux issues in:

- https://github.com/wailsapp/wails/issues/5557
- https://github.com/wailsapp/wails/issues/5552

After upgrading to `v3.0.0-alpha2.103` and removing local Wails source patches, the crashes still happen.

### Environment

- Wails: `v3.0.0-alpha2.103`
- Go: `go1.25.5 linux/amd64`
- GTK: `3.24.41`
- WebKitGTK: `2.50.4`
- GLib: `2.80.1`
- OS: Linux x86_64

### Crash frequency

After deployment, the watchdog recorded 10 crashes within about 15 hours.

All exits were:

```text
signal: aborted (core dumped)
```

### Main crash signature

9 out of 10 crashes have the same signature.

stderr:

```text
SIGSEGV: segmentation violation
PC=0x... sigcode=1 addr=0x28
signal arrived during cgo execution

goroutine 1 [syscall, locked to thread]:
runtime.cgocall(...)
github.com/wailsapp/wails/v3/pkg/application._Cfunc_g_application_run(...)
github.com/wailsapp/wails/v3/pkg/application.appRun(...)
    github.com/wailsapp/wails/v3@v3.0.0-alpha2.103/pkg/application/linux_cgo_gtk3.go:639
github.com/wailsapp/wails/v3/pkg/application.(*linuxApp).run(...)
github.com/wailsapp/wails/v3/pkg/application.(*App).Run(...)
main.main()
```

Native coredump stack example:

```text
Signal: 6 (ABRT)

#9  libjavascriptcoregtk-4.1.so.0 + 0x1950c99
#11 libwebkit2gtk-4.1.so.0 + 0x99295e
#12 libwebkit2gtk-4.1.so.0 + 0x30f9791
#13 libwebkit2gtk-4.1.so.0 + 0xde82cd
#14 libwebkit2gtk-4.1.so.0 + 0xde83fd
#15 libwebkit2gtk-4.1.so.0 + 0xe3e701
#16 g_object_unref
[...]

PR #5715, opened 2026-07-03
feat(linux): CGO-free Linux backend via purego (GTK4 + WebKitGTK 6.0)

taliesin-ai:
## What

Revives CGO-free Linux support (removed as unmaintained in #5661), rebuilt from scratch against the current GTK4 + WebKitGTK 6.0 cgo backend. Build with:

```sh
CGO_ENABLED=0 go build -tags purego ./...
```

The `purego` tag selects the new backend; the default cgo backend (and the gtk3 variant) are unchanged and mutually exclusive via `!purego` tags. Cross-compiling Linux GUI apps from any host now works, since no C toolchain is involved.

## Design

- **`linux_purego_lib.go`** — dlopen layer with per-distro soname fallbacks and one aggregate, actionable error when libraries/symbols are missing (package names per distro instead of a linker one-liner). Full typed function registry for GLib/GObject/Gio/GTK4/WebKitGTK-6.0/JSC/libsoup.
- **Runtime capability guards replace compile-time gates**: version-dependent symbols resolve optionally and are nil-checked — e.g. `gdk_monitor_get_scale` (GTK 4.14+) falls back to `gdk_monitor_get_scale_factor` on older GTK4, where the cgo build won't even start; GDK X11 symbols are optional so Wayland-only GTK builds work.
- **`linux_purego_callbacks.go`** — pure-Go port of `linux_cgo.c`: a fixed set of `purego.NewCallback` trampolines (never per-window/item), GAction menus, GTK4 dialogs, DND, clipboard, X11 helpers via `dlsym(RTLD_DEFAULT)`, and the #5527 SA_ONSTACK signal fix (incl. the SIGUSR1/JSC exemption) via libc `sigaction`.
- **`internal/assetserver/webview/*_linux_purego.go`** — preserves the #5631/#5668 main-thread confinement design exactly (atomic enabled-check + schedule, inline-after-shutdown).
- The dbus theme monitor, portal global shortcuts, permissions helpers and the system tray were already pure Go and are now shared via `(cgo || purego)` tags. X11 global shortcuts (XGrabKey) ported against dlopen'd libX11.
- **Not bug-for-bug**: defects found in the cgo backend during translation are fixed in the port and catalogued in `v3/pkg/application/BUGS_FOUND.md` (invalid free of a GLib-owned string, NULL GList deref, clipboard reentrancy, unsynchronised menu maps, >100-file dialog deadlock, message-dialog zombie/UAF, vestigial `import "C"` that made the tray cgo-only). Design notes: `v3/pkg/application/PUREGO_LINUX.md`.

## Verified

- `GOOS=linux GOARCH=amd64|arm64 CGO_ENABLED=0 go build -tags purego ./...` (also `purego,production` / `purego,devtools`), vet clean apart from the standard purego unsafeptr idiom notices.
- On Ubuntu 26.04 (GTK 4.22.2, WebKitGTK 2.52.3): `examples/window` built CGO-free **runs on a real X server** — native window + GTK4 menu bar render, popover menus with registered accelerators work, a menu action creates a second window, the `wails://` scheme serves assets, and the JS→Go bridge round-trips (clicking the page's `setPos` moved the native window via XMoveWindow, verified with xdotool). Main-window close exits 0.
- Regression checks on the same box: default cgo build + tests pass, gtk3 cgo variant builds, purego `go test ./pkg/application` passes.
- Known environmental issue documented in BUGS_FOUND.md: closing a child window on DRI-less Xvfb crashes with an X `BadDrawable` — reproduced byte-identically with the **cgo** binary, so it's upstream WebKitGTK software-rendering teardown, not a port regression.

### DIFF
--- a/v3/go.mod
+++ b/v3/go.mod
@@ -12,6 +12,7 @@ require (
 	github.com/charmbracelet/glamour v0.10.0
 	github.com/charmbracelet/huh v0.8.0
 	github.com/coder/websocket v1.8.14
+	github.com/ebitengine/purego v0.10.1
 	github.com/go-json-experiment/json v0.0.0-20251027170946-4849db3c2f7e
 	github.com/go-ole/go-ole v1.3.0
 	github.com/godbus/dbus/v5 v5.2.2
--- a/v3/go.sum
+++ b/v3/go.sum
@@ -139,6 +139,8 @@ github.com/dominikbraun/graph v0.23.0 h1:TdZB4pPqCLFxYhdyMFb1TBdFxp8XLcJfTTBQucV
 github.com/dominikbraun/graph v0.23.0/go.mod h1:yOjYyogZLY1LSG9E33JWZJiq5k83Qy2C6POAuiViluc=
 github.com/dustin/go-humanize v1.0.1 h1:GzkhY7T5VNhEkwH0PVJgjz+fX1rhBrR7pRT3mDkpeCY=
 github.com/dustin/go-humanize v1.0.1/go.mod h1:Mu1zIs6XwVuF/gI1OepvI0qD18qycQx+mFykh5fBlto=
+github.com/ebitengine/purego v0.10.1 h1:dewVBCBT2GaMu1SrNTYxQhgQBethzfhiwvZiLGP/qyY=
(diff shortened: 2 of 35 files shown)


taliesin-ai:
Follow-up: the 8 cgo-backend defects catalogued in `BUGS_FOUND.md` are now fixed in the cgo backend itself in #5718.

Issue #5733, opened 2026-07-03
[v3][Windows] WebView2 ProcessFailed is unhandled — browser-process death leaves the window permanently blank until app restart


randalmurphal:
# Description

Wails v3 on Windows never registers WebView2's `ProcessFailed` event. When the WebView2 **browser process** dies — for any reason — the controller is left in a permanently invalid state: every subsequent COM call returns `ERROR_INVALID_STATE (0x8007139F)`, the window renders blank, and only restarting the host app recovers. The app process itself stays alive, so the user is left staring at a dead white/blank window with no indication of what happened.

Concrete triggers:

- The mixed-DPI GPU-kill loop in #5732 (reproducible chain: degenerate transform → repeated GPU process kills → `FATAL ... GPU process isn't usable. Goodbye.` → browser process exits).
- Killing `msedgewebview2.exe` from Task Manager (trivial repro of the same end state).
- Any WebView2 runtime crash or out-of-resource kill.

Renderer-level failures (`RENDER_PROCESS_EXITED` / `RENDER_PROCESS_UNRESPONSIVE`) similarly leave a dead view where a simple re-navigation would recover.

Observable aftermath in a host log once the browser process is gone (every focus/resize event storms these):

```
[WebView2] Focus failed: The group or resource is not in the correct state to perform the requested operation.
[WebView2] SetSize failed: The group or resource is not in the correct state to perform the requested operation.
```

This is closely related to #5580's audit (transient COM errors exiting the process via `errorCallback`): alpha2.112 stopped the *crash*, but what remains is a window that can never recover. Handling `ProcessFailed` is the recovery half of that story.

# Fix, validated

Branch: https://github.com/randalmurphal/wails/tree/fix/webview2-process-failed-recovery (single commit off the v3.0.0-alpha2.112 tag).

- Register `chromium.ProcessFailedCallback` in `setupChromium`.
- `RENDER_PROCESS_EXITED` / `RENDER_PROCESS_UNRESPONSIVE` → re-navigate to the last host-set URL (tracked in `setURL`; the dead webview can't be queried for it).
- `BROWSER_PROCESS_EXITED` → rebuild: fresh `edge.NewChromium()`, re-run `setupChromium`, re-navigate. The old instance is abandoned rather than re-embedded — after a browser-process exit all its COM references dangle, and `Embed`'s init-wait loop keys on a per-instance flag a used instance has already set, so re-embedding would return before a new controller exists.
[...]

PR #5734, opened 2026-07-03
[v3][windows] Re-enable WebView2 monitor-scale detection to fix mixed-DPI GPU-process crash (#5732)

### LINKED ISSUE #5701: [Windows] WebView2 content disappears on mixed-DPI drag persists after #5689 — render process terminates on scale-up
First, thank you for #5689 — it successfully resolved the `os.Exit(1)` crash from #5650, and we appreciate the quick fix.

We'd like to share follow-up findings from testing #5689 on real mixed-DPI hardware. The crash is gone, but a visual artifact remains: after dragging a window across a mixed-DPI monitor boundary (scale-up direction, e.g. 100% → 150%), WebView2 content shrinks then disappears permanently until app restart.

We've done extensive diagnosis and believe the root cause is deeper than the missing `Resize()` that #5689 addressed. We'd like to share what we found and ask whether exposing `ShouldDetectMonitorScaleChanges` as a per-window option might be a feasible direction.

**To Reproduce**

1. Wails v3 app on Windows, two monitors with different DPI (primary 1920×1080 @ 100%, secondary 1920×1080 @ 150%)
2. Drag window from 100% monitor to 150% monitor
3. Window freezes ~3 seconds, content briefly appears shrunk, then disappears entirely
4. Only the window background colour remains; app must be restarted
5. Reverse direction (150% → 100%) works fine

**What we observed**

We added debug logging to `resyncWebviewRasterizationScale` and the `WM_DPICHANGED` handler. On scale-up (100% → 150%):

```
WM_DPICHANGED enter dpi=144
  resync: dpiX=144 computedScale=1.5000 currentScale=1.0000   ← controller alive
  resync: PutRasterizationScale(1.5000) succeeded               ← COM returns S_OK
  resync returned true
  chromium.Resize() called                                      ← #5689 fix executes
WM_DPICHANGED exit

# 3 seconds later:
  resync: dpiX=144 computedScale=1.5000 currentScale=0.0000    ← unexpected: scale is 0
  resync: PutRasterizationScale(1.5000) succeeded               ← COM S_OK but no effect
  [WebView2] Eval failed: ERROR_INVALID_STATE  (continuous 10+ seconds)
  [WebView2] SetSize failed: ERROR_INVALID_STATE
```

The `currentScale=0.0000` value appears to indicate the render process has terminated and failed to recreate. After that point, all COM calls (`PutRasterizationScale`, `PutBounds`, `Resize`, even `Navigate`) return S_OK but have no effect — the COM proxy seems to be talking to a dead render process.

Scale-down (150% → 100%) does not exhibit this: `currentScale` stays at valid values throughout.

**What we tried**

We explored several recovery approaches in the `WM_DPICHANGED` handler, but none could revive the render process once `currentScale` reached 0:

| Approach | Outcome |
|----------|---------|
| `chromium.Resize()` after resync (#5689) | `PutBounds` also fails with ERROR_INVALID_STATE |
| Unconditional `Resize()` (ignoring resync gate) | Same — `PutBounds` fails |
| Deferred retry with backoff (7 attempts, 500ms–15s) | `currentScale` stays 0 across all retries |
| `NotifyParentWindowPositionChanged()` | No effect |
| `Navigate()` to force render process recreation | No effect — COM proxy appears non-functional |

Our conclusion: once the render process terminates on scale-up, application-level COM calls cannot recover it. The issue needs to be prevented rather than recovered from.

**What resolved it for us**

We noticed `PutShouldDetectMonitorScaleChanges(false)` is set in `chromium.go` (line ~359), which we understand is intentional for the raw-pixels bounds mode. Out of curiosity, we tried enabling it (`true`) to let WebView2 handle DPI transitions natively, and it resolved the issue completely:

```go
controller3.PutShouldDetectMonitorScaleChanges(true)
```

With native DPI detection enabled, the scale-up log looks like this — no `currentScale=0`, no errors:

```
WM_DPICHANGED enter dpi=144
  resync: dpiX=144 computedScale=1.5000 currentScale=1.0000   ← alive
  resync: PutRasterizationScale(1.5000) succeeded
WM_DPICHANGED exit
# No errors, content flashes once and recovers
```

Both drag directions now work correctly on our hardware.

To be transparent: we're currently running with a local patch
(`go.mod` replace directive) that enables this flag, which fully
resolves the issue for our use case. We're sharing these findings
in case they're useful — there's no urgency on our end, and we
respect that the maintainers may have considerations we're not
aware of.

**A question about direction**

We understand the raw-pixels-bounds-mode rationale for disabling native DPI detection by default, and we're not suggesting that default be changed. However, for developers on mixed-DPI multi-monitor setups, this behavior makes the app unusable without a workaround.

Would exposing `ShouldDetectMonitorScaleChanges` as a `WebviewWindowOptions` field (e.g. `EnableNativeMonitorScaleDetection bool // default false`) be a feasible approach? This would:

- Preserve the existing default behavior and performance characteristics
- Add zero regression risk (opt-in only)
- Let affected developers self-resolve without maintaining a fork

We'd be happy to open a PR for this if the direction seems reasonable.

**Environment**

- Wails: v3.0.0-alpha2.109
- webview2 wrapper: `github.com/wailsapp/wails/webview2@v1.0.27`
- OS: Windows 11 (Build 26200)
- Monitors: 27" 1920×1080 @ 100% + 15" 1920×1080 @ 150%, extended display
- Go: 1.26.2

**Related issues**

- #5677 — original report, closed by #5689
- #5689 — added `chromium.Resize()` to `WM_DPICHANGED` handler; resolves the crash but the visual artifact persists on mixed-DPI hardware
- #5650 — crash on this code path, fixed by webview2 v1.0.26
- #5544 — un-minimise variant; `Resize()` works there because the render process stays alive

### LINKED ISSUE #5732: [v3][Windows] Mixed-DPI monitor drag kills WebView2 GPU process repeatedly until the browser process exits — window permanently blank
# Description

On a mixed-DPI multi-monitor setup, dragging a Wails v3 window from an external monitor onto the (different scale factor) laptop display makes the embedded WebView2 browser process compute a degenerate `scale(0,0)` transform. The viz compositor rejects the resulting frame and kills the GPU process; the kill repeats on every submitted frame until Chromium's crash limit is hit, at which point the **browser process itself exits** and the window is left permanently blank — every subsequent controller COM call fails with `ERROR_INVALID_STATE (0x8007139F)` until the app is restarted.

(Before alpha2.112 this was a hard host-process crash rather than a blank window, because transient COM errors used to reach `errorCallback` → `os.Exit(1)` — the path audited in #5580. alpha2.112 made it survivable; this issue is about the underlying WebView2 death.)

# Steps to reproduce

1. Windows machine with monitors at different scale factors (repro box: laptop display + 3 external monitors, mixed scaling, Intel iGPU).
2. Run any Wails v3 (v3.0.0-alpha2.112) app.
3. Drag the window across the DPI boundary onto the differently-scaled display, possibly a few times.
4. Window renders for a moment, blanks, and never recovers (pre-alpha2.112: host app crashes).

Reproduces reliably on the affected machine; needs the mixed-DPI hardware condition.

# Evidence (`chrome_debug.log` from the WebView2 profile, `--enable-logging --v=1`)

The full chain from the dying session:

```
[25604:7252:0703/132849.877:ERROR:ui\gfx\geometry\transform.cc:505] NOTREACHED hit. [ 0 0 0 0
  0 0 0 0
  0 0 1 0
  0 0 0 1 ]
 is not invertible
[25604:7252:0703/132851.029:ERROR:content\browser\gpu\gpu_process_host.cc:1063] GPU process exited unexpectedly: exit_code=-2147483645
[25604:7252:0703/132900.954:ERROR:ui\gfx\geometry\transform.cc:505] NOTREACHED hit. [ 0 0 0 0 ...] is not invertible
[25604:7252:0703/132901.656:ERROR:content\browser\gpu\gpu_process_host.cc:1063] GPU process exited unexpectedly: exit_code=-2147483645
[25604:7252:0703/132905.150:ERROR:ui\gfx\geometry\transform.cc:505] NOTREACHED hit. [ 0 0 0 0 ...] is not invertible
[25604:7252:0703/132905.799:ERROR:content\browser\gpu\gpu_process_host.cc:1063] GPU process exited unexpectedly: exit_code=-2147483645
```

When the kills come fast enough, Chromium gives up and the browser process exits:

```
FATAL:content\browser\gpu\gpu_data_manager_impl_private.cc GPU process isn't usable. Goodbye.
```

`exit_code=-2147483645` is `0x80000003` (STATUS_BREAKPOINT — a CHECK kill, not a GPU driver fault). No Crashpad dumps are produced; the browser side treats it as an orderly shutdown.

# Analysis

wails runs WebView2 in the go-webview2 module's raw-pixels configuration: `PutBoundsMode(USE_RAW_PIXELS)` + `PutShouldDetectMonitorScaleChanges(false)`, with the host pushing `PutRasterizationScale` from its `WM_DPICHANGED` handler ([webview_window_windows.go WM_DPICHANGED path](https://github.com/wailsapp/wails/blob/v3.0.0-alpha2.112/v3/pkg/application/webview_window_windows.go#L1809)).

That host-managed mode is doc-legal, but it's a path almost no WebView2 embedder exercises (mainstream hosts run DIP bounds + detection on). During a monitor cross, the host's scale/bounds writes race the browser process's own internal display bookkeeping (which keeps running even with detection off), and the browser lands on a zero x/y scale when building the compositor transform. That looks like the same underlying race as #5677's shrink-then-disappear symptom — #5677's fix (re-asserting bounds after the scale re-put) treats one observable symptom of it.

# Fix, validated

Two changes, one behavioral unit — make Edge the *only* writer of the rasterization scale:

1. Re-enable `ShouldDetectMonitorScaleChanges` (the WebView2 default) right after `Embed`. Bounds stay raw-pixels.
2. Gate `resyncWebviewRasterizationScale` off while detection is on, so the host never writes the scale concurrently with Edge. (With both writers active, a transient NOTREACHED + one GPU kill per crossing still occurred — self-healing, but visible as a blank flash.)

Branch: https://github.com/randalmurphal/wails/tree/fix/webview2-mixed-dpi-rasterization-scale (single commit off the v3.0.0-alpha2.112 tag).

After this change the repro machine shows **zero** NOTREACHED / GPU-process exits across repeated boundary drags, with correct rendering at every scale factor.

Caveats stated up front, since this touches scar tissue: it flips a deliberate go-webview2 design choice, and it's validated on one machine / one WebView2 runtime (149.0.4022.98, Windows 11 26200). The #5544 / #5605 minimize-restore scenarios route around the now-dormant host resync and were not individually re-tested — with detection on, Edge's own (heavily exercised) monitor-change path handles them, and the host makes strictly fewer COM calls into possibly-suspended controllers, but maintainer judgment on that history is exactly why this is an issue with a linked branch rather than a direct PR.

Happy to submit the branch as a PR if this direction looks right to you.

# Environment

- Wails: v3.0.0-alpha2.112 (tagged)
- OS: Windows 11 Enterprise 25H2 (build 26200)
- WebView2 Runtime: 149.0.4022.98
- Hardware: laptop (Intel iGPU) + 3 external monitors, mixed scale factors
- Related: #5733 — the missing `ProcessFailed` handling that turns any browser-process death into a permanent blank window

taliesin-ai:
Fixes #5732
Fixes #5701 (same root cause; the two-part fix verified independently by @eleclin in that thread is exactly what this PR implements)

## What this does

On mixed-DPI multi-monitor setups, dragging a v3 window across a scale-factor boundary can make the embedded WebView2 browser compute a degenerate `scale(0,0)` compositor transform. The viz compositor rejects the frame and kills the GPU process on every submission until Chromium hits its crash limit and the browser process exits, leaving the window permanently blank.

Root cause: wails ran WebView2 in a host-managed scale mode — `PutShouldDetectMonitorScaleChanges(false)` with the host pushing `PutRasterizationScale` from `WM_DPICHANGED`. During a monitor cross, the host's scale writes race the browser process's own display bookkeeping, which can land on a zero x/y scale. Survey of other embedders: **wry/Tauri, webview/webview, and upstream jchv/go-webview2 all leave the property at its default (detection on) and never write the rasterization scale** — the disable was a wails-fork addition, and Microsoft's docs reserve detection-off for apps that want a *custom* (non-monitor) scale.

Two commits, one behavioral unit — make Edge the only writer of the rasterization scale:

1. **App layer** (@randalmurphal's validated commit, cherry-picked with authorship preserved): re-enable `ShouldDetectMonitorScaleChanges` right after `Embed`, and stand the host-side `resyncWebviewRasterizationScale` down while detection is on (`monitorScaleDetectionOn` flag). This makes the fix effective against the published webview2 module `v1.0.27`, which still carries the disable. If the re-enable call fails, the flag stays unset and the previous host-managed behavior is preserved unchanged.
2. **Module layer**: remove the `PutShouldDetectMonitorScaleChanges(false)` call from `webview2/pkg/edge/chromium.go` so future module releases run the WebView2 default natively (matching upstream go-webview2). Once a fixed module is tagged and v3 bumps to it, the app-layer re-enable becomes a no-op re-assert of the default. Bounds mode stays raw pixels — orthogonal knob, and the documented default anyway.

The now-dormant host resync machinery (`resyncWebviewRasterizationScale` and the minimize-restore resync paths from #5544/#5605/#5677) is deliberately kept as the automatic fallback; it can be deleted once this has soaked for a release or two.

## Validation

- Reporter's repro hardware (Windows 11 26200, WebView2 149.0.4022.98, Intel iGPU + 3 mixed-scale monitors): zero NOTREACHED / GPU-process exits across repeated boundary drags, correct rendering at every scale factor (see issue for `chrome_debug.log` evidence).
- Independent mixed-DPI verification on a second machine (Windows 11 26200, WebView2 149.0.4022.98, AMD iGPU, dual displays at different scale factors): automated boundary-crossing + minimize/restore soak test against both this branch and the alpha2.112 baseline, comparing `devicePixelRatio` traces and `chrome_debug.log`. Results to be posted in this PR.
- `GOOS=windows go build` passes for both `v3/pkg/application` and the `webview2` module.

Related: #5733 (missing `ProcessFailed` handling) — separate issue, not addressed here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved display scaling behavior on Windows, especially when moving the app between monitors with different DPI settings.
  * Reduced the chance of visual scaling glitches and repeated crashes related to monitor-scale changes.
  * Kept window rendering in sync more reliably when screen scaling changes during use.


taliesin-ai:
## Independent validation results (second machine)

Ran an automated A/B soak test on a second mixed-DPI setup — different GPU vendor than the reporter's (AMD Radeon iGPU vs Intel; same Windows 11 build 26200 and WebView2 runtime 149.0.4022.98). Dual displays at 100% / 150%, WebView2 launched with `--enable-logging --v=1`.

**Test harness**: a self-driving Wails v3 app that performs 10 stepped (drag-like, 48px/16ms) window crossings of the DPI boundary, then 5 minimize/restore cycles split across both displays. The page reports `devicePixelRatio` + `screenX` to the Go side every 500ms. Run twice, identical results both rounds; behavior also observed live.

**Baseline (`v3.0.0-alpha2.112` + published webview2 v1.0.27):**
- Page telemetry dies ~10s in, at the first boundary crossing, after degenerate readings: `dpr=0.5`, `screenX=-2147483648` (INT_MIN), `innerWidth=0`.
- Window content goes permanently black (observed live + desktop screenshot).
- `chrome_debug.log` shows the exact chain from the issue: 12× `ui\gfx\geometry\transform.cc:505 NOTREACHED hit. [0 0 0 0 ...] is not invertible`, 6× `GPU process exited unexpectedly: exit_code=-2147483645`, ending in `FATAL:...gpu_data_manager_impl_private.cc GPU process isn't usable. Goodbye.`

**This branch:**
- Full telemetry for the entire run, both rounds (120 reports): clean `dpr` 1.0 ↔ 1.5 flips at the boundary on every crossing, correct rendering at both scales.
- Minimize/restore retained the correct scale on both displays (visible in telemetry as the parked-window coordinates `-32000` on the 100% display and `-32000/1.5 = -21333` on the 150% one) — i.e. the #5544 / #5605 restore scenarios behave under detection-on.
- `chrome_debug.log`: **zero** NOTREACHED / GPU-process errors across 347 log lines.

Together with the reporter's Intel-GPU validation, the fix now has clean results on two machines / two GPU vendors, and the failure signature is confirmed hardware-independent.

leaanthony:
Recording of the test!

https://github.com/user-attachments/assets/f034ff2b-d189-466a-b971-0785baf81267

leaanthony (approved):

PR #5742, opened 2026-07-04
fix(windows): hide the WebView2 controller on minimise/hide

wayneforrest:
# Description

Under `UseVisualHosting` , show() sets the `WebView2` controller visible but `hide()` and minimise never hid it again. That leaves a live input surface where the window was, which silently **swallows mouse clicks** (an invisible "dead zone").

Application Startup Options:
**Env Var** : `COREWEBVIEW2_FORCED_HOSTING_MODE=WINDOW_TO_VISUAL`
```
Windows: application.WindowsOptions {
      UseVisualHosting: true,
}
```

## Type of change
  
# How Has This Been Tested?
  
- [x] Windows
- [ ] macOS
- [ ] Linux
      
## Test Configuration

# Checklist:

- [ ] I have performed a self-review of my own code

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows WebView2 rendering during minimize and restore, reducing blank or “dead zone” hit-testing.
  * Enhanced synchronization between window visibility transitions and the embedded web content’s show/hide behavior to prevent stale or unresponsive states.


wayneforrest:
cc @leaanthony  @atterpac

v3/pkg/application/webview_window_windows.go
```diff
@@ -1284,6 +1284,14 @@ func (w *windowsWebviewWindow) hide() {
 	w.windowShown = false
 	w.showRequested = false
 
+	// Symmetric with show()'s chromium.Show(): under UseVisualHosting a bare
+	// SW_HIDE leaves the DirectComposition input surface hit-testing where the
+	// window was (a desktop right-click "dead zone"). Restore is always
+	// programmatic via Show() -> show() -> chromium.Show().
+	if w.chromium != nil {
+		_ = w.chromium.Hide()
+	}
```
atterpac: We likely need to guard all of these on `w.webviewNavigationCompleted` as well. `.Show()` does this the same but inverted up top we could also continue that pattern 

```go
    if w.chromium == nil && !w.webviewNavigationCompleted {
           return
    }
```

Might honestly be worth refactoring to a helper like

```go
func (w *windowsWebviewWindow) setChromiumVisible(visible bool) {
      if w.chromium == nil || !w.webviewNavigationCompleted {
              return
      }
      if visible {
              _ = w.chromium.Show()
      } else {
              _ = w.chromium.Hide()
      }
}
```

Then moving all these defensive checks to 

```go
// from
if w.chromium != nil && w.webviewNavigationCompleted {
    _ = w.chromium.Show()
}
//to 
w.setChromiumVisible(true)
```

leaanthony (changes requested):
The approach is right (hiding the controller on minimize is the WebView2-recommended behaviour and it removes the visual-hosting input dead zone), and CI is green. Two changes requested before merge:

1. **Nil-controller guard is insufficient.** `Chromium.Hide()`/`Show()` dereference `e.controller` with no nil check (`chromium.go:299-305`), and `e.controller` is only assigned in the async `CreateCoreWebView2ControllerCompleted` callback. `w.chromium != nil` does not protect against that: `w.chromium` exists from window construction. Ordinary main-thread sequences are safe because `Embed()` blocks in a nested message pump until the controller exists, but a `window.Hide()` posted from a background goroutine during creation is dispatched inside that very pump and lands on a nil controller, which panics and (via the default panic handler) terminates the app. Every other controller-touching site in this file uses the stronger guard `w.chromium.GetController() != nil` (e.g. lines 927, 1009, 2025). Suggest either using that guard at the three new sites, or adding the nil check inside `Chromium.Show`/`Hide` the way `NotifyParentWindowPositionChanged` does (`chromium.go:601`), which would harden every caller at once.

2. **The un-minimize `chromium.Show()` should respect logical visibility.** It currently fires unconditionally, so restoring a minimized window that the app has hidden (`windowShown=false`, e.g. Hide() called while minimized) re-shows the composition surface and resurrects the exact dead zone this PR fixes. Suggest gating it on `w.windowShown || w.showRequested`.

Nit: the comment above both un-minimize blocks reads "Re-assert controller visibility hidden on SIZE_MINIMIZED", which describes the minimize branch, not the restore branch it sits in.

Related: there is a deterministic interaction with the first-navigation Hide/Show nudge that can undo this PR's `hide()` for windows hidden before first paint; details and the one-line fix are on #5743, which is the right place to solve it.

leaanthony:
@wayneforrest pushed 09cce7a60 to this branch (maintainer edit) applying the two review asks so the PR can move: (1) the new Hide/Show sites now use the `w.chromium.GetController() != nil` guard the file's other controller call sites use — `w.chromium` exists from construction but the controller is nil until async creation completes, and a background-goroutine `Hide()` gets dispatched inside Embed's nested message pump; (2) the un-minimise `Show()` is now gated on `windowShown || showRequested` so restoring a window that was hidden while minimised doesn't resurrect the invisible input surface; (3) reworded the un-minimise comments that described the opposite branch. Verified with a windows/amd64 build; hardware verification of the whole batch is in progress.

leaanthony:
**Hardware testing on win-node1 (Windows 11 26200, WebView2 149.0.4022.98, `UseVisualHosting: true`), with a caveat worth recording.**

Method: a self-driving Wails v3 test app (one visible window, one `Hidden: true` never-shown window, one window hidden 30ms after creation, i.e. before first NavigationCompleted), driven remotely via desktop automation that probes `WindowFromPoint` over the hidden windows' screen areas and reports the owning window's class, process, and `GWL_EXSTYLE`.

Findings:

1. On **both** master and the fixed branch, the "empty desktop" over a hidden window's rect is owned by a `Chrome_RenderWidgetHostHWND` (root `Chrome_WidgetWin_1`, `msedgewebview2.exe`) — the invisible surface is real and detectable, matching the report here.
2. **However, on this machine that surface carries `WS_EX_TRANSPARENT` (0x20)** in both builds, so real mouse input passes through it: the user-facing right-click dead zone does not reproduce on this box even on master. The box uses virtual displays (HDMI dummies + RDP), which is the likely reason its DirectComposition input behaviour differs from physical-GPU setups. Conclusion: this environment can neither reproduce nor disprove the user-facing symptom.
3. No regressions from the fix: hidden windows never flash, the visible window minimizes/restores with correct repaint, zoom behaves, and app teardown is clean.
4. One observation for future work: after the fix, the `Chrome_RenderWidgetHostHWND` over the hidden rect remains `WS_VISIBLE` even though the controller was set `IsVisible(false)` — i.e. controller visibility does not hide that particular child HWND under visual hosting on this machine. If the dead zone persists for anyone on physical hardware after this lands, that HWND is where to look next.

@wayneforrest since your machine reproduces the real input dead zone, a quick confirmation on your hardware once this merges would close the loop — the probe evidence says the mechanism your fix targets is the right one, but only a repro machine can prove the symptom is gone.

(Current CI failures on this PR are repo-wide infra breakage, fix in #5746.)

v3/pkg/application/webview_window_windows.go
leaanthony: Good call — done in f7b7cd9e0: both un-minimise `Show()` sites now also gate on `w.webviewNavigationCompleted`, matching `show()`'s inverted form. I deliberately did NOT gate the `Hide()` sites (hide()/SIZE_MINIMIZED): hiding a not-yet-navigated controller is harmless, and before the first NavigationCompleted the controller is visible by default, so an early minimize/hide is exactly when the invisible input surface needs to be taken down. Once navigation completes, the first-paint nudge plus the #5743 guard re-establish the correct state.

leaanthony (approved):
All requested changes are in: GetController guards (09cce7a60), un-minimise gating on logical visibility, comment fixes, and atterpac's webviewNavigationCompleted gating on the Show sites (f7b7cd9e0). Hardware-tested on win-node1 (no regressions vs baseline; details in the test comment above). Approving.

PR #5785, opened 2026-07-15
chore: remove stray v3/old symlink pointing outside the repo

kobihikri:
Hi, and thank you for Wails.

Small cleanup: the repo tracks a symlink `v3/old` whose target is `../../../GolandProjects/ios` — a path outside the repository (a local JetBrains GoLand projects dir):

```
$ gh api repos/wailsapp/wails/contents/v3/old --jq '{type,target}'
{ "type": "symlink", "target": "../../../GolandProjects/ios" }
```

It looks like it was committed by accident. It's a broken/dangling link for everyone who clones the repo (the target doesn't exist), and dangling symlinks pointing outside the tree are worth removing. This PR just deletes it (`git rm v3/old`).

For transparency: I used AI assistance to spot and draft this; I verified the symlink target myself.

## Summary by CodeRabbit

* **Chores**
  * Removed an obsolete path reference.
  * No user-facing functionality changes.


leaanthony (approved):
Confirmed on master: `v3/old` is a committed symlink (mode 120000) pointing at `../../../GolandProjects/ios`, a path on somebody's laptop. The module proxy tolerates it (symlinks are skipped when module zips are built, which is why the alphas published fine), but everyone who clones the repo gets a broken symlink, and it leaks a local directory layout. Good catch, and not something to carry into a release.

PR #5789, opened 2026-07-16
fix(v3/windows): don't fatal on GetCursorPos failure (locked/secure desktop)

wayneforrest:
## Problem

On Windows, `w32.GetAllScreens()` queries `GetCursorPos` (only to set the cosmetic `Screen.IsCurrent` flag) and treats a failure as a hard error:

```go
ret, _, _ := procGetCursorPos.Call(uintptr(unsafe.Pointer(&cursor)))
if ret == 0 {
    return nil, fmt.Errorf("GetCursorPos failed")
}
```

`GetCursorPos` returns `FALSE` / `ERROR_ACCESS_DENIED` whenever the calling process is **not attached to the interactive input desktop** — the workstation is locked, on the secure/UAC desktop, or mid session switch (logon, RDP, fast-user-switch).

At startup this is fatal. `processAndCacheScreens()` runs during `App.Run()` (via `newPlatformApp`) and turns the error into `os.Exit(1)`:

```go
err = result.processAndCacheScreens()
if err != nil {
    app.handleFatalError(err) // -> os.Exit(1)
}
```

So an app launched onto a non-interactive desktop dies before it ever shows, printing the *"catastrophic failure … GetCursorPos failed"* banner. This is reproducible for packaged/MSIX apps registered as a **StartupTask** (the shell launches them at logon, before the session's input desktop is switched away from Winlogon), and for background-update restarts that land while the machine is locked. We hit it in the field on a Microsoft Store build.

## Fix

The cursor position is used only to compute `IsCurrent` (which monitor the pointer is on), and that field has **no readers** — it isn't even propagated to the application-level `Screen`. So a failed query should degrade gracefully instead of aborting:

- `GetAllScreens` now reads the cursor through a small package seam and, when it fails, leaves `IsCurrent = false` and keeps enumerating. The flag self-corrects on the next `WM_DISPLAYCHANGE` / `SPI_SETWORKAREA` re-cache, which already handles errors non-fatally via `handleError`.
- `Win32Menu.ShowAtCursor()` had the same `globalApplication.fatal("GetCursorPos failed")`. A context menu simply cannot be shown on a non-input desktop, so it now returns (with a debug log) — matching the already-graceful `windowsMenu.ShowAtCursor()`, which does `if !ok { return }`.

A Windows regression test overrides the seam to simulate the failure and asserts `GetAllScreens` no longer returns a cursor error and marks no screen current.

## Notes

- Behaviour-preserving on the happy path; no public API change.
- The two spots were inconsistent before this: `windowsMenu.ShowAtCursor()` already tolerated the failure while `Win32Menu.ShowAtCursor()` and `GetAllScreens()` treated it as fatal.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented cursor-position errors from crashing the application when opening popup menus.
  * Improved screen detection reliability when the cursor position cannot be retrieved.
  * The app now continues operating without incorrectly marking a screen as current when cursor information is unavailable.

* **Tests**
  * Added coverage for cursor-position failures during popup menu and screen handling.


wayneforrest:
cc @leaanthony @atterpac 

One more PR - then my app will be crash free.

leaanthony (approved):
This is a crash-class fix and worth landing for the beta.

Two separate failures, both real. `Win32Menu.ShowAtCursor` called `fatal("GetCursorPos failed")`, so a context menu invoked while the workstation is locked or the secure desktop is up killed the application. And `GetAllScreens` returned an error on the same condition, which `processAndCacheScreens` treats as fatal — meaning an app launching on a locked workstation exits before it ever draws a window. That second one is the more serious of the pair and the less obvious.

Degrading is right in both places: the cursor position is only used for the cosmetic `IsCurrent` flag, and the comment correctly notes it self-corrects on the next display-change re-cache.

The package-level `cursorPosForScreens` var is a slightly unusual seam, but it is documented, it is the only practical way to simulate an access-denied `GetCursorPos` in a test, and it comes with a test that exercises the failure path. Good trade.

PR #5805, opened 2026-07-23
[v3] Fix 32-bit build: type updater maxArchiveTotalSize as int64

### LINKED ISSUE #5804: [v3] 32-bit build (GOARCH=386) fails: maxArchiveTotalSize overflows int in updater
**Description**

Building the v3 updater package for a 32-bit target fails to compile:

```
v3/pkg/updater/extract.go:252: cannot use maxArchiveTotalSize (untyped int constant 2147483648) as int value in argument to fmt.Errorf (overflows)
v3/pkg/updater/extract.go:293: cannot use maxArchiveTotalSize (untyped int constant 2147483648) as int value in argument to fmt.Errorf (overflows)
```

`maxArchiveTotalSize` is `2 << 30` (2 GiB = 2147483648). It is an untyped constant, so passing it to `fmt.Errorf` gives it the default type `int` — which is 32 bits on `386` and cannot hold 2147483648.

**To Reproduce**

```
GOOS=windows GOARCH=386 go build ./v3/pkg/updater/
# or, without cgo:
GOOS=linux GOARCH=386 CGO_ENABLED=0 go build ./v3/pkg/updater/
```

**Expected behaviour**

The package compiles for 32-bit targets.

**Version**

v3.0.0-alpha2.115 (also present on current `master`).

gilad-ch:
# Description

`maxArchiveTotalSize` (2 GiB) is an untyped constant that defaults to `int` when passed to `fmt.Errorf`. On 32-bit targets `int` is 32 bits, so the value overflows and the updater package fails to build (`GOARCH=386`). Typing the constant `int64` keeps it `int64` at the format calls; every other use is already `int64` arithmetic, so behaviour is unchanged.

Fixes #5804

## Type of change

# How Has This Been Tested?

This is a compile-time fix. Verified `go build ./v3/pkg/updater/` **fails before** and **succeeds after** the change for a 32-bit target:

- [x] Windows (`GOOS=windows GOARCH=386`)
- [x] Linux (`GOOS=linux GOARCH=386 CGO_ENABLED=0`)

## Test Configuration

Compile-time only (no runtime behaviour change); `go version go1.25`. `wails doctor` not applicable to a build-constant fix.

# Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a 32-bit build issue in the updater that could cause integer overflow when reporting archive extraction errors.
  * Confirmed archive extraction limits and validation behavior remain unchanged.

* **Documentation**
  * Added the fix to the unreleased changelog.


leaanthony (approved):
Verified locally rather than on the description: on master, `GOOS=linux GOARCH=386 go build ./pkg/updater/` fails at extract.go:252 and :293 with `cannot use maxArchiveTotalSize (untyped int constant 2147483648) as int value in argument to fmt.Errorf (overflows)`. With this change the same build is clean. Correct fix, minimal, and it has a changelog entry.

PR #5807, opened 2026-07-24
fix(android): support HTML file inputs via WebChromeClient

### LINKED ISSUE #5781: [v3] File picker fields do not function on android
### Wails version family

v3

### Operating System

Android (experimental)

### Description

Before I start off this report, yes I am aware that a native File select dialog exists, but in this case it's MUCH more of a hassle to implement and use, since I need access to things only provided through the html file field (for example being able to show image previews of selected images, the size, and being able to upload the file along with a form).

When you have an HTML file field inside the app, clicking it does nothing. There are no errors, messages or anything else logged in the debug console either.

Through the normal desktop versions, it works as expected.

As far as I am aware, the web engine on android (not sure which one exactly wails uses), is capable of doing this, but support for it needs to implemented.

### To Reproduce

1. Clone the Android example as per this GitHub repo
2. Add a simple `<input type="file" />` field into it
3. Launch the app through the emulator, or a real device and click the input. Nothing happens

### Expected behaviour

To have the same, or similar response as to what happens when you click on such a field in the mobile browser, or mostly any other android app that runs in a webview

### Screenshots

_No response_

### Attempted Fixes

I have tried to just use the native file dialogs, but as mentioned earlier, this is much more of a hassle to maintain and implement, when it would specifically only apply to android

### System Details

```shell
Wails v3.0.0-alpha2.105 › Wails Doctor
                                                                                                                                                                                                                                    
# System
        
┌────────────────────────────────────────────────────────────────────┐
| Name                | LunaWolf Linux                               |
| Version             | Unknown                                      |
| ID                  | lunawolf                                     |
| Branding            |                                              |
| Platform            | linux                                        |
| Architecture        | amd64                                        |
| Desktop Environment | KDE                                          |
| NVIDIA Driver       | 580.119.02 (98F03F1FC7BAA1EA7283C59)         |
| XDG_SESSION_TYPE    | wayland                                      |
| CPU                 | AMD Ryzen 5 5500                             |
| GPU                 | GM206 [GeForce GTX 950] (NVIDIA Corporation) |
| Memory              | 32GB                                         |
└────────────────────────────────────────────────────────────────────┘
                   
# Build Environment
                   
┌──────────────────────────────────┐
| Wails CLI    | v3.0.0-alpha2.105 |
| Go Version   | go1.25.0          |
| -buildmode   | exe               |
| -compiler    | gc                |
| CGO_CFLAGS   |                   |
| CGO_CPPFLAGS |                   |
| CGO_CXXFLAGS |                   |
| CGO_ENABLED  | 1                 |
| CGO_LDFLAGS  |                   |
| GOAMD64      | v1                |
| GOARCH       | amd64             |
| GOOS         | linux             |
└──────────────────────────────────┘
              
# Dependencies
              
┌──────────────────────────────────────────────────────────────────────────────────────────────┐
| *Android NDK            | /home/hypherionsa/Android/Sdk/ndk/30.0.14904198                    |
| *Android SDK            | /home/hypherionsa/Android/Sdk                                      |
| *Android platform-tools | Installed                                                          |
| *Java (Android)         | openjdk version "17.0.19" 2026-04-21                               |
| gcc                     | 16.1.1+r12+g301eb08fa2c5-1                                         |
| gtk3 (legacy)           | 1:3.24.52-1                                                        |
| gtk4                    | 1:4.22.4-1                                                         |
| npm                     | 10.9.8                                                             |
| pkg-config              | 2.5.1-1                                                            |
| webkit2gtk (legacy)     | 2.52.4-1                                                           |
| webkitgtk-6.0           | 2.52.4-1                                                           |
| docker                  | *Docker version 29.5.2, build 79eb04c7d8 (cross-compilation ready) |
|                                                                                              |
└────────────────────────────────── * - Optional Dependency ───────────────────────────────────┘
         
# Signing
         
┌─────────────────────────────────────────────┐
| macOS Signing   | Not configured            |
| Windows Signing | Not configured            |
| Linux Signing   | GPG key: AA2319E4D334290A |
└─────────────────────────────────────────────┘
                     
# Checking for issues
                     
 SUCCESS  No issues found
           
# Diagnosis
           
(The OS is a custom Arch Based distro)
```

### Additional context

_No response_

mortenolsrud:
# Description

HTML `<input type="file">` elements do nothing on Android. The generated `MainActivity` sets only a `WebViewClient` and never a `WebChromeClient`, so the WebView has no `onShowFileChooser` handler to open the system document picker — clicking a file input silently does nothing (no error, no picker, nothing in the console).

This adds a `WebChromeClient` whose `onShowFileChooser` launches the intent from `FileChooserParams` and delivers the selected URIs back in `onActivityResult`. Multi-select (`<input type="file" multiple>`) is fully supported: the result handler reads `getClipData()` (where Android places multiple selections) with a `getData()` / `parseResult` fallback for single-file pickers. Note: Chromium's `FileChooserParams.parseResult()` only reads `getData()` ([source](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/android_webview/java/src/org/chromium/android_webview/AwContentsClient.java#285)), so it silently drops multi-select results — the `getClipData()` path is required.

As a side effect the `WebChromeClient` also enables the other behaviours it is required for (JS `alert`/`confirm`/`prompt`, `console` message forwarding, and in-page permission prompts such as geolocation / `getUserMedia`).

**Scope:** the change is to the scaffold template (`v3/internal/commands/build_assets/android/app/src/main/java/com/wails/app/MainActivity.java`) only. The `examples/*/build/android` copies are regenerated build output and are already substantially divergent from the current template, so they are intentionally not hand-patched here — happy to update them (or regenerate) if maintainers prefer them kept in sync.

Fixes #5781

## Type of change

# How Has This Been Tested?

Verified end-to-end on an Android emulator (Pixel 7, API 35, x86_64, headless) using a throwaway `wails3 init` vanilla project built with this branch's templates, with a simple `<input type="file">` added to the frontend:

**Single-file selection:**
1. Tapping the file input **opens the Android system document picker** — confirmed via `dumpsys` (`topResumedActivity=com.google.android.documentsui/...PickActivity`) and screenshot. Before this change, tapping did nothing.
2. Selecting a file returns control to the app (`onActivityResult`), and the URI is delivered back to the WebView.
3. The `<input>` is populated and the JS `change` event fires — the File API reports the correct name and size.

**Multi-file selection (`<input type="file" multiple>`):**
4. Using a test app with `multiple`, long-pressed a file in the picker to enter multi-select mode, selected all 3 test files (alpha.txt, beta.txt, gamma.txt), and tapped "Select".
5. All 3 files were delivered via `getClipData()` → the WebView showed **"Selected 3: alpha.txt(48B), beta.txt(47B), gamma.txt(48B)"** with correct names and sizes for each.

Also:
- Cross-compiled the Android host (`pkg/application`) for `android/arm64` against the NDK.
- Ran the host-side command/template test suite (`go test ./internal/commands/...`) — passes.

- [ ] Windows
- [ ] macOS
- [x] Linux

Linux: Ubuntu 24.04.4 LTS (build host). Target/verification: Android (emulator, API 35 x86_64).

## Test Configuration

- Wails CLI: v3.0.0-alpha2.117
- Go: go1.26.5
- Build host: Ubuntu 24.04.4 LTS (Noble), amd64
- Android NDK: 26.3.11579264
- Emulator: Pixel 7, Android API 35, x86_64 (headless, software GPU)

# Checklist:

- [ ] (v2 only) I have updated `website/src/pages/changelog.mdx` with details of this PR (v3 changelog entries are added automatically) — n/a (v3)

Notes on the unchecked boxes (in the interest of an honest checklist):

- **Documentation:** no separate docs — the change is internal to the generated Android host. Glad to add docs if wanted.
- **Tests:** verification here is a manual on-device (emulator) round-trip, not an automated test. This touches the Android host, which is build-tagged native/Java code (`//go:build android` + `MainActivity.java`); the v3 tree currently has no host-runnable harness for the Android runtime (there are no `//go:build android` tests), so I couldn't add a meaningful unit test without standing up new mobile test infrastructure. The existing host template/command tests pass.

## Summary by CodeRabbit

* **New Features**
  * Added support for selecting and uploading files through HTML file inputs on Android.
  * Supports both single and multiple file selection using the system file chooser.
  * Added clearer handling when file-selection dialogs are canceled or unavailable.
  * Preserved existing camera capture and document-picker functionality.

### DIFF
--- a/v3/internal/commands/build_assets/android/app/src/main/java/com/wails/app/MainActivity.java
+++ b/v3/internal/commands/build_assets/android/app/src/main/java/com/wails/app/MainActivity.java
@@ -22,6 +22,10 @@ import android.provider.MediaStore;
 import android.provider.OpenableColumns;
 import android.util.Base64;
 import android.util.Log;
+import android.webkit.ConsoleMessage;
+import android.webkit.ValueCallback;
+import android.webkit.WebChromeClient;
+import android.webkit.WebChromeClient.FileChooserParams;
 import android.webkit.WebResourceRequest;
 import android.webkit.WebResourceResponse;
 import android.webkit.WebSettings;
@@ -68,6 +72,11 @@ public class MainActivity extends AppCompatActivity {
     private static final int PHOTO_CAPTURE_REQUEST = 7002;
     private static final int VIDEO_CAPTURE_REQUEST = 7003;
     private static final int CAMERA_PERMISSION_REQUEST = 7010;
+
+    // In-flight callback for an HTML <input type="file"> chooser opened by the
+    // WebView's WebChromeClient (null when idle).
+    private static final int WEBVIEW_FILE_CHOOSER_REQUEST = 7004;
+    private ValueCallback<Uri[]> webViewFileChooserCallback;
     private File pendingCaptureFile;
     private boolean pendingCaptureIsVideo;
 
@@ -197,6 +206,45 @@ public class MainActivity extends AppCompatActivity {
 
         // Add JavaScript interface for Go communication
         webView.addJavascriptInterface(new WailsJSBridge(bridge, webView), "wails");
+
+        // A WebChromeClient is required for HTML <input type="file"> elements to
+        // work; without onShowFileChooser the WebView silently ignores clicks on
+        // file inputs. It also forwards JS console messages to logcat for
+        // debugging. Note: in-page permission prompts (geolocation, getUserMedia)
+        // require an onPermissionRequest override which is not provided here;
+        // those requests will be denied by the default implementation until a
+        // scoped permissions handler is added.
+        webView.setWebChromeClient(new WebChromeClient() {
+            @Override
+            public boolean onConsoleMessage(ConsoleMessage msg) {
+                if (DEBUG) {
+                    Log.d(TAG, "[WebView] " + msg.messageLevel() + ": " + msg.message()
+                            + " (" + msg.sourceId() + ":" + msg.lineNumber() + ")");
+                }
+                return true;
+            }
+
+            @Override
+            public boolean onShowFileChooser(WebView view,
+                                             ValueCallback<Uri[]> filePathCallback,
+                                             FileChooserParams fileChooserParams) {
+                // Cancel any previous, still-pending chooser before starting a new one.
+                if (webViewFileChooserCallback != null) {
+                    webViewFileChooserCallback.onReceiveValue(null);
+                }
+                webViewFileChooserCallback = filePathCallback;
+
+                Intent intent = fileChooserParams.createIntent();
+                try {
+                    startActivityForResult(intent, WEBVIEW_FILE_CHOOSER_REQUEST);
+                } catch (android.content.ActivityNotFoundException e) {
+                    webViewFileChooserCallback = null;
+                    filePathCallback.onReceiveValue(null);
+                    return false;
+                }
+                return true;
+            }
+        });
     }
 
     private void loadApplication() {
@@ -451,6 +499,41 @@ public class MainActivity extends AppCompatActivity {
             handleCaptureResult(resultCode, data);
             return;
         }
+        if (requestCode == WEBVIEW_FILE_CHOOSER_REQUEST) {
+            if (webViewFileChooserCallback == null) {
+                return;
+            }
+            Uri[] results = null;
+            if (resultCode == RESULT_OK && data != null) {
+                if (data.getClipData() != null) {
+                    // Multiple selection (<input type="file" multiple>) is
+                    // returned as ClipData. FileChooserParams.parseResult only
+                    // reads getData(), so it would drop all but one (and return
+                    // null when getData() is empty) — read every item explicitly.
+                    int count = data.getClipData().getItemCount();
+                    java.util.List<Uri> validated = new java.util.ArrayList<>(count);
+                    for (int i = 0; i < count; i++) {
+                        Uri uri = data.getClipData().getItemAt(i).getUri();
+                        if (isFileChooserUriSafe(uri)) validated.add(uri);
+                    }
+                    if (!validated.isEmpty()) {
+                        results = validated.toArray(new Uri[0]);
+                    }
+                } else if (data.getData() != null) {
+                    Uri uri = data.getData();
+                    if (isFileChooserUriSafe(uri)) {
+                        results = new Uri[]{ uri };
+                    }
+                }
+            }
+            if (results == null) {
+                // Fallback for pickers that report the result differently.
+                results = FileChooserParams.parseResult(resultCode, data);
+            }
+            webViewFileChooserCallback.onReceiveValue(results);
+            webViewFileChooserCallback = null;
+            return;
+        }
         if (requestCode != FILE_PICKER_REQUEST) {
             return;
         }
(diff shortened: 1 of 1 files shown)


v3/internal/commands/build_assets/android/app/src/main/java/com/wails/app/MainActivity.java
```diff
@@ -451,6 +485,15 @@ protected void onActivityResult(int requestCode, int resultCode, @Nullable Inten
             handleCaptureResult(resultCode, data);
             return;
         }
+        if (requestCode == WEBVIEW_FILE_CHOOSER_REQUEST) {
+            if (webViewFileChooserCallback == null) {
+                return;
+            }
+            Uri[] results = FileChooserParams.parseResult(resultCode, data);
```
atterpac: It looks like parseResult will actually just return one item from the list even if multiple are selected
source: https://chromium.googlesource.com/chromium/src/+/refs/heads/main/android_webview/java/src/org/chromium/android_webview/AwContentsClient.java#285

I think we want to use `getClipData()` https://developer.android.com/reference/android/content/Intent#getClipData() with a fallback to `parseResult`

mortenolsrud:
Good catch — you're right, `parseResult` only reads `getData()` ([Chromium source, L289](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/android_webview/java/src/org/chromium/android_webview/AwContentsClient.java#289)), so multi-select via `ClipData` is silently dropped.

Fixed: `onActivityResult` now reads `getClipData()` first (iterating all items for multi-select) with a `getData()`/`parseResult` fallback — mirroring the existing pattern in the SAF native picker handler in the same file (L508-513).

Verified end-to-end on the emulator: selected 3 files via long-press multi-select, all 3 delivered to the WebView ("Selected 3: alpha.txt(48B), beta.txt(47B), gamma.txt(48B)"). Force-pushed.

mortenolsrud:
Thanks for the security note. I considered URI validation here and opted not to add it for these reasons:

1. **Trust model**: The Wails WebView loads only `wails.localhost` (app-local content) — there is no remote/untrusted web content that could use the File API to exfiltrate data to an external server. The selected files become DOM `File` objects accessible to the *same app's own frontend*.

2. **Precedent**: Chromium's own `AwContentsClient.parseFileChooserResult` passes URIs through with zero validation. The existing SAF native-picker handler in the same file (L508-513) does the same — iterates `getClipData()` and delivers URIs directly. This change follows the established pattern.

3. **Over-scoping risk**: Validating "app-private" URIs is non-trivial (ContentProvider URIs are opaque — you can't reliably distinguish "suspicious" from "legitimate" without resolving them), and rejecting valid user selections would break the feature. The Android docs recommend "treat as untrusted" for apps serving untrusted *web* content; that doesn't apply to a local-only WebView shell.

That said, if the maintainers would prefer a validation layer here, I'm happy to add one in a follow-up — perhaps gated on a debug-mode assertion rather than a production rejection, given the trust model.

mortenolsrud:
Addressed in two ways:

1. **In this PR** (force-pushed): the comment has been reworded to accurately state that in-page permission prompts require an `onPermissionRequest` override, and that they will be denied by the default implementation. No longer overstates what the `WebChromeClient` provides by default.

2. **Follow-up branch** (`feat/android-webview-permissions`): adds a scoped `onPermissionRequest` handler that maps WebView resources (`VIDEO_CAPTURE`, `AUDIO_CAPTURE`) to Android runtime permissions, only grants for the `wails.localhost` origin (denying any other origin as a security gate), and holds the request until `onRequestPermissionsResult` delivers the verdict. Also adds `RECORD_AUDIO` to the manifest. This will be a separate PR since it's a feature addition rather than the file-picker bug fix.

mortenolsrud:
All three Copilot findings addressed in the latest force-push:

1. **L223 — `onConsoleMessage` unguarded**: Now wrapped in `if (DEBUG)` to match the rest of the template's `Log.d` calls.

2. **L240 — Missing `onReceiveValue(null)` on `ActivityNotFoundException`**: Fixed — `filePathCallback.onReceiveValue(null)` is now called before returning `false`, so the WebView's file-input promise resolves rather than hanging.

3. **L216 — PR description vs comment mismatch**: The comment already says permissions will be denied until an `onPermissionRequest` override is added (reworded in a prior push). The PR description was also previously updated to match. A separate PR (#5855) adds the scoped permission handler.

PR #5711, opened 2026-07-02
refactor(v3): fold webview2 module into v3/internal/webview2

taliesin-ai:
## Status: ready for review — fully synced, hardware-tested

Now carries everything through today's merges: the #5734 monitor-scale fix, and the full #5741/#5742/#5743 batch (COM by-value marshalling incl. the per-arch `putdouble_*`/`putrect_*` helpers, controller visibility on minimize/hide, hidden-window re-hide) — all relocated into `v3/internal/webview2` (rename-detection for modified files, manual moves for the new helper files).

**Verification (win-node1, Windows 11 26200, WebView2 149.0.4022.98):**
- Self-driving test app built against this branch (no `wailsapp/wails/webview2` requirement in its go.mod — the module dependency is gone), driven via remote desktop automation, two consecutive runs:
  - Zero invisible input surfaces over hidden windows (`window_from_point` grid probes report only the desktop) — the #5742/#5743 behavior, working from the folded location.
  - `SetZoom(1.5)` → `GetZoom() == 1.5` — the #5741 marshalling path, working from the folded location.
  - Minimize/restore repaints identically to the pre-fold build (pixel-checked).
- `go build` + `go vet` clean for windows/amd64 and windows/arm64 on `v3/pkg/application` + `v3/internal/webview2/...`. (windows/386: pre-existing unrelated `pkg/updater` constant overflow; v3's CI matrix doesn't target 386.)

**Impact on existing projects: none.** Old v3 versions keep resolving the published `webview2/v*` tags forever; upgrading projects just lose the transitive requirement at `go mod tidy`; no exported v3 API references webview2 types (verified — the binding was always behind unexported plumbing). Details in the earlier analysis.

PR #5814, opened 2026-07-26
chore: remove the root go.work workspace file

taliesin-ai:
## What

Removes the root `go.work` workspace, and the CI plumbing that only existed to support it.

After the webview2 fold (#5711) the workspace covered a single module, `./v3`, and every CI workflow already sets `GOWORK: "off"`, so it only ever affected local development. There it did more harm than good:

- An active workspace shadows the standalone example modules nested under `v3/`. That is why `v3/examples/mac-window-tabs` has to force `GOWORK: "off"` in two Taskfiles: without it, binding generation finds 0 services and `go mod tidy` targets v3 instead.
- It forced `build-and-test-v3.yml` to run `go work use .` on each generated template project so Go 1.25 workspace mode would not reject it.

Contributors who want a workspace can still create one; it is now gitignored so it stays local.

## Changes

- delete `go.work` and `go.work.sum`
- `build-and-test-v3.yml`: drop the `go.work` / `go.work.sum` path filters, both cache keys, and the `go work use .` step
- `cross-compile-test-v3.yml`: drop the same filters and cache key
- `.gitignore`: ignore `go.work` and `go.work.sum`
- `mac-window-tabs` Taskfiles: keep `GOWORK: "off"` (still correct defensively, for contributors with a local workspace) but reword the comments, which referred to a root workspace that no longer exists

Second commit, while in the same lines: the `webview2/**` path filters and `webview2/go.sum` cache entries in both workflows have been dead since the module was folded into `v3/internal/webview2`. Removed.

## Verification

- `go build ./...` in `v3/` with no workspace active: clean
- a module created at the repo root builds with nothing to reject it, which is what the removed `go work use .` step was compensating for
- both workflows and both Taskfiles parse as YAML
- no `go work` invocation remains anywhere in the repo

https://claude.ai/code/session_01FigQqUQbNu9ngE4a2CSNm8

## Summary by CodeRabbit

- **Chores**
  - Improved automated build and test workflow targeting for v3 changes.
  - Updated caching configuration to make validation runs more focused and reliable.
  - Added local Go workspace files to ignored files.

- **Documentation**
  - Clarified workspace and standalone-module guidance in macOS window tab example tasks.
  - Improved instructions for running example build and development tasks directly or through repository-level commands.


leaanthony (approved):
Verified locally: v3 builds with no workspace active, and a module created at the repo root builds without workspace rejection (what the removed `go work use .` step was compensating for). CI template builds are the real test and are green so far.

The webview2 filter cleanup in the second commit is dead-reference removal from the same lines, no behaviour change.

PR #5815, opened 2026-07-26
ci(v3): harden the release workflow before its first run

taliesin-ai:
## Why now

`release-v3.yml` was added in #5318 and **has never executed** — 0 runs, ever. Every alpha has shipped through `nightly-release-v3.yml`. With a beta coming, the tag-triggered release path is about to run for the first time on the release that matters most.

Reviewing it turns up four things that would each produce a failed or misleading release.

**The largest:** none of the six Apple secrets the workflow reads exist on this repository. Current repo secrets are `CHANGELOG_PUSH_TOKEN`, `CLAUDE_CODE_OAUTH_TOKEN`, `CROWDIN_PERSONAL_TOKEN`, `NPM_TOKEN`, `OPENROUTER_API_KEY`, `SEMGREP_APP_TOKEN`, `SPONSORS_TOKEN`, `WAILS_REPO_TOKEN` — no `APPLE_*`, and none inherited from the org. On the first real run the macOS job fails inside `security import` after three platforms have already built, the `release` job (which `needs` it) never runs, and **nothing is published**.

## Changes

- **Preflight job** that fails in seconds instead of ten minutes in. It resolves the tag for both trigger paths, checks out that tag, and verifies `v3/internal/version/version.txt` matches it. That file is `go:embed`-ed, so it is what `wails3 version` reports: a mismatch ships a binary that misreports itself with no error at all.
- **Preflight names the missing Apple secrets explicitly**, and either stops or — with the new `allow_unsigned_macos` input — continues with signing and notarization skipped. No silent downgrade in either direction.
- **`SHA256SUMS` published** with the binaries. Today a user has no way to verify a download.
- **Build provenance attestation** for every binary. The npm package already ships provenance; the binaries lagging behind it is an inconsistency users can see.
- **New `draft` input**, so the entire path can be rehearsed end to end and the resulting release deleted, without publishing anything.
- Build jobs take their checkout ref from preflight's resolved tag, so all three platforms provably build the same verified source.

## Verification

- `actionlint` clean.
- Version-guard logic exercised locally: rejects `v3.0.0-beta.1` against the current `version.txt` (`v3.0.0-alpha2.117`), accepts the matching tag.
- Missing-secret detection exercised under `bash -eo pipefail` with the secrets both unset and set (the `&&`-in-loop idiom does not trip `set -e` in this position — checked, not assumed).
- Checksum generation exercised, including `sha256sum -c` round-trip.
- **Not exercised in CI**, because that means creating a tag and a release. See below.

## The dry run this enables, and one trap in it

With `draft: true` the release can be rehearsed without publishing. The tag it needs is not free of consequence though: a throwaway tag must sort **below** the real beta under semver precedence, or it becomes what `@latest` resolves to for anything that considers prereleases. `v3.0.0-test.1` would **outrank** `v3.0.0-beta.1` (`test` > `beta` alphabetically) and hijack resolution — the same failure mode as the old `alpha.98-tui` tag. A tag in the existing `alpha2.` series (e.g. `v3.0.0-alpha2.999`) sorts below `beta.1` and is safe.

https://claude.ai/code/session_01FigQqUQbNu9ngE4a2CSNm8

## Summary by CodeRabbit

* **New Features**
  * Added manual release options for draft publishing and unsigned macOS builds when signing is unavailable.
  * Added build provenance attestations and generated SHA256 checksums for release artifacts.

* **Bug Fixes**
  * Releases now consistently build from the exact resolved tag across all supported platforms.
  * Added validation to ensure the release tag matches the application version.
  * macOS signing and notarization now run only when valid signing credentials are available.

### DIFF
--- a/.github/workflows/release-v3.yml
+++ b/.github/workflows/release-v3.yml
@@ -15,6 +15,16 @@ on:
         required: false
         default: true
         type: boolean
+      allow_unsigned_macos:
+        description: 'Publish unsigned, un-notarized macOS binaries (only if the Apple secrets are not configured yet)'
+        required: false
+        default: false
+        type: boolean
+      draft:
+        description: 'Create the release as a draft (use this to rehearse the release without publishing anything)'
+        required: false
+        default: false
+        type: boolean
 
 env:
   GO_VERSION: '1.25'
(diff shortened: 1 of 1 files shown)


leaanthony (approved):
The missing-secrets finding is the important one: with no APPLE_* secrets on the repo, the first tag push would have built three platforms, died in `security import`, and published nothing. Failing in preflight with the names listed is the right shape.

Version guard is worth having independently of that — an embedded version.txt that disagrees with the tag produces a binary that lies quietly, which is the worst class of release bug.

Checksums and provenance bring the binaries in line with what the npm package already does.

Approving on review; the path itself still needs the draft dry run before we trust it.

PR #5816, opened 2026-07-26
docs: state the terms contributions are accepted under

taliesin-ai:
`CONTRIBUTING.md` never said what licence contributions come in under, whether a CLA applies, or whether a sign-off trailer is required. That is a routine question for anyone contributing from inside a company, and leaving it implicit means each of them has to ask or guess.

This states the position that already applies rather than introducing a new one: MIT in, MIT out, no CLA, no DCO trailer, and third-party code disclosed in the pull request.

Found while auditing release readiness for the beta: "contribution terms documented" is a standard legal gate for a major release, and it was the only one of that group failing.

If you would rather require DCO sign-off, that is a different and larger change (it needs a bot and it affects every contributor), so I have not assumed it here.

## Summary by CodeRabbit

* **Documentation**
  * Added guidance on licensing and contribution provenance.
  * Clarified MIT licensing expectations for contributions and third-party code.
  * Noted that no contributor license agreement or sign-off trailer is required.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,3 +23,18 @@ Thanks for your interest in contributing! Wails has two active tracks. Pick the 
 - Update the relevant changelog file (see above).
 - Follow the existing coding style in the area you're touching.
 - Add tests where it makes sense.
+
+## Licence and provenance
+
+Wails is released under the [MIT Licence](LICENSE), and contributions come in
+under the same terms. By opening a pull request you confirm that:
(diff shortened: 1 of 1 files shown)


leaanthony (approved):
States the position that already applies rather than inventing one, which is the right scope for this. MIT in/out matches the LICENSE and the package metadata, and not requiring a CLA or DCO trailer is the status quo — there is no bot for either.

PR #5817, opened 2026-07-26
docs(v3): add the v2 vs v3 comparison page to the migrate command

taliesin-ai:
Stacked on #5758 (base is `feat/wails3-migrate-command`, not master), so it can be merged into that branch whenever convenient. I did not push directly to `feat/wails3-migrate-command` because it is checked out in an active local worktree.

## Why

There are two open `wails3 migrate` drafts: #5758 (this base) and #5755. #5758 is the better implementation — the migration logic lives in its own `internal/migrate` package rather than in `internal/commands`. But #5755 has one thing #5758 does not: `docs/src/content/docs/migration/comparison.mdx`, a side-by-side v2→v3 reference. That page is a stated beta requirement, and it is what someone actually reaches for mid-migration instead of reading a 750-line guide top to bottom.

This ports that page across, so #5755 can be closed without losing it.

## Changes

- `migration/comparison.mdx` from #5755, and its sidebar registration.
- The `UNRELEASED_CHANGELOG.md` entry this branch was missing (the repo's convention is one entry per user-visible change; the automation uses yours if present).
- Rewords the one FAQ line describing v3 as "in alpha/beta" to talk about pre-1.0 API stability instead. Release-status wording in reference docs goes stale the moment the status changes, and it belongs in the release announcement rather than scattered through the docs.

## Verification

- `go build ./...` in v3: clean.
- `go test ./internal/migrate/... ./internal/commands/`: pass.
- Docs site builds: 1721 pages, all locales.

## Note on #5755

Once this lands, #5755 is fully superseded: its command implementation is the weaker of the two and its only unique asset is now here. Worth closing it with a pointer to #5758 so contributors are not looking at two competing drafts.


leaanthony (approved):
Right call to stack this rather than push into the worktree branch, and right call to keep the comparison page with the command rather than landing it on master ahead of `wails3 migrate` existing. The alpha/beta reword is worth having independently: release-status wording in reference docs is stale the day the status changes.

Not merging this into feat/wails3-migrate-command myself since that branch is checked out in an active worktree — merge it when convenient.

PR #5820, opened 2026-07-26
ci(v3): keep @wailsio/runtime in lockstep with the CLI version

taliesin-ai:
## The problem

`publish-npm.yml` bumped the package with:

```
npm --no-git-tag-version --force version prerelease
```

That increments the trailing numeric identifier and knows nothing about the release channel. It is exactly why the two version series drifted apart: the CLI moved `alpha` → `alpha2` → `beta` while npm carried on counting `alpha.N`.

Today the CLI reports `v3.0.0-alpha2.117` and npm's latest is `3.0.0-alpha.97`. At beta that becomes user-visible: a `v3.0.0-beta.0` CLI sitting next to an `@wailsio/runtime` that still says `alpha`, and every upgrading user has to be told the mismatch is fine.

## The change

The runtime version now comes from `v3/internal/version/version.txt` — the same file `go:embed` puts into the CLI — so the two always agree.

Two consequences handled:

- **Republish.** The version now only moves when `version.txt` moves, so a run triggered by a runtime source change between releases would try to republish an existing version and fail. Those runs now skip the publish step while still committing regenerated assets, with a notice explaining why.
- **Trigger.** `version.txt` is added to both the trigger paths and the detect job's file list, so a release that moves it does publish.

## Merging this publishes nothing

A workflow-file-only change does not match the detect job's file lists, so the publish job will not run on merge. I deliberately did **not** bump `package.json` in this PR for that reason.

The next publish happens when `version.txt` next moves, and it will be `3.0.0-alpha2.117` rather than `3.0.0-alpha.98`. That sorts higher under semver, so `latest` resolution and every downstream pin behave correctly. It is a visible jump in the npm series, which is the one-off cost of ending the drift.

## Verification

- `actionlint` clean.
- Version derivation exercised against the real `version.txt` (`v3.0.0-alpha2.117` → `3.0.0-alpha2.117`) and against the beta branch value (`v3.0.0-beta.0` → `3.0.0-beta.0`).
- The already-published guard exercised against the live registry: correctly reports `3.0.0-alpha.97` as present (would skip) and `3.0.0-beta.0` as absent (would publish).

Once this is in, the alpha-to-beta upgrade page on the beta branch loses its "the runtime version will not match your CLI" caveat, which is a better outcome than documenting the confusion.

## Summary by CodeRabbit

* **New Features**
  * Runtime releases now use the exact version specified for the release.
  * Publishing automatically skips versions already available on npm while continuing related release updates.
  * Release publishing now applies the appropriate npm `latest` tag behavior.
* **Bug Fixes**
  * Changes to the release version now reliably trigger the publishing workflow.

### DIFF
--- a/.github/workflows/publish-npm.yml
+++ b/.github/workflows/publish-npm.yml
@@ -6,6 +6,9 @@ on:
       # Self-heal: if bundle bytes ever change on master without a matching
       # source change, this rebuild overwrites them with source-derived output.
       - 'v3/internal/assetserver/bundledassets/**'
+      # The runtime version is derived from this file, so a release that moves
+      # it must trigger a publish.
+      - 'v3/internal/version/version.txt'
       - '.github/workflows/publish-npm.yml'
   workflow_dispatch:
 
@@ -47,6 +50,7 @@ jobs:
             v3/internal/runtime/desktop/@wailsio/runtime/tsconfig.json
             v3/internal/runtime/desktop/@wailsio/runtime/src/**
             v3/internal/assetserver/bundledassets/**
+            v3/internal/version/version.txt
             v3/pkg/events/events.txt
             v3/tasks/events/**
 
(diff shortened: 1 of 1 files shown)


leaanthony (approved):
Deriving from version.txt is the right anchor: it is already the single source the CLI embeds, so lockstep falls out rather than needing to be maintained.

The republish guard is the part I would have missed. Without it, the first runtime source change after a release fails the workflow at npm publish, which is precisely the failure mode that left npm frozen at alpha.79 for months last time.

Agreed on not bumping package.json here so the merge stays inert.

PR #5818, opened 2026-07-26
chore(beta): merge master into releases/v3-beta

taliesin-ai:
`releases/v3-beta` had drifted **31 commits behind master**. Everything landing on master since — the `go.work` removal (#5814), the release workflow hardening (#5815), the crash fixes going in now — was absent from the branch the beta will actually be cut from. That kind of drift is only ever discovered at the worst possible moment.

Base is `releases/v3-beta`, not master. Nothing here goes near master, and nothing is published.

## Conflicts, both resolved in favour of the beta branch

- `v3/internal/version/version.txt` keeps `v3.0.0-beta.0` rather than master's `v3.0.0-alpha2.117`.
- The FAQ production-readiness answer keeps the branch's beta wording rather than master's "active pre-release" phrasing.

## Verification

- No conflict markers remain; the beta wording in `faq.mdx` and `status.mdx` survived the merge.
- `go build ./...` in v3 behaves **identically to master** — the `examples/` failures are pre-existing on both, not introduced here. (Worth a separate look at some point: `go build ./...` from `v3/` has been failing on master for those example packages, even though CI's `task test:examples` passes.)
- Docs site builds: 1711 pages, all locales.
- `go.work` correctly absent after the merge.

## What this branch already carries

The beta prep that was done earlier: FAQ rewritten with verified commands and platform support, docs wording promoted from alpha to beta, `version.txt` at `v3.0.0-beta.0`, SECURITY.md supported-versions, channel-aware release-notes and changelog validation, and nightly change-detection extended to beta/rc tags.

PR #5821, opened 2026-07-26
chore(beta): merge master into releases/v3-beta (second sync)

taliesin-ai:
Second sync. The branch was **six commits behind master again within hours** of the first one (#5818), which is the whole point: a release branch does not announce that it is aging, it just quietly stops containing things.

Clean merge, no conflicts. Picks up:

- the contribution terms (#5816), which clears the `legal-provenance` audit warning on this branch
- the release workflow hardening (#5815)
- the runtime/CLI version lockstep (#5820)
- the crash fixes merged since the first sync

## Invariants checked after merging

- `version.txt` still `v3.0.0-beta.0`
- beta wording still present in `faq.mdx` and `status.mdx`
- contribution terms present in CONTRIBUTING.md
- lockstep logic present in `publish-npm.yml`
- preflight present in `release-v3.yml`

## This is now enforced

`relman` gained a `release-branch-current` blocker that compares the release branch against its base and fails the audit when it falls behind, with the fix in the message. It caught this drift:

```
FAIL release-branch-current     Release branch is not behind its base [blocker]
       6 commit(s) on origin/master are not on this branch
       fix: merge origin/master in and re-run the audit; do this immediately
            before tagging, not the day before
```

Re-syncing immediately before the tag should be a step in the release runbook, not something anyone has to remember.


leaanthony (approved):
Six commits of drift in a few hours makes the point better than the first sync did. Clean merge, invariants checked, and the new audit blocker means the next one gets caught rather than noticed.

PR #5822, opened 2026-07-26
chore(beta): add the release record and comms drafts

taliesin-ai:
**Draft, on `releases/v3-beta`. Nothing here is published and nothing goes near master.**

Puts the release plan, the gate configuration and the per-channel comms drafts in the repository beside the code, so the beta is resumable by someone who is not the person who prepared it. `.release/v3.0.0-beta.0/`.

## What is filled in

- **`comms/launch-sequence.md`** — the ordered runbook, made specific to this project rather than generic: `release.go` tags and bumps `version.txt`, the tag push triggers `release-v3.yml`, `version.txt` moving triggers the npm publish, docs deploy to Cloudflare Pages. Plus the abort criteria, and an honest rollback section noting that npm cannot be unpublished after 72 hours and the Go module proxy caches permanently, so the real recovery is a fast patch.
- **`comms/packagers.md`** — the facts downstream needs: the version-series move, the runtime now matching the CLI version, checksums and provenance now shipping, artifact names unchanged, `go.work` gone, and macOS signing conditional on credentials.
- **`plan.json`** — schedule, gates and the audit configuration for this repo, so `relman audit` keeps working for whoever runs the release.
- **`README.md`** — what each file is, and the preconditions that still block tagging.

## What is deliberately empty

`positioning` and `headlines` in `plan.json`, and therefore the TODOs in the announcement and social drafts.

I had filled these with placeholder copy while dry-running the tooling, and stripped it before committing. Inventing a release's one-liner, problem statement and proof point produces text that reads plausibly and says something nobody decided to say, which is worse than a blank field. Fill those five fields and re-run `relman kit -plan .release/v3.0.0-beta.0/plan.json` and every channel draft regenerates with real copy.

## Audit state on this branch

17 passing, 0 warnings, 2 blockers. `changelog-entry` clears at release time when `release.go` folds the unreleased changelog in; `ci-green` just needs the plan's branch name to match what CI runs against.

### DIFF
--- a/.release/v3.0.0-beta.0/README.md
+++ b/.release/v3.0.0-beta.0/README.md
@@ -41,3 +41,43 @@ Against this branch: 17 checks passing, 0 warnings, 2 blockers.
    tag in the `alpha2.` series before using it for real.
 3. This branch drifts behind master quickly. Re-sync and re-audit immediately
    before tagging; `relman audit` now blocks on it.
+
+## Findings worth acting on after the beta
+
+Noticed while preparing this release. None are release-blocking; all of them
+cost something later if forgotten.
+
+**v3 CI compiles the GTK4 backend against exactly one GTK version.** The v3 test
+matrix is `[windows-latest, ubuntu-latest, macos-latest]`, so `linux_cgo.go` is
+only ever built against whatever GTK ubuntu-latest carries. The single
+`ubuntu-22.04` job in the repository lives in `build-and-test.yml`, which is the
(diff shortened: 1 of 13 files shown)


leaanthony (approved):
Worth having on the branch rather than sitting in a draft: if the beta gets cut before this merges, the branch carries the code and the docs but not the gate record or the launch runbook.

Positioning and headlines being empty is the right call. Generated copy that reads plausibly while saying something nobody decided is worse than a blank field, and the note explains how to fill them and regenerate.

PR #5678, opened 2026-06-26
fix(v3/macOS): prevent OpenFile dialog crash when UTType cannot resolve filter extensions

phergul:
# Description

Fixes a crash on macOS when a file dialog filter contains extensions that `UTType` cannot resolve

`showOpenFileDialog` builds `allowedContentTypes` by calling `UTType typeWithFilenameExtension` for each filter extension. That API returns nil when it is not a 'recognised' UTType. Inserting nil into the `NSMutableArray` throws `NSInvalidArgumentException` and crashes the app.

- `dialogs_darwin.go`: Skip nil UTTypes and add unrecognized extensions to the legacy `setAllowedFileTypes` API (same path that is used for macOS < 11). Recognized extensions continue to use `setAllowedContentTypes`.
- `dialogs_darwin_delegate.m`: Match files by filename suffix instead of pathExtension, so compound extensions like tar.gz are selectable (`pathExtension` only returns gz for example.tar.gz). 

I checked how [Electron handles this](https://github.com/electron/electron/blob/88cd4b418618424fbcd11917fffee489f534ad72/shell/browser/ui/file_dialog_mac.mm#L110), and they simply take only the last extension from the users input (e.g. .tar.gz -> .gz) and filter on that. Comparing using the filename suffix lets matching be more accurate.

Fixes #5675 

## Type of change
  
Please select the option that is relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

- [ ] Windows
- [x] macOS
- [ ] Linux

- go test ./...
- go mod replace with branch, tested the unrecognised extensions (tar.gz, busted.exe). Also tested with full list of both  recognised and unrecognised ones.
  
## Test Configuration

```
Wails v3.0.0-alpha2.106 › Wails Doctor

# System

┌──────────────────────────────┐
| Name          | MacOS        |
| Version       | 26.5.1       |
| ID            | 25F80        |
| Branding      | MacOS 26.5.1 |
| Platform      | darwin       |
| Architecture  | arm64        |
| Apple Silicon | true         |
| CPU           | Apple M4 Pro |
| CPU           | Apple M4 Pro |
| GPU           | 16 cores     |
| Memory        | 24 GB        |
└──────────────────────────────┘

# Build Environment

┌──────────────────────────────────────────────────────────────────────┐
[...]

PR #5823, opened 2026-07-26
chore(beta): merge master into releases/v3-beta (third sync)

taliesin-ai:
Third sync. Picks up the four fixes merged since the last one:

- macOS OpenFile dialog crash when UTType cannot resolve filter extensions (#5678)
- the stray `v3/old` symlink pointing at a local `GolandProjects` path (#5785)
- Windows fatal on `GetCursorPos` failure on locked and secure desktops (#5789), which also fixed an app launching on a locked workstation exiting before it drew a window
- the accompanying changelog entries

Clean merge, no conflicts. Verified after merging: `version.txt` still `v3.0.0-beta.0`, and `v3/old` is correctly absent.

Batched deliberately rather than syncing after each merge: every sync is cheap on this branch, but a PR per fix is noise in the branch history.

### DIFF
--- a/v3/UNRELEASED_CHANGELOG.md
+++ b/v3/UNRELEASED_CHANGELOG.md
@@ -24,6 +24,7 @@ After processing, the content will be moved to the main changelog and this file 
 
 ## Fixed
 <!-- Bug fixes -->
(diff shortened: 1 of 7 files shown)


leaanthony (approved):
Three crash-class fixes and the symlink removal, batched into one sync. Invariants verified after the merge rather than assumed.

PR #5793, opened 2026-07-17
fix(v3): guard nil AllowDarkModeForWindow to avoid startup panic

roachadam:
# Description

Fixes a nil-pointer panic on startup when a window uses a Dark (or system-dark) native title bar on Windows builds that do not load the dark-mode `uxtheme` APIs — e.g. Windows 10 1809 / Windows Server 2019 (build 17763).

`pkg/w32/theme.go` loads `AllowDarkModeForWindow` (and siblings) only on build `>= 18334`, so on 1809 the package-level `w32.AllowDarkModeForWindow` is `nil`. `(*windowsWebviewWindow).run` called it **unguarded** in the `Dark` and dark `SystemDefault` branches (`webview_window_windows.go), while `w32.SetMenuTheme` already guards the identical call. 

The fix extracts the call into a small applyDarkMode helper that nil-guards it (mirroring the guard w32.SetMenuTheme already uses) and calls it from both branches, so the window degrades gracefully (default title bar) instead of panicking. No behavior change on builds that load the procs; no new dependencies.

Fixes #5792 

## Type of change
  
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?
  
Built a minimal v3 app with `Windows: application.WindowsWindow{Theme: application.Dark}` and ran it on Windows Server 2019 (build 17763): before, it panicked on startup; after, the window opens normally. 
Reproduction repo: https://github.com/roachadam/wails-1809-dark-title-bar-poc. 

- [x] Windows
- [ ] macOS
- [ ] Linux
      
## Test Configuration
```
Wails: v3.0.0-alpha2.117
OS:    Windows Server 2019 Standard Evaluation, 1809 (Build 17763)
Arch:  amd64
WebView2 runtime: 150.0.4078.65
```

Wails is not installed on this VM, so there is no wails3 doctor output. Reproduced with a minimal app instead: my existing app (built on v3.0.0-alpha.68) failed to start on Windows Server 2019 Standard Evaluation (build 17763). I reproduced the panic in isolation using the POC above built against the tagged v3.0.0-alpha2.117 release, then rebuilt the POC against this fix and confirmed the window opens normally.

# Checklist:

- [ ] (v2 only) I have updated `website/src/pages/changelog.mdx` with details of this PR (v3 changelog entries are added automatically)
[...]

PR #5827, opened 2026-07-26
docs: make English the single source for diagram source, and stop committing the renders

taliesin-ai:
Two problems with one cause, found while working out why 110 SVGs had landed on master.

## The diagrams were never localised

Every label in every locale is English. What the per-locale copies held was **drift**: the translation pipeline copied the d2 block when a page was translated and never re-synced it when the English diagram changed.

The signature is unmistakable — for `architecture-0.svg`, de/fr/ja/ru were **byte-identical to each other** and different from English:

| Variant | Locales |
|---|---|
| `3555f454…` | en, id (current) |
| `7a8a77aa…` | de, fr, ja, ru (one older vintage) |
| `8bf91064…` | zh-cn (older still) |

Seven locales were rendering an outdated architecture diagram, with English labels, on a page explaining how Wails works. The German page's d2 source is a *different diagram* (`direction: right`, a `User` person shape), not a translated one.

## The renders were committed build output

110 SVGs were committed in #5793 — **that was my mistake**, a `git add -A` while resolving a changelog conflict in a worktree where I'd run the docs build. astro-d2 writes one SVG per diagram per locale, so the count is 13 diagrams × 10 locales and grows as translation coverage improves. It is build output, not source.

## Changes

- Sync the d2 source from each English page into its translated counterparts where the diagram counts line up (7 files).
- Untrack `docs/public/d2` (110 files) and gitignore it.

Done with a small Go tool that discovers English pages carrying diagrams rather than hard-coding a list, so a new diagram page is picked up automatically. It refuses to guess when counts differ.

## Nine files deliberately not synced

They need a human, and the tool says so rather than guessing:

- `zh-cn/concepts/architecture.mdx` — 4 diagrams where English has 8
- eight locales' `concepts/lifecycle.mdx` — 1 diagram where English has 2

In both cases the English page gained diagrams the translation never received. Inserting a diagram into translated prose is not a mechanical substitution.

## Verification

- **Prose provably untouched**: stripped the d2 blocks from each changed file before and after, and confirmed the remainder is byte-identical. German prose still reads as German.
- **Clean rebuild** after `rm -rf public/d2 dist`: 1711 pages, same as before, no page loss.
- **Convergence**: `architecture-0` went from three variants to one shared render everywhere except zh-cn, which is one of the deliberately skipped files. Every other diagram is now a single variant across all locales.
- **No broken images**: all 110 diagram references in the built HTML resolve to a file on disk.
- **Gitignore works**: after the build, `git status` on `docs/public/d2` is empty.

## Not included

`contributing/architecture.mdx` and `concepts/build-system.mdx` still show `**[... Diagram Placeholder]**`. The sources are recoverable from `0d02db8ce`, but that commit removed them *because they broke the build*, so restoring them means fixing the d2 first. Worth doing, separately.

## Summary by CodeRabbit

- **Documentation**
  - Updated architecture diagrams across localized documentation to clarify application startup, frontend–backend communication, and system API interactions.
  - Revised event-system diagrams to illustrate event subscriptions, JSON-based delivery, and handling across multiple windows.
- **Chores**
  - Excluded generated documentation diagram output from source control.


leaanthony:
Retargeted from master to `releases/v3-beta` at Lea's request. The diff is unchanged (118 files, +270/-20637): the beta branch already contains everything on master, so this merges as exactly the one commit.

leaanthony (approved):
The finding underneath this is the part that matters: the per-locale diagrams were never translations, they were stale copies, so seven locales were serving an outdated architecture diagram with English labels.

Verification is the right shape too — prose proven untouched by hashing the non-d2 remainder rather than eyeballing the diff, convergence measured per diagram, and every built image reference checked against disk.

Refusing to sync the nine mismatched files is correct. Inserting a diagram into translated prose needs someone who can read the prose.

PR #5833, opened 2026-07-26
docs: make English the single source for diagram source, and stop committing the renders

taliesin-ai:
Cherry-pick of `cb210d15f`, already merged into `releases/v3-beta` as #5827. Applying it to master too so the branches agree.

Without this, master keeps the 110 committed SVGs and the stale diagram source in seven locales, and the next `master → releases/v3-beta` sync quietly reintroduces the divergence the beta branch just fixed.

## What it does

**The per-locale diagrams were never translations.** Every label in every locale is English. The translation pipeline copied the d2 block when a page was translated and never re-synced it when the English diagram changed, so de/fr/ja/ru were byte-identical to *each other* and different from English — the signature of a stale copy. Seven locales rendered an outdated architecture diagram, with English labels, on a page explaining how Wails works.

**The 110 SVGs are build output.** astro-d2 writes one per diagram per locale, so the count is 13 × 10 and grows with translation coverage. They were committed by mistake in #5793 — mine, a `git add -A` in a worktree where I'd run the docs build.

- Syncs d2 source from each English page into its translated counterparts where diagram counts line up (7 files).
- Untracks `docs/public/d2` and gitignores it.
- Leaves nine files alone deliberately: zh-cn's architecture page has 4 diagrams where English has 8, and eight locales' lifecycle pages have 1 where English has 2. The English pages gained diagrams the translations never received, and inserting one into translated prose is not mechanical.

## Verification (from the original PR, plus a clean cherry-pick)

- Prose provably untouched: the non-d2 remainder of each changed file hashes identically before and after; German prose still reads as German.
- Clean rebuild after deleting `public/d2` and `dist`: 1711 pages, no page loss.
- `architecture-0` went from three variants across locales to one shared render everywhere except zh-cn, which is one of the skipped files.
- All 110 diagram references in the built HTML resolve to a file on disk.
- Cherry-picked onto current master with no conflicts.

## Summary by CodeRabbit

- **Documentation**
  - Updated architecture diagrams across multiple languages to clearly illustrate app initialization, frontend–backend JSON communication, and OS API interactions.
  - Reworked event system diagrams to show window subscriptions, backend event emission, routing, and delivery to subscribers.
  - Excluded generated documentation diagram output from source control.


leaanthony (approved):
Cherry-pick of cb210d15f, already on releases/v3-beta as #5827. Applying it to master keeps the branches from diverging on this, which matters because the divergence is the kind a routine sync silently undoes.

Issue #3578, opened 2024-06-28
Double click on hidden titlebar with wails draggable should maximize window


RichardLindhout:
### Description

Double click on hidden titlebar with wails draggable should maximize window

### To Reproduce

1. Hide
```
Mac: &mac.Options{
			TitleBar: &mac.TitleBar{
				TitlebarAppearsTransparent: true,
				HideTitle:                  true,
				HideTitleBar:               false,
				//FullSizeContent:            false,
				//UseToolbar:                 false,
				//HideToolbarSeparator:       true,
			}```
			
			```
  if (isWails() && isMac()) {
    return (
      <div
        style={{
          // @ts-ignore
          '--wails-draggable': 'drag',
          height: 24,
          // @ts-ignore
          WebkitAppRegion: 'drag',
        }}
      />
    );
  }

```

### Expected behaviour

To maximize window on mac os if title is hidden

### Screenshots

_No response_

### Attempted Fixes

_No response_

### System Details

```shell
OSX
```

RichardLindhout:
Sort of resolved this now 
```go
func (a *App) TitleBarPressedDouble() {
	// maximize window or minimize window in wails
	if wailsruntime.WindowIsMaximised(a.ctx) {
		wailsruntime.WindowUnmaximise(a.ctx)
	} else {
		wailsruntime.WindowMaximise(a.ctx)
	}
}
```
```tsx
    // detect double click
    return (
      <div
        onDoubleClick={() => {
          TitleBarPressedDouble();
        }}
        style={{
          // @ts-ignore
          '--wails-draggable': 'drag',
          height: 24,
          // @ts-ignore
          WebkitAppRegion: 'drag',
        }}
      />
    );
  }
```

josefzemanp:
Yes, I also encounter the same requirement. :)

leaanthony:
There's a workaround posted so closing this for now.

leaanthony:
Just seen your PR 👍

josefzemanp:
Yes, I think that could be a good and easy solution

josefzemanp:
@leaanthony can I have a review, please? [ https://github.com/wailsapp/wails/pull/4996](https://github.com/wailsapp/wails/pull/4996)

PR #5852, opened 2026-07-31
feat(v3): support native titlebar double-click on macOS

leaanthony:
## Summary

Ports and improves the macOS custom-titlebar behaviour proposed for v2 in #4996.

Double-clicking a v3 draggable region now follows the native macOS title-bar preference:

- **Maximize** toggles maximised/restored state
- **Minimize** minimises the window
- **None** leaves the window unchanged
- fullscreen windows ignore the gesture

## Design

The v2 PR injects an additional `WKUserScript` and sends a magic message directly through the WebKit bridge. The v3 port instead:

- reuses the existing `--wails-draggable` hit-testing in the bundled runtime;
- emits a dedicated drag double-click message only on macOS;
- accepts the gesture only in the top-level document, preventing iframe-triggered window actions;
- routes it through `WebviewWindow.HandleMessage`;
- reads `AppleActionOnDoubleClick` natively;
- defaults a missing preference to macOS zoom behaviour and safely ignores unknown future values;
- keeps non-macOS platforms as explicit no-ops.

Related: #4996
Addresses the v3 side of #3578.

## Validation

- TypeScript: `tsc --noEmit`
- Runtime tests: **1,169 passed**
- Go 1.25 focused preference-mapping test: passed
- `git diff --check`: passed
- CodeRabbit local review: no findings

A full Linux `pkg/application` build was attempted in the Go 1.25 container, but the container does not include GTK4/WebKitGTK 6 development packages. The failure was dependency discovery through `pkg-config`, before compiling these changes. macOS runtime verification remains required because the native implementation is Darwin-only.

### DIFF
--- a/v3/internal/runtime/desktop/@wailsio/runtime/src/drag.ts
+++ b/v3/internal/runtime/desktop/@wailsio/runtime/src/drag.ts
@@ -8,7 +8,7 @@ The electron alternative for Go
 (c) Lea Anthony 2019-present
 */
 
-import { invoke, IsWindows, IsLinux } from "./system.js";
+import { invoke, IsWindows, IsLinux, IsMac } from "./system.js";
 import { GetFlag } from "./flags.js";
 import { canTrackButtons, eventTarget } from "./utils.js";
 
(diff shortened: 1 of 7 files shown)


taliesin-ai:
Closing and recreating this automated draft under the required taliesin-ai account; the branch and code are unchanged.

PR #5875, opened 2026-08-02
fix(v3/w32): call GetStockObject instead of GetDeviceCaps

### LINKED ISSUE #5874: [v3] w32.GetStockObject calls GetDeviceCaps and always returns 0
### Wails version family

v3

### Exact Wails version

v3.0.0-beta.2

### Operating System

Windows

### Description

`w32.GetStockObject` invokes the wrong Win32 export and returns 0 for every input. It has never worked.

`v3/pkg/w32/gdi32.go`:

```go
func GetStockObject(fnObject int) HGDIOBJ {
	ret, _, _ := procGetDeviceCaps.Call(
		uintptr(fnObject))

	return HGDIOBJ(ret)
}
```

It calls `procGetDeviceCaps`, not `procGetStockObject` — even though `procGetStockObject` is declared a few lines above, at `gdi32.go:40`, and is otherwise unused.

`GetDeviceCaps` takes `(HDC, index)`. Passing a stock-object constant such as `DEFAULT_GUI_FONT` (17) as the `HDC` is an invalid handle, so the call returns 0. Every `GetStockObject` call therefore yields a null GDI handle.

There are no callers in v3 today, which is why this has gone unnoticed. It only surfaces for a new caller, and then indirectly: the null handle is passed to some other GDI function, which fails for reasons that point away from the actual cause. I lost time to exactly this while prototyping owner-drawn menus — `SelectObject(hdc, GetStockObject(DEFAULT_GUI_FONT))` terminated the process, and the null handle was several steps removed from the visible symptom.

This is the same class of defect as the `DestroyMenu` argument bug already covered by `v3/pkg/w32/menu_windows_test.go`, where a syscall wrapper passed the wrong arguments and silently returned failure for every call.

### To Reproduce

No special hardware or Windows version required.

```go
//go:build windows

package w32_test

import (
	"testing"

	"github.com/wailsapp/wails/v3/pkg/w32"
)

func TestRepro(t *testing.T) {
	for name, obj := range map[string]int{
		"DEFAULT_GUI_FONT": w32.DEFAULT_GUI_FONT,
		"SYSTEM_FONT":      w32.SYSTEM_FONT,
		"BLACK_BRUSH":      w32.BLACK_BRUSH,
		"WHITE_BRUSH":      w32.WHITE_BRUSH,
	} {
		if got := w32.GetStockObject(obj); got == 0 {
			t.Errorf("GetStockObject(%s) = 0, want a valid handle", name)
		}
	}
}
```

`go test ./pkg/w32/` on Windows: every case fails with 0.

### Expected behaviour

`GetStockObject` returns a valid, usable GDI handle for standard stock objects. Stock objects are always available, so a 0 return indicates the wrapper is calling the wrong export.

### Screenshots

_No response_

### Attempted Fixes

_No response_

### System Details

```shell
Wails: v3.0.0-beta.2
OS:    Any Windows build — not version-specific
Arch:  amd64

Found on Windows Server 2019 (build 17763) but the defect is in the wrapper
itself and is independent of Windows version.
```

### Additional context

Found while investigating menu theming on Windows; unrelated to that work, so filing separately.

While confirming this, I also noticed that `SelectObject` panics on failure rather than returning 0. That turns a null handle from this bug into a process termination, since `SelectObject` is called from inside Windows message handlers. I am not proposing a change to it here: panicking on failure is the established convention across `pkg/w32` (39 call sites in 10 files, including `SetTextColor` and `SetBkMode` in the same file), so changing one function in isolation would be arbitrary. Happy to raise that separately as a design question if it is worth discussing.

**AI Usage Disclosure**: I used AI in investigating and writing this issue to confirm my own initial findings.

roachadam:
# Description

`w32.GetStockObject` invoked the wrong Win32 export and returned 0 for every input. It called `procGetDeviceCaps` rather than `procGetStockObject`, even though `procGetStockObject` is declared a few lines above at `gdi32.go:40` and is otherwise unused.

`GetDeviceCaps` takes `(HDC, index)`, so passing a stock-object constant such as `DEFAULT_GUI_FONT` (17) as the `HDC` is an invalid handle and the call returns 0. The function had never worked.

There are no callers in v3 today, which is why it went unnoticed. It only surfaces for a new caller, and then indirectly — the null handle is passed to some other GDI function, which fails for reasons that point away from the actual cause.

This is the same class of defect as the `DestroyMenu` argument bug already covered by `pkg/w32/menu_windows_test.go`, where a syscall wrapper passed the wrong arguments and silently returned failure for every call.

Fixes #5874

## Type of change

# How Has This Been Tested?

Adds `v3/pkg/w32/gdi32_windows_test.go`, which runs on Windows CI via the `//go:build windows` constraint. **It fails on the pre-fix code**, where every call returns 0.

The test checks several stock objects are non-zero, then verifies the `DEFAULT_GUI_FONT` handle is genuinely usable — selecting it into a memory DC (not the screen DC, so there are no side effects on the CI desktop) and confirming a previous object comes back.

No manual testing was required; the behaviour is fully reproducible in CI.

`GOOS=windows GOARCH=amd64 go build ./pkg/w32/ ./pkg/application/` succeeds, and `go vet ./pkg/w32/` reports the same pre-existing findings as master — this branch adds none.

Since `GetStockObject` has no callers in v3, this cannot regress existing behaviour.

- [x] Windows
- [ ] macOS
- [ ] Linux

Not applicable to macOS or Linux: the change is inside a `//go:build windows` file.

## Test Configuration

```
Wails:  v3.0.0-beta.2
OS:     Any Windows build — the defect is in the wrapper and is not version-specific
Arch:   amd64
```

Verified via the added test on Windows CI. My development machine is macOS, so I cannot provide `wails3 doctor` output; the test is the verification here.

# Checklist:

- [ ] (v2 only) I have updated `website/src/pages/changelog.mdx` with details of this PR (v3 changelog entries are added automatically) — n/a, v3 change
- [x] I have made corresponding changes to the documentation — n/a; no public API or documented behaviour change

I have also added an entry to `v3/UNRELEASED_CHANGELOG.md` to control the wording, as described in CONTRIBUTING.md.

**Note on a related observation:** while confirming this, I noticed `SelectObject` panics on failure rather than returning 0, which turns a null handle from this bug into a process termination since it is called from inside Windows message handlers. I deliberately did **not** change it here — panicking on failure is the established convention across `pkg/w32` (39 call sites across 10 files, including `SetTextColor` and `SetBkMode` in this same file), so altering one function in isolation would be arbitrary. Happy to raise it separately as a design question.

**AI Usage Disclosure**: I used AI in investigating and preparing this change to confirm my own initial findings.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where retrieving standard Windows graphics objects returned invalid results.
  * Standard stock objects now return usable handles as expected.
  * Fixed unreadable dark-mode menu text on Windows 10 version 1809 and Windows Server 2019.

* **Tests**
  * Added regression coverage for stock object retrieval and usage.

* **Documentation**
  * Documented these fixes in the unreleased changelog.


leaanthony:
@copilot resolve the merge conflicts on this branch.

leaanthony:
Thanks for this! :pray:

leaanthony (approved):

PR #5877, opened 2026-08-02
fix(v3/windows): pass HWND to AllowDarkModeForWindow

leaanthony:
## Summary

Follow-up to #5873.

- passes both native arguments required by ordinal 133: `HWND` and the allow flag
- adds a Windows-only AST regression test that verifies the `SyscallN` call preserves the native ABI argument order

## Why

The wrapper was declared as `AllowDarkModeForWindow(HWND, bool)` but invoked `SyscallN` with only the Boolean value. Lowering the uxtheme gate in #5873 expanded that broken call path to Windows build 17763, so this corrects the ABI immediately while keeping the original fix independently merged.

## Testing

CI should exercise the Windows package and the new regression test.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dark mode handling on Windows by correctly applying the setting to the selected window.

* **Tests**
  * Added Windows-specific validation to ensure dark mode requests use the correct parameters.

### DIFF
--- a/v3/pkg/w32/theme.go
+++ b/v3/pkg/w32/theme.go
@@ -114,7 +114,7 @@ func init() {
 					if allow {
 						allowInt = 1
 					}
-					ret, _, _ := syscall.SyscallN(procAllowDarkModeForWindow, uintptr(allowInt))
+					ret, _, _ := syscall.SyscallN(procAllowDarkModeForWindow, uintptr(hwnd), uintptr(allowInt))
 					return ret
 				}
 			}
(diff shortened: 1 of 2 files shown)


roachadam:
I was just about to file this. I have another few related issues PRs that I will open too.

leaanthony:
Reviewed for Beta 3. The native call fix is correct and should be included. The Windows CI failure is isolated to the new test: importing `go/token` as `token` collides with the package-level `token` declaration in `gdiplus.go`. Alias that import (for example `gotoken`) and rerun the Windows suite; no behavioral issue was found in the production change.

PR #5879, opened 2026-08-03
fix(v3/windows): preserve native menu readability

### LINKED ISSUE #5878: Dark-on-dark menu text when wails dark theme is set on a light-mode system
### Wails version family

v3

### Exact Wails version

v3.0.0-beta.2

### Operating System

Windows

### Description

A window that sets `Windows.Theme: Dark` while the OS is in **light** mode gets a dark menu background with dark menu text. The menu items are unreadable.

This is the case left over from #5872 — that issue's before/after matrix listed `light OS + Theme: Dark` as unreadable both before and after the fix, and #5873 did not address it. I have now confirmed it still reproduces on current master (`a2680fef5`).

**Cause**
Two pieces of code decide the menu's appearance, and they use different definitions of "dark".

The **background** comes from `updateTheme` in `webview_window_windows.go`, which paints a `#212121` brush via `MENUINFO`/`HbrBack` based on the *window's* resolved theme. `Theme: Dark` means dark, regardless of the OS.

The **text** comes from uxtheme, and `SetMenuTheme` in `pkg/w32/theme.go:203` downgrades itself to the light theme whenever the *system* is light:
```go
if useDarkMode && ShouldAppsUseDarkMode != nil && !ShouldAppsUseDarkMode() {
    useDarkMode = false
}
```

So when the window says dark and the system says light, the background follows the window and the text follows the system. Dark on dark.

### To Reproduce

1. Set Windows to light mode theme
2. Build a v3 app with a window using `Windows: application.WindowsWindow{Theme: application.Dark}` and an application menu with a few items.
3. Open a menu.

The popup is dark and the item text is dark.
Verified with High Contrast off and DWM `ColorPrevalence` at 0, so neither is masking the result.

### Expected behaviour

A window that asks for `Theme: Dark` gets a readable menu — or, if that is not achievable, one that is at least legible rather than dark-on-dark.

### Screenshots

Image

### Attempted Fixes

I tested three approaches on Windows Server 2019, build 17763, in light mode with a window requesting `Theme: Dark`. All three binaries were built from unmodified master (`a2680fef5`) and differed only in what the app did at startup.

**1. `SetPreferredAppMode(PreferredAppModeForceDark)` — does not work.**
No change; menus stayed unreadable. The call was accepted and returned the previous mode, so it was a genuine no-op rather than a failure:
```
SetPreferredAppMode(ForceDark=2) returned 1 (previous mode)
ShouldAppsUseDarkMode() before = false
ShouldAppsUseDarkMode() after  = false
```
Three problems, only one of which needed testing:
- It is process-wide, so it cannot express a per-window `Theme` — one dark window alongside one light window is impossible.
- It does not exist below build 18334. On 17763 ordinal 135 is `AllowDarkModeForApp(BOOL)`, where `2` merely reads as truthy. The platform from #5872 would stay broken.
- `ShouldAppsUseDarkMode()` is unmoved by the process's preferred mode, as the log shows, so `SetMenuTheme`'s guard would still downgrade. This was never a one-line fix on any build.

**2. Remove the guard and apply `DarkMode_Explorer` anyway — does not work.**
This is the obvious cheap fix, so I tested it directly, bypassing the guard:
```
SetWindowTheme(0x1102ea, DarkMode_Explorer) hr=0x0 err=The operation completed successfully.
```
The call succeeded — `hr=0x0` — and the menu was still unreadable. The window's theme class does not control popup menu item text colour. (Established on 17763, where the process's immersive policy is stuck light.)

**3. Match the background to what uxtheme will draw — works, but concedes the case.**
Downgrade the menu brush whenever `ShouldAppsUseDarkMode()` is false, so background and text agree. Roughly ten lines, negligible risk, fixes readability on every build.
It is not a fix for the underlying problem: a window asking for `Theme: Dark` gets a light menu, and `MenuBarTheme`'s implied control over menu text stays unimplementable. It resolves the contradiction by discarding the requested theme.

**4. Owner-draw the popup items — works.**
Setting `MF_OWNERDRAW` and handling `WM_MEASUREITEM` / `WM_DRAWITEM` lets Wails supply both background and text, so an explicit theme renders correctly on any build and any system setting. I have a working prototype that renders dark menus with readable text on 17763 in the failing configuration — roughly 300 lines of new drawing code.

One implementation note that cost me some time and is worth recording: there are two menu builders, `windowsMenu.processMenu` in `menu_windows.go` and `Win32Menu.buildMenu` in `popupmenu_windows.go`. `setMenu` calls both but attaches only the `HMENU` from the second, so anything configured in the first is silently discarded. `MF_OWNERDRAW` has to be set in `buildMenu` to have any effect.

### System Details

```shell
Wails: v3.0.0-beta.2 (built from master a2680fef5)
OS:    Windows Server 2019 Standard Evaluation, Version 1809 (Build 17763)
Arch:  amd64
High Contrast:       off
DWM ColorPrevalence: 0
AppsUseLightTheme:   1 (apps light)

The cause is not version-specific — the guard and the brush behave the same way on every build — so I would expect this on Windows 10 and 11 too. I have only measured 17763, which is the hardware I have.
Development machine is macOS; binaries cross-compiled with `GOOS=windows GOARCH=amd64`.
```

### Additional context

Follow-up to #5872 / #5873, which fixed the dark-OS cases on build 17763 but not this one.

**Question for maintainers before I open a PR:** options 3 and 4 are the survivors, and they are very different in scope. Option 3 is a small PR I could open today. Option 4 changes menu rendering for every Windows user and reads to me like it needs a WEP under "significant platform-specific work" — I have a draft ready if that is the preferred direction. Happy to go either way; I would rather ask than guess.

**AI Usage Disclosure**: I used AI in investigating and writing this issue to confirm my own initial findings.

leaanthony:
## Summary

- keep a requested dark application window dark
- use the matching light native menu palette when Windows is in light app mode and cannot render matching dark menu text
- add focused coverage for the native-menu policy decision

Fixes #5878

## Why

Native popup-menu text is selected from Windows process-level colour policy, while the menu background is selected per Wails window. A dark Wails window on a light Windows system therefore produced dark text on a dark background. This is the smallest safe Beta fix: it preserves legibility without introducing owner-drawn menus or a new public theming API.

## Validation

- `go test ./pkg/application -run 'Test(UseDarkNativeWindowsMenu|EffectiveZoomButtonState)$' -count=1`\n- `GOOS=windows GOARCH=amd64 go test -c ./pkg/application`\n- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows native menu readability when dark mode is requested while the app uses a light theme.
  * Native menu colors now correctly respect the system’s dark-mode preference.
* **Documentation**
  * Added a changelog entry describing the Windows menu appearance fix.

### DIFF
--- a/v3/UNRELEASED_CHANGELOG.md
+++ b/v3/UNRELEASED_CHANGELOG.md
@@ -23,6 +23,7 @@ After processing, the content will be moved to the main changelog and this file 
 
 ## Fixed
 <!-- Bug fixes -->
+- Prevent unreadable native menu text when a Windows application requests dark mode while the Windows app theme is light. The menu now uses the matching light native background until Windows can render dark menu text.
 - Fix unreadable native menus on Windows 10 1809 / Windows Server 2019 (build 17763). The dark-mode uxtheme exports were gated on build 18334, so the app-level dark-mode opt-in never ran on those hosts: the menu background was painted dark but Windows kept drawing menu text in the light theme, leaving dark text on a dark background. The ordinals exist from 17763, so the gate now matches.
 
 ## Deprecated
--- a/v3/pkg/application/webview_window_options.go
+++ b/v3/pkg/application/webview_window_options.go
@@ -33,6 +33,15 @@ func effectiveZoomButtonState(a, b ButtonState) ButtonState {
 	return a
 }
 
+// useDarkNativeWindowsMenu reports whether Windows can draw a dark native menu
+// for a dark application window. A menu background can be selected per window,
+// but Windows selects native menu text from its process-level colour policy. If
+// the system policy is light, retaining a dark background makes the text
+// unreadable, so callers must fall back to the matching light menu instead.
+func useDarkNativeWindowsMenu(windowIsDark bool, systemAppsUseDarkMode func() bool) bool {
+	return windowIsDark && (systemAppsUseDarkMode == nil || systemAppsUseDarkMode())
+}
+
 type WindowStartPosition int
 
 const (
--- a/v3/pkg/application/webview_window_options_test.go
+++ b/v3/pkg/application/webview_window_options_test.go
@@ -457,3 +457,25 @@ func TestEffectiveZoomButtonState(t *testing.T) {
 		})
 	}
 }
+
+func TestUseDarkNativeWindowsMenu(t *testing.T) {
+	tests := []struct {
+		name       string
+		windowDark bool
+		systemDark func() bool
+		want       bool
+	}{
+		{"light window", false, nil, false},
+		{"dark window with unavailable policy", true, nil, true},
+		{"dark window with dark system policy", true, func() bool { return true }, true},
+		{"dark window with light system policy", true, func() bool { return false }, false},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := useDarkNativeWindowsMenu(tt.windowDark, tt.systemDark); got != tt.want {
+				t.Fatalf("useDarkNativeWindowsMenu() = %t, want %t", got, tt.want)
+			}
+		})
+	}
+}
--- a/v3/pkg/application/webview_window_windows.go
+++ b/v3/pkg/application/webview_window_windows.go
@@ -1480,9 +1480,13 @@ func (w *windowsWebviewWindow) updateTheme(isDarkMode bool) {
 	}
 
 	w32.SetTheme(w.hwnd, isDarkMode)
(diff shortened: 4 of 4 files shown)


leaanthony:
Manual Windows validation passed on win-node1 (Windows 11 build 26200) using commit a581ef319. With `AppsUseLightTheme=1` and the Wails menu example forced to `Windows.Theme: Dark`, both the menubar and the native Demo popup rendered a light background with dark, legible text. The popup exercise covered disabled, checked, radio, submenu, and accelerator entries.

PR #5897, opened 2026-08-04
Fix/mac background app focus steal

julianstorer:
# Description

`activateIgnoringOtherApps` was called unconditionally in `applicationDidFinishLaunching`, dragging even `Accessory`/`Prohibited` (background/agent) apps to the foreground and taking focus from whatever the user was doing — for instance the terminal that launched a headless process.

This force-activates only when the activation policy is `Regular`, so background/agent apps launch without grabbing focus.

## Type of change

# How Has This Been Tested?
Launched an app with an `Accessory` activation policy from a terminal and
confirmed focus stayed with the terminal; confirmed a normal `Regular` app still
activates on launch as before.
- [ ] Windows
- [x] macOS   
- [ ] Linux

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved macOS application activation behavior.
  * Accessory and background applications no longer appear unexpectedly when launched.

### DIFF
--- a/v3/pkg/application/application_darwin.go
+++ b/v3/pkg/application/application_darwin.go
@@ -314,7 +314,14 @@ func (m *macosApp) run() error {
 				C.bool(m.parent.options.Mac.ApplicationShouldTerminateAfterLastWindowClosed),
 			)
 			C.setActivationPolicy(C.int(m.parent.options.Mac.ActivationPolicy))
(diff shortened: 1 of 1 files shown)


leaanthony (approved):

leaanthony:
Legend!! 🙏

PR #5898, opened 2026-08-04
Fix/linux isvisible before widget

julianstorer:
# Description

`isVisible()` called `gtk_widget_is_visible(w.gtkWidget())` unconditionally, but the GTK widget is created lazily. On GTK4 that's deferred until after the app's `activate` signal: `Run` sets `w.impl` and then blocks in `waitForActivation` before the widget exists, leaving a startup window where `w.impl != nil` but `w.window` is NULL. An `IsVisible()` poll during that gap passed NULL to `gtk_widget_is_visible` and tripped a GTK-CRITICAL assertion. This guards the nil widget and reports not-visible, which is what such a window is.

## Type of change

# How Has This Been Tested?
Polled `IsVisible()` during startup (before the activate signal) and confirmed
it returns false with no GTK-CRITICAL output, on both GTK4 and GTK3 builds.
- [ ] Windows
- [ ] macOS
- [x] Linux 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved window visibility handling during startup.
  * Prevented visibility checks before the native window is fully initialized, reducing potential startup issues.

### DIFF
--- a/v3/pkg/application/linux_cgo.go
+++ b/v3/pkg/application/linux_cgo.go
@@ -1166,6 +1166,17 @@ func (w *linuxWebviewWindow) isMinimised() bool {
 }
 
 func (w *linuxWebviewWindow) isVisible() bool {
+	// The GTK widget is created lazily in run() (windowNew). On GTK4 that only
+	// happens after the application's "activate" signal fires: WebviewWindow.Run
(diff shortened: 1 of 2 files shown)


leaanthony (approved):

leaanthony:
Nice find! 🙏

PR #5908, opened 2026-08-05
fix(v3): unblock iOS and Android cross-compiles (5 build-tag/API bugs)

### LINKED ISSUE #5907: v3 mobile: iOS and Android cross-compiles fail to build (5 distinct build-tag/API bugs)
### Description

Cross-compiling a real v3 app (backend Go service + vanilla HTML/JS frontend, no mobile-specific code beyond what the templates generate) for `GOOS=ios` and `GOOS=android` fails outright on `v3.0.0-alpha.98` — before even reaching app-specific code. Filed with a PR that fixes all five.

### Environment

- Wails CLI: v3.0.0-alpha.98
- Go: go1.26.3 darwin/arm64
- macOS 26.6, Xcode 26.6
- Android NDK 27.1.12297006, `compileSdk 34`, `minSdk 21`

### Bugs found (all reproduced by direct `go build`, not just via `wails3 task`)

**1. `wails3 ios overlay:gen` can't run outside the Wails monorepo checkout**

`internal/commands/ios_overlay_gen.go`'s `repoRoot()` walks up from the current working directory looking for a folder named `v3/internal/commands`. That only exists inside a checkout of `wailsapp/wails` itself — any normal downstream project (including the one generated by `wails3 init -t vanilla`) has no such path, so `wails3 ios overlay:gen -out build/ios/xcode/overlay.json -config build/config.yml` always fails with a bare `file does not exist`. This blocks `task ios:build`/`task ios:run` for literally any project.

**2. iOS cross-compile: `linux_cgo.c`/`linux_cgo.h` build tags don't exclude Android — wait, this one's Android, see #4.**

**3. iOS cross-compile: `messageprocessor.go:201` references `androidMethodNames` unconditionally**

`pkg/application/messageprocessor.go` has a Debug-logging switch that references both `iosMethodNames` and `androidMethodNames` with no build tag on the file itself. `messageprocessor_ios.go` (`//go:build ios`) defines `iosMethodNames` but never defines `androidMethodNames` — only `messageprocessor_android.go` (`//go:build android`) and the desktop stub (`//go:build !ios && !android`) do. Result: `GOOS=ios go build` fails with `undefined: androidMethodNames`.

**4. iOS cross-compile: `events_common_ios.go:9` references `events.IOS`, which doesn't exist**

`pkg/events/events.go:488` defines `var iOS = newIOSEvents()` — lowercase, unexported — while every sibling (`Common`, `Linux`, `Mac`, `Windows`) is exported. `events_common_ios.go` was written against the exported name (`events.IOS.ApplicationDidFinishLaunching`), so it fails with `undefined: events.IOS`. Looks like a partial rename.

**5. Android cross-compile: `linux_cgo.c`/`linux_cgo.h` build tag is missing `&& !android`**

Both files have `//go:build linux && !gtk3 && !server`. Go implicitly sets the `linux` build tag when `GOOS=android` (Android is Linux-based), so these GTK-dependent cgo files get pulled into Android builds too and fail on `gtk/gtk.h: file not found`. `application_linux.go`/`linux_cgo.go` already correctly exclude `!android` — just the `.c`/`.h` files were missed.

**6. Android cross-compile: `messageprocessor_android.go` is missing `iosMethodNames`** — same root cause as #3, mirrored: `GOOS=android` fails with `undefined: iosMethodNames`.

**7. Android cross-compile: `pkg/events` has no `Android` events var at all**

`events_common_android.go:9` references `events.Android.ActivityCreated`, but `pkg/events` never defines an `Android` var/struct (only `Common`/`Linux`/`Mac`/`Windows`/`IOS`). Fails with `undefined: events.Android`.

**8. Android cross-compile: `application_android.go`'s JNI page-finished handler calls `runtime.Core()` with no arguments**

Every other platform calls `runtime.Core(globalApplication.impl.GetFlags(globalApplication.options))` (see `webview_window_darwin.go`, `webview_window_windows.go`, `webview_window_linux.go`). The Android call site just does `runtime.Core()`, which doesn't match `runtime.Core(flags map[string]any)`'s signature — `androidApp.GetFlags` already exists, it's just never wired into this call.

### Reproduction

```bash
# iOS
SDK_PATH=$(xcrun --sdk iphonesimulator --show-sdk-path)
GOOS=ios CGO_ENABLED=1 GOARCH=arm64 \
CGO_CFLAGS="-isysroot $SDK_PATH -target arm64-apple-ios15.0-simulator -mios-simulator-version-min=15.0" \
CGO_LDFLAGS="-isysroot $SDK_PATH -target arm64-apple-ios15.0-simulator" \
go build -buildmode=c-archive -tags ios,debug -o /tmp/app.a .
# → undefined: androidMethodNames, undefined: events.IOS

# Android (NDK r27)
export CC=".../aarch64-linux-android21-clang" CGO_ENABLED=1 GOOS=android GOARCH=arm64
go build -buildmode=c-shared -tags android,debug -o /tmp/libwails.so .
# → fatal error: 'gtk/gtk.h' file not found
# (after working around that) → undefined: iosMethodNames, undefined: events.Android,
#   not enough arguments in call to runtime.Core
```

### Fix

PR incoming with all fixes applied — verified not just compile-clean but running: packaged and launched a real `.app` in the iOS Simulator, and a real debug APK in an arm64-v8a Android emulator, both booting the Go runtime and completing the native JS-bridge handshake without crashing.

mgueregath:
Fixes #5907.

Five small, independent bugs that keep `GOOS=ios` and `GOOS=android` from compiling at all on `v3.0.0-alpha.98`, found while cross-compiling a real third-party app (Go service backend + plain HTML/JS frontend, nothing exotic).

## What's fixed

- **`pkg/application/linux_cgo.c` / `linux_cgo.h`**: build tag was `linux && !gtk3 && !server`, missing `&& !android`. Go implicitly sets the `linux` tag when `GOOS=android` (Android is Linux-based under the hood), so these GTK-dependent cgo files were pulled into Android builds and failed on a missing `gtk/gtk.h`. The `.go` counterparts (`application_linux.go`, `linux_cgo.go`) already excluded `!android` correctly — only the cgo files were missed.
- **`pkg/application/messageprocessor_ios.go`**: added the (empty) `androidMethodNames` map. `messageprocessor.go`'s Debug-logging switch references both `iosMethodNames` and `androidMethodNames` unconditionally, but the iOS file only defined its own map, leaving `androidMethodNames` undefined on iOS builds — even though `processAndroidMethod` already has an iOS-side stub right next to it.
- **`pkg/application/messageprocessor_android.go`**: same fix, mirrored — added the empty `iosMethodNames` map so Android builds don't fail on the same reference in reverse.
- **`pkg/events/events.go`**: exported `iOS` → `IOS`. `events_common_ios.go` references `events.IOS.ApplicationDidFinishLaunching`, but the package only defined the lowercase, unexported `iOS` — every sibling (`Common`, `Linux`, `Mac`, `Windows`) is already exported, so this reads like a partial rename that never got finished. Confirmed no other internal references to the lowercase name exist.
- **`pkg/events/events.go`**: added a minimal `Android` events var (`androidEvents{ ActivityCreated }`) — `events_common_android.go` references `events.Android.ActivityCreated`, but no `Android` var existed at all in `pkg/events`. Kept intentionally minimal (just the one field that's actually referenced) rather than guessing at a fuller Android lifecycle event set.
- **`pkg/application/application_android.go`**: wired `globalApplication.impl.GetFlags(globalApplication.options)` into the JNI page-finished handler's `runtime.Core(...)` call. Every other platform (`webview_window_darwin.go`, `webview_window_windows.go`, `webview_window_linux.go`) already calls `runtime.Core(globalApplication.impl.GetFlags(globalApplication.options))`; the Android call site called `runtime.Core()` with no arguments, which doesn't match `runtime.Core(flags map[string]any) string`'s signature. `androidApp.GetFlags` already existed, just wasn't being used here.

## Verification

Not just compile-clean — built and ran both targets end to end:

- **iOS**: `go build -buildmode=c-archive -tags ios,debug ...` succeeds, linked into a `.app` with `clang`, installed and launched in the iOS Simulator (`xcrun simctl install/launch`) via `codesign --sign -`. `simctl` logs show the app boots, `application_ios.go`'s `processApplicationEvent` fires for `ApplicationDidBecomeActive` (the exact `events.IOS.*` path from bug #4) with no crash.
- **Android**: `go build -buildmode=c-shared -tags android,debug ...` succeeds against NDK r27 (`aarch64-linux-android21-clang`), packaged into a debug APK via the generated Gradle project, installed and launched on an arm64-v8a emulator (Pixel 9 Pro API 35). `adb logcat` shows `WailsBridge` initializing, the JNI `nativeOnPageFinished` callback firing, and `runtime.Core()`'s log line executing successfully — the exact call site from bug #6 — with no `FATAL`/crash.

Both apps render a blank page in this test (an unrelated debug-mode asset-serving issue, `WailsPathHandler: Asset not found: /index.html` — happening because I built the c-archive/c-shared manually instead of through `wails3 dev`'s asset pipeline, not something this PR touches), but the Go runtime itself boots and the native bridge handshake completes cleanly on both platforms, which is what these five bugs were blocking.

## Notes

- Bug 1 from the linked issue (`wails3 ios overlay:gen` failing outside the Wails monorepo, in `internal/commands/ios_overlay_gen.go`'s `repoRoot()`) is **not** fixed by this PR — it's a CLI tooling issue, not a build-tag bug, and felt like a separate fix. Happy to take a pass at that in a follow-up if useful.
- Didn't attempt to flesh out a fuller `events.Android` lifecycle event set beyond `ActivityCreated` since I don't have visibility into what the Android-side native code actually emits — kept the fix scoped to what's referenced today.

## Summary by CodeRabbit

* **New Features**
  * Added Android activity-created event support for improved application lifecycle integration.
  * Improved Android runtime configuration handling so application settings are applied consistently.
  * iOS webviews now load the configured start URL when available.

* **Bug Fixes**
  * Improved cross-platform compatibility and diagnostic behavior on Android and iOS.
  * Prevented platform-specific build components from being included in unsupported environments.
  * Corrected iOS event naming for clearer and more consistent API usage.
  * Improved iOS application packaging to identify the correct executable during builds.

### DIFF
--- a/v3/internal/commands/updatable_build_assets/ios/Info.dev.plist.tmpl
+++ b/v3/internal/commands/updatable_build_assets/ios/Info.dev.plist.tmpl
@@ -3,7 +3,7 @@
 <plist version="1.0">
 <dict>
     <key>CFBundleExecutable</key>
-    <string>{{.BinaryName}}</string>
+    <string>$(EXECUTABLE_NAME)</string>
     <key>CFBundleIdentifier</key>
     <string>{{.ProductIdentifier}}.dev</string>
     <key>CFBundleName</key>
--- a/v3/internal/commands/updatable_build_assets/ios/Info.plist.tmpl
+++ b/v3/internal/commands/updatable_build_assets/ios/Info.plist.tmpl
@@ -3,7 +3,7 @@
 <plist version="1.0">
 <dict>
     <key>CFBundleExecutable</key>
-    <string>{{.BinaryName}}</string>
+    <string>$(EXECUTABLE_NAME)</string>
     <key>CFBundleIdentifier</key>
     <string>{{.ProductIdentifier}}</string>
     <key>CFBundleName</key>
--- a/v3/pkg/application/application_android.go
+++ b/v3/pkg/application/application_android.go
@@ -433,7 +433,7 @@ func Java_com_wails_app_WailsBridge_nativeOnPageFinished(env *C.JNIEnv, obj C.jo
 		if win != nil {
 			androidLogf("info", "🤖 [JNI] Injecting runtime.Core() into window %d", id)
 			// Get the runtime core JavaScript
-			runtimeJS := runtime.Core()
+			runtimeJS := runtime.Core(globalApplication.impl.GetFlags(globalApplication.options))
 			androidLogf("info", "🤖 [JNI] Runtime JS length: %d bytes", len(runtimeJS))
 			app.windowsLock.RUnlock()
 			// IMPORTANT: We must bypass win.ExecJS because it queues if runtimeLoaded is false.
--- a/v3/pkg/application/linux_cgo.c
+++ b/v3/pkg/application/linux_cgo.c
@@ -1,4 +1,4 @@
-//go:build linux && !gtk3 && !server
+//go:build linux && !gtk3 && !server && !android
 
 #include "linux_cgo.h"
 
--- a/v3/pkg/application/linux_cgo.h
+++ b/v3/pkg/application/linux_cgo.h
@@ -1,4 +1,4 @@
-//go:build linux && !gtk3 && !server
+//go:build linux && !gtk3 && !server && !android
 
 #ifndef LINUX_CGO_H
 #define LINUX_CGO_H
--- a/v3/pkg/application/messageprocessor_android.go
+++ b/v3/pkg/application/messageprocessor_android.go
@@ -12,6 +12,11 @@ const (
 	AndroidToast          = 2
 )
 
+// iosMethodNames is empty on Android: messageprocessor.go's Debug logging
+// path references it unconditionally (for the iosRequest case), but this
+// platform never dispatches iOS requests (see processIOSMethod below).
+var iosMethodNames = map[int]string{}
+
 var androidMethodNames = map[int]string{
 	AndroidHapticsVibrate: "Haptics.Vibrate",
 	AndroidDeviceInfo:     "Device.Info",
--- a/v3/pkg/application/messageprocessor_ios.go
+++ b/v3/pkg/application/messageprocessor_ios.go
@@ -18,6 +18,12 @@ const (
 	IOSUserAgentSet                     = 8
 )
 
+// androidMethodNames is empty on iOS: messageprocessor.go's Debug logging
+// path references it unconditionally (for the androidRequest case), but
+// this platform never dispatches Android requests (see
+// processAndroidMethod below).
+var androidMethodNames = map[int]string{}
+
 var iosMethodNames = map[int]string{
 	IOSHapticsImpact:                    "Haptics.Impact",
 	IOSDeviceInfo:                       "Device.Info",
--- a/v3/pkg/application/webview_window_ios.go
+++ b/v3/pkg/application/webview_window_ios.go
@@ -17,6 +17,8 @@ void ios_window_set_background_color(void* viewController, unsigned char r, unsi
 import "C"
 import (
 	"unsafe"
+
+	"github.com/wailsapp/wails/v3/internal/assetserver"
 )
 
 // iosWebviewWindow implements the webviewWindowImpl interface for iOS
@@ -215,6 +217,18 @@ func (w *iosWebviewWindow) run() {
 				w.nativeHandle,
 				C.uchar(rgba.Red), C.uchar(rgba.Green), C.uchar(rgba.Blue), C.uchar(rgba.Alpha),
 			)
+
+			// Unlike every other platform's run() (see e.g.
+			// webview_window_darwin.go), this never told the webview to
+			// actually navigate anywhere — the WKWebView got created (so
+			// the background colour showed) but never loaded options.URL,
+			// leaving it permanently blank regardless of what the asset
+			// server serves. Mirrors macOS's run(): resolve the start URL
+			// and load it right after creating the native handle.
+			startURL, err := assetserver.GetStartURL(w.parent.options.URL)
+			if err == nil {
+				w.setURL(startURL)
+			}
 		}
 	}
 }
--- a/v3/pkg/events/events.go
+++ b/v3/pkg/events/events.go
@@ -485,7 +485,7 @@ func newWindowsEvents() windowsEvents {
 	}
 }
 
-var iOS = newIOSEvents()
+var IOS = newIOSEvents()
 
 type iosEvents struct {
 	ApplicationDidBecomeActive             ApplicationEventType
@@ -536,6 +536,18 @@ func newIOSEvents() iosEvents {
 		WebViewDidFinishNavigation:             1264,
 		WebViewDidFailNavigation:               1265,
 		WebViewDecidePolicyForNavigationAction: 1266,
+	}
+}
+
+var Android = newAndroidEvents()
+
+type androidEvents struct {
+	ActivityCreated ApplicationEventType
+}
+
+func newAndroidEvents() androidEvents {
+	return androidEvents{
+		ActivityCreated: 1300,
 	}
 }
 


mgueregath:
Added one more fix to this branch (pushed as a follow-up commit) while actually installing and running the iOS build on a physical device, not just the simulator.

**`CFBundleExecutable` is hardcoded to `"wailsapp"` in both `Info.plist.tmpl` and `Info.dev.plist.tmpl`.** They render it from `{{.BinaryName}}`, but `ios_xcode_gen.go` never wires that field to anything config-derived — see the comment right above it: `// BinaryName remains default unless we later add config support`. It's permanently the literal default, regardless of the app's actual product name.

This breaks real-device installs outright:

```
The item at <name>.app is not a valid bundle.
Failure Reason: The path to the provided bundle's main executable could not be determined.
```

`xcrun devicectl device install app` (and Xcode's own Run) can't find an executable named `wailsapp` inside the bundle, because the actual binary is named after `PRODUCT_NAME`/`EXECUTABLE_NAME` (correctly derived from `build/config.yml`) — just the `Info.plist` disagreed with what's actually in the bundle.

Fixed by switching both templates to Xcode's own `$(EXECUTABLE_NAME)` build variable instead of a Go-side field, so it can never drift out of sync with what the linker actually produces — same approach any Xcode-generated project uses. Verified: installed and launched a real app on a physical iPhone (`xcrun devicectl device install app` + `device process launch`, both succeeded) after this fix; failed with the error above before it.

mgueregath:
Two more fixes on this branch, both found while actually getting a real app to render (not just compile) on iOS Simulator and an Android emulator — both were stuck on a permanently blank/white screen even though everything else (bindings, native bridge handshake) worked.

## iOS: `run()` never calls `setURL()`

`iosWebviewWindow.run()` (`webview_window_ios.go`) creates the native WKWebView (`ios_create_webview_with_id`) and applies the background colour, but never tells it to navigate anywhere. Every other platform's `run()` does this — see `webview_window_darwin.go`, which resolves `options.URL` via `assetserver.GetStartURL()` and calls `w.setURL()` right after creating the native window. iOS's `setURL()` itself is fine (mirrors macOS's), it's just never invoked from `run()`.

Symptom: the WKWebView gets created (so `BackgroundColour` shows) but sits permanently blank — confirmed via full `log stream` capture that `AssetFileServerFS` is serving `index.html`/`style.css`/the JS bundle successfully and the JS runtime even sends `wails:runtime:ready`, but nothing ever called `loadRequest` on the webview, so there was never any page to begin with.

Fix: mirror macOS's `run()` — resolve the start URL and call `setURL()` immediately after creating the native handle.

## Not a library bug, but related: `examples/android/main_android.go`'s placement matters and isn't obvious

Separate from the two PRs' actual library changes — while chasing an analogous blank-screen symptom on Android, found that `RegisterAndroidMain(main)` must be called from an `init()` in a file that lives in the **project root** (same directory as `main.go`), matching `examples/android/main_android.go`. If that registration never runs, `signalAppReady()` is never called, and every single asset request blocks for the full `waitForAppReady(10 * time.Second)` timeout in `application_android.go` before failing — same "looks like it's working but the page never loads" symptom as the iOS bug above, just with a 10s stall first. Not filing this as a library bug since the mechanism is intentional and the example does show the right placement — just flagging it here since it cost real time to track down, in case a `wails3 android` scaffolding/doctor check ever wants to catch a misplaced `main_android.go`.

Verified both platforms end-to-end after the `setURL` fix: built, installed, and launched real apps — iOS Simulator and an arm64-v8a Android emulator both now render the actual app UI (buttons, styling, everything), not a blank page. Screenshots taken to confirm, not just log inspection.

mgueregath:
Closing this — while working through the merge conflicts I found that #5562/#5602 (merged since this PR was opened against `v3.0.0-alpha.98`) already fix everything reported here:

- `linux_cgo.c`/`.h` GTK build tags now correctly exclude Android.
- `messageprocessor_ios.go`/`messageprocessor_android.go` both define their empty stub maps.
- `pkg/events` now has a full `Android` event set.
- `application_android.go`'s JNI page-finished handler no longer manually calls `runtime.Core()` at all — it was rewritten to forward a proper window event instead.
- iOS's `run()` now calls `setURL()` after creating the native handle (same fix I'd made here, apparently done independently).

Verified directly: built and ran real apps on both platforms against `v3.0.0-beta.4` with **none** of this branch's patches applied, and both worked. Filed a follow-up issue for one remaining real bug found in the process (missing `-ObjC` linker flag in the generated Xcode project — separate from anything in this PR).

Thanks for the mobile rewrite — sorry for the noise on an already-fixed issue.

PR #5914, opened 2026-08-06
Fix/window not found noise

julianstorer:
# Description

`handleWindowEvent` logs a warning whenever a queued window event refers to a
window that's no longer in the manager. That lookup miss is a normal part of
the window lifecycle, not a fault:

- The default `Common.WindowClosing` listener removes the window from the
  manager immediately, but AppKit keeps posting `windowWillClose`,
  `windowDidResignKey` etc. for the same `NSWindow`. Because
  `events_darwin.go`'s `hasListeners` is a placeholder that always returns
  true, each of those is queued and then warns after the window is already gone.
- `App.cleanup` tears every window down and nils `a.windows`, so any in-flight
  events hit the same miss.

The result is a warning on every window close in a multi-window app and on every
quit, none of it actionable. This demotes it to a debug log, keeping the miss
visible when debugging without spamming normal output.

## Type of change

# How Has This Been Tested?
Opened and closed windows in a multi-window app and quit the app; confirmed the
warning no longer appears at the default log level.
- [ ] Windows
- [x] macOS 
- [ ] Linux

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced log severity for unrecognized window events, preventing unnecessary warning messages while preserving diagnostic details in debug logs.


leaanthony (approved):

leaanthony:
Thanks for your many contributions! :pray:

leaanthony (approved):

PR #5920, opened 2026-08-06
[WEP] keep WebViews running while hidden

julianstorer:
Add an opt-in per-window preference that stops the platform WebView from throttling JavaScript while the window is hidden

FYI, here's my reference implementation:
https://github.com/juggler-ai/wails/tree/feat/mac-dock-bounce-on-flash


leaanthony:
Will be reviewing this post v3 release :+1:

PR #5924, opened 2026-08-07
[v2] Fix nil pointer crash and swallowed error when downloading the WebView2 bootstrapper

jannskiee:
# Description

`go vet` already flags this on current `master`:

```
$ cd v2 && go vet ./internal/webview2runtime/
internal/webview2runtime/webview2runtime.go:57:8: using resp before checking for errors
```

In `downloadBootstrapper` (`v2/internal/webview2runtime/webview2runtime.go`, lines 46-68) two `defer` statements sit above their own error checks, and the transport error is then overwritten. The net effect is that when the WebView2 bootstrapper cannot be downloaded, the app dies at line 57 instead of showing the error dialog the caller already implements.

No linked issue: I found this by code inspection rather than from a bug report, so I am raising it directly as a PR.

This is v2 only. `downloadBootstrapper` exists in exactly one file repo-wide, and v3 sidesteps the whole path by embedding the bootstrapper at package time (`v3/internal/commands/generate_webview2.go`), so there is no v3 fix to port.

### Why it bites

`v2/internal/wv2installer/download.go` is the default installation strategy (build tag `windows && !wv2runtime.error && !wv2runtime.browser && !wv2runtime.embed`), reached from `PreflightChecks` before any window exists. It already handles the error properly:

```go
installedCorrectly, err := webview2runtime.InstallUsingBootstrapper()
if err != nil {
    _ = webview2runtime.Error(err.Error(), messages.Error)   // download.go:27
    return err
}
```

That native error MessageBox is unreachable today on this path. The user gets the "WebView2 installation required" prompt, clicks OK, and the process then disappears with no further output. Because the binary is linked `-H windowsgui` there is no console, so the panic trace goes nowhere visible.

This PR does not add error handling. It makes the existing error handling reachable.

### The three lines

**1. `defer resp.Body.Close()` at line 57 panics at the defer statement itself.** Per the [spec](https://go.dev/ref/spec#Defer_statements), "Each time a `defer` statement executes, the function value and parameters to the call are evaluated as usual and saved anew". Evaluating the operand `resp.Body` loads a field from the nil `*http.Response`, so it panics there, before the `if err != nil` on line 58 is reached. Confirmed by stack frame: the panicking frame points at the defer line, and a marker statement placed after it never executes.

**2. `err = out.Close()` at line 59 discards the transport error.** Currently unreachable, since line 57 panics first, but it means moving only the defer is not enough. The assignment unconditionally throws away the real cause: if `Close` returns nil you get `("", nil)`, and the caller then runs `runInstaller("")` and surfaces a misleading `exec: no command` dialog instead of the network error. If `Close` returns anything else you get an unrelated error. Either way the actual failure is destroyed.

**3. `defer out.Close()` at line 52** has the same shape but is harmless, because `(*os.File).Close` is nil-receiver safe (`os/file_posix.go`: `if f == nil { return ErrInvalid }`). Included only so the function is internally consistent. Cosmetic, not a bug.

Introduced in 045e58778 (2021-12-27) and untouched since. It survived later sweeps because the file is `//go:build windows` while CI vets on Linux, and because `go test`'s built-in vet subset (`atomic, bool, buildtags, directive, errorsas, ifaceassert, nilfunc, printf, stringintconv, tests`) excludes `httpresponse`.

## Type of change

## The change

```diff
 	// Download installer
 	out, err := os.Create(installer)
-	defer out.Close()
 	if err != nil {
 		return "", err
 	}
+	defer out.Close()
+
 	resp, err := http.Get(bootstrapperURL)
-	defer resp.Body.Close()
 	if err != nil {
-		err = out.Close()
 		return "", err
 	}
+	defer resp.Body.Close()
+
 	_, err = io.Copy(out, resp.Body)
```

No nil guard is needed on either defer: on a transport failure `http.Get` returns a nil `*http.Response`, so placing the defer after the error check is sufficient and idiomatic. No double close is introduced, since removing the explicit `out.Close()` leaves each defer firing exactly once.

# How Has This Been Tested?

- [x] Windows
- [ ] macOS
- [ ] Linux

I found this by code inspection while packaging a Wails v2 Windows app, and have not hit it in the field, which is why the `wails doctor` output below shows WebView2 present. I reproduced the failure mode in isolation instead.

`go vet ./internal/webview2runtime/` goes from the diagnostic above to clean, and `go build` and `gofmt -l` are both clean after the change. I confirmed the diagnostic comes from the `httpresponse` analyzer: `go vet -httpresponse` reproduces it and `go vet -httpresponse=false` is silent.

Behaviour, using `http://127.0.0.1:1/` in place of the Microsoft URL so the failure is deterministic and needs no network:

| variant | result |
| --- | --- |
| current `master` | `panic: runtime error: invalid memory address or nil pointer dereference`, frame at the `defer` line |
| defer moved, `err = out.Close()` kept | returns `("", nil)`: no error, empty path |
| this PR | returns `Get "http://127.0.0.1:1/": dial tcp 127.0.0.1:1: connectex: No connection could be made because the target machine actively refused it.` |

Standalone reproducer (`go run .`, no network required):

```go
package main

import (
	"fmt"
	"net/http"
)

func main() {
	resp, err := http.Get("http://127.0.0.1:1/")
	defer resp.Body.Close() // panics here
	fmt.Println("never reached")
	if err != nil {
		return
	}
	_ = resp
}
```

I also confirmed the success path still writes a complete file and that the handle is closed by the time `InstallUsingBootstrapper` calls `runInstaller`, by reading the file back and removing it immediately after the function returns (on Windows the `os.Remove` would fail with a sharing violation had a write handle still been open).

## Test Configuration

```
# Wails
Version | v2.12.0

# System
| OS           | Windows 10 Pro                                             |
| Version      | 2009 (Build: 26200)                                        |
| ID           | 25H2                                                       |
| Branding     | Windows 11 Pro                                             |
| Go Version   | go1.26.3                                                   |
| Platform     | windows                                                    |
| Architecture | amd64                                                      |
| CPU          | AMD Ryzen 7 5700X 8-Core Processor                         |
| GPU          | NVIDIA GeForce RTX 3060 Ti (NVIDIA) - Driver: 32.0.15.9186 |
| Memory       | 16GB                                                       |

# Dependencies
| WebView2   | N/A          | Installed | 151.0.4129.59 |
| Nodejs     | N/A          | Installed | 22.18.0       |
| npm        | N/A          | Installed | 11.10.1       |
| *nsis      | N/A          | Installed | v3.12         |
```

# Checklist:

- [x] (v2 only) I have updated `website/src/pages/changelog.mdx` with details of this PR (v3 changelog entries are added automatically)
- [x] My changes generate no new warnings (it removes one)

On the last two: `internal/webview2runtime` has no test files today (`go test` reports `[no test files]`), the whole file is `//go:build windows`, and `downloadBootstrapper` is unexported with a hardcoded URL and no injection point, so a real test needs a signature change I deliberately kept out of a bug fix. Happy to add one if you would like it, here or as a follow-up.

Two related things I noticed and deliberately left alone, to keep this reviewable. Say the word and I will open separate PRs:

1. A failed download leaves a zero byte `MicrosoftEdgeWebview2Setup.exe` in `%TEMP%`. It does not accumulate, since the path is fixed and `os.Create` truncates.
2. In `runInstaller` in the same file, if `cmd.Wait()` returns a non-`*exec.ExitError`, or `Sys()` is not a `syscall.WaitStatus`, control falls through both `if` blocks to `return true, nil` and reports a successful install on a failure.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed an issue that could cause the app to exit unexpectedly while downloading the WebView2 bootstrapper.
  - Installation errors are now displayed correctly when the download fails.
  - Improved error handling to preserve the original network or file-creation error.
  - Installer cleanup failures are now reported when no earlier error has occurred.

- **Documentation**
  - Added a changelog entry describing the WebView2 bootstrapper download fixes.


jannskiee:
Good catch, thanks. Addressed in c030327.

`downloadBootstrapper` now uses named returns and captures the deferred `out.Close()` error, but only when nothing has already failed, so a transport or `io.Copy` error is never masked by a later close problem. That matters here specifically because the caller executes the file this writes, so a failed flush must not come back as a usable path.

I checked all four combinations before pushing:

| earlier error | close fails | result |
| --- | --- | --- |
| no | no | `(installer, nil)` |
| no | yes | `("", closeErr)` |
| yes | no | `("", earlierErr)` |
| yes | yes | `("", earlierErr)` |

`go vet ./internal/webview2runtime/`, `go build`, and `gofmt` are all still clean.

leaanthony (approved):
Nice catch!

PR #5942, opened 2026-08-10
feat(v3): streams — WebSocket semantics without a listening socket

leaanthony:
# Description

Adds **streams**: a named, ordered, bidirectional byte channel between Go and the frontend
with the WebSocket programming model, and **no listening socket**.

A WebSocket cannot be spoken over a custom URL scheme, so the only way to get one inside a
webview today is to bind a real TCP port. For a desktop app that means an open local port
reachable by any other process on the machine, needing an origin check and a token to be
safe, and visible to every firewall and endpoint-security product the user runs. Apps do
this anyway, because there was no other way to push a continuous feed to the frontend.
This removes the need.

```go
app.HandleStream("telemetry", func(c *application.StreamConn) {
    defer c.Close()
    for {
        frame, err := c.Receive()   // blocks; err on reload, close, or shutdown
        if err != nil {
            return
        }
        c.Send(reply(frame))        // blocks like a socket write
    }
})
```

```js
const s = Stream("telemetry");      // synchronous, like new WebSocket(url)
s.onmessage = (ev) => decode(ev.data);
s.send(bytes);
```

`Stream(name)` returns synchronously with `readyState === CONNECTING`, so connections can
be created at module scope — the shape generated bindings would use. `JSONStream` plus
`SendJSON`/`ReceiveJSON` give the same thing in objects rather than bytes.

## How it works

Go→JS is a **held poll** over the asset server: the request parks until a frame exists, so
delivery latency is ~0 with no polling interval and nothing adaptive. Frames arriving while
a response is in flight ride the next one, which makes the round trip itself the batching
window — it widens under load without anything measuring it. JS→Go is an ordinary POST per
frame. One poll in flight per window multiplexes every connection on that page.

The handler goroutine's lifetime is the connection's lifetime, exactly like a
gorilla/coder WebSocket handler, so reload, shutdown and cleanup fall out of that single
choice rather than each needing a policy.

Server builds (`-tags server`) swap the transport for a **real WebSocket**, since there is
already a listener to upgrade on. `HandleStream` and `StreamConn` are shared verbatim;
only the sink differs.

Design decisions carry over the lessons from the recent event-transport work, and each is
commented where it lives: nothing in the Go→JS path touches the main thread (that
inverted 4.4% of events), nothing reaches `evaluateJavaScript` at any size (that retained
11.6 GB on macOS at 100 × 1 MB/s), control data travels in headers rather than bodies, and
every buffer is bounded and dropped with its window.

## Type of change

- [x] This change requires a documentation update *(included)*

Nothing existing changes behaviour. Events, `Emit`, and the runtime call path are
untouched.

# How Has This Been Tested?

Unit tests (`go test ./pkg/application/ -run TestStream -race`) cover the protocol,
ordering under eight concurrent senders, backpressure and both caps, poll supersede,
session close, chunk reassembly, and the connection-refusal path. Server mode has its own
suite under `-tags server`.

A load harness is included at `v3/tests/stream-performance` (`-upload`, `-reloads`, and
scenario sweeps). Across the three platforms, **~41 million frames with 0 drops and 0
reorders in every Go→JS scenario**.

| | Go→JS peak | JS→Go peak |
|---|---:|---:|
| macOS / WebKit-Cocoa | 2,141 MB/s | 1,507 MB/s |
| Linux / WebKitGTK | 226 MB/s | 727 MB/s |
| Windows / WebView2 | 100 MB/s | 99 MB/s |

Small frames go far faster downward than upward — 634,000 frames/s vs ~6,200/s on macOS —
because a poll response coalesces up to 256 frames while every JS→Go frame is its own
request. Latency does not degrade with load: 20,000 frames/s measured a *lower* p99 than
100 frames/s.

Also verified: reload closes the connection promptly (6 reloads → 7 connects, 6
disconnects, never more than one live), `wails3 dev` with a Vite dev server in front of the
asset server, and the included example end to end.

- [x] Windows
- [x] macOS
- [x] Linux

## Test Configuration

- macOS 26.4.1 arm64, WebKit-Cocoa, Go 1.26.2 — full sweeps, reload check, server-mode
  tests, `wails3 dev`, example
- Ubuntu (kernel 7.0.0-29) x86_64, WebKitGTK 2.52.3 under Xvfb with software GL, Go 1.26.2
  — both directions
- Windows 11 26200 x86_64, WebView2 151.0.4129.72, MinGW-w64 15.2.0, Go 1.26.2 — both
  directions, run in the interactive console session

The three machines are not comparable hardware, so the shape of each platform's curve is
the meaningful comparison rather than one platform's MB/s against another's.

# Known gaps

Documented in the internals guide rather than hidden:

- **The platform layer does not report a cancelled request**, so a poll belonging to a page
  that navigated away stays parked until the hold expires. The *connection* still closes
  promptly, because a new session for a window supersedes the old one. Plumbing
  `stopURLSchemeTask` and its equivalents through to a `context.CancelFunc` is the
  outstanding item.
- Buffer sizes are compile-time constants, not options.
- No typed streams; frames are `[]byte` by decision.
- Linux and Windows both plateau on the Go→JS path (~213 and ~90 MB/s) where macOS keeps
  scaling. Each has a candidate mechanism in its response writer — a pipe on Linux,
  whole-body buffering plus a UI-thread hop on Windows — but both are hypotheses, not
  findings.
- `InitialHTML` windows cannot use streams: they load with `origin === "null"` and cannot
  reach the asset server.

# Checklist:

- [ ] (v2 only) I have updated `website/src/pages/changelog.mdx` with details of this PR

Documentation added: a user guide, an internals page written for whoever changes this code
next, a mechanical WebSocket-to-Streams migration guide, and a minimal example under
`v3/examples/streams`. `AGENTS.md` gains a pointer to the internals page and a section on
the two frontend runtime build outputs, which are easy to confuse.

## Summary by CodeRabbit

* **New Features**
  * Added bidirectional Go–JavaScript Streams for desktop and server applications.
  * Added binary and JSON stream APIs with lifecycle, buffering, backpressure, and error handling.
  * Added WebSocket-compatible runtime support and server-side WebSocket transport.
  * Added configurable WebSocket origin allowlisting for server applications.
  * Added a Streams example and cross-platform performance testing tools.

* **Documentation**
  * Added guides for Streams, transport internals, WebSocket migration, and runtime installation and build workflows.


v3/pkg/application/stream_server.go
```diff
@@ -0,0 +1,126 @@
...
+		cancel: cancel,
+	}
+	c.inCond = sync.NewCond(&c.inMu)
+
+	// Read pump. A read error is the socket closing, which is what ends the
+	// handler — the same signal a reload gives the desktop build.
+	go func() {
+		defer c.shutdown()
+		for {
+			_, data, err := conn.Read(ctx)
```
leaanthony: **[P1] Raise the server-mode WebSocket read limit**

This comment applies only to the actual `coder/websocket` transport used under `-tags server`. `coder/websocket` v1.8.14 defaults to a 32,768-byte message read limit, and this connection never calls `SetReadLimit`. A frontend-to-Go frame over 32 KiB therefore closes the server-mode connection even though Streams documents a 64 MB JS→Go frame limit. Please set an intentional limit after `Accept` (presumably `streamMaxSendBytes`) and add a large client-to-server test; the current large-frame test only exercises Go→client traffic.

v3/pkg/application/stream_session.go
```diff
@@ -0,0 +1,301 @@
...
+		return
+	}
+	s.conns[connID] = c
+	s.mu.Unlock()
+
+	_ = s.enqueue(c, connID, frameOpen, nil, false)
+
+	go func() {
+		defer handlePanic()
+		defer c.shutdown()
```
leaanthony: **[P1] Notify the frontend when a desktop handler returns**

The wrapper defers `shutdown()`, which cancels and unregisters the Go connection but does not enqueue `frameClose`. Consequently, a handler that simply returns—as the Streams API documentation explicitly permits—leaves the frontend `WailsSocket` open and `onclose` never fires. Because the frontend keeps polling, session TTL does not repair this. The handler-return path needs to perform the peer notification promised by the lifecycle contract.

v3/pkg/application/stream.go
```diff
@@ -0,0 +1,428 @@
...
+	c.inMu.Lock()
+	c.inCond.Broadcast()
+	c.inMu.Unlock()
+	c.sink.wake()
+}
+
+// deliver queues an inbound frame and wakes a waiting Receive.
+func (c *StreamConn) deliver(data []byte) {
+	c.inMu.Lock()
+	c.in = append(c.in, data)
```
leaanthony: **[P1] Bound the inbound Streams queue**

`deliver` always appends to `c.in` without a byte or depth limit. The desktop send endpoint responds after appending and can accept the next POST, while the server transport's read pump immediately reads again. A slow or non-reading handler therefore allows frontend traffic to grow host memory without bound. This conflicts with the feature's stated bounded-memory design; producer progress should be tied to `Receive` freeing space in a bounded per-connection queue.

v3/pkg/application/stream_session.go
```diff
@@ -0,0 +1,301 @@
...
+		// The frontend allocates connection ids; a duplicate means a buggy or
+		// hostile client. Refuse rather than replacing a live connection.
+		s.mu.Unlock()
+		cancel()
+		return
+	}
+	s.conns[connID] = c
+	s.mu.Unlock()
+
+	_ = s.enqueue(c, connID, frameOpen, nil, false)
```
leaanthony: **[P1] Do not discard the open acknowledgement under backpressure**

If another connection has filled the shared session queue, this nonblocking enqueue returns `ErrStreamFull` and the error is ignored. The handler still starts, but the frontend remains `CONNECTING` forever because the acknowledgement is never retried. Refusal and close control frames have the same best-effort loss risk. Control frames need reserved capacity or another reliable delivery path independent of ordinary data backpressure.

v3/pkg/application/stream_server.go
```diff
@@ -0,0 +1,126 @@
...
+	w.mu.Lock()
+	defer w.mu.Unlock()
+
+	if w.closed {
+		return ErrStreamClosed
+	}
+
+	switch kind {
+	case frameData:
+		if err := w.conn.Write(w.ctx, websocket.MessageBinary, data); err != nil {
```
leaanthony: **[P1] Preserve the documented `TrySend` contract in server mode**

The sink ignores the `block` argument, and `coder/websocket.Conn.Write` can block until the peer progresses or the connection closes. Thus `StreamConn.TrySend` is nonblocking on the desktop transport but potentially blocking under `-tags server`. This matters for the migration guide's fan-out and broker-callback examples, which select `TrySend` specifically to avoid stalling those producers. Either provide bounded queued writes here or document/expose a genuinely different server-mode contract.

v3/pkg/application/stream_session.go
```diff
@@ -0,0 +1,301 @@
...
+			(len(s.out) == 0 || s.outBytes+len(data) <= streamOutQueueBytes) {
+			break
+		}
+		if !block {
+			return ErrStreamFull
+		}
+		s.space.Wait()
+	}
+
+	s.out = append(s.out, outFrame{connID: connID, kind: kind, data: data})
```
leaanthony: **[P1] Snapshot the Go payload before `Send` reports acceptance**

The queued frame retains the caller's `[]byte` directly, while `Send` returns before a poll encodes the frame. A caller that reuses the buffer or returns it to a pool can therefore alter a frame already reported as accepted. Please copy the bytes while accepting the frame, or explicitly define ownership/lifetime semantics that require the caller to retain immutable storage until delivery.

v3/internal/runtime/desktop/@wailsio/runtime/src/stream.ts
```diff
@@ -0,0 +1,466 @@
...
+    send(data: string | ArrayBufferLike | ArrayBufferView | Blob): void {
+        if (this.readyState === WailsSocket.CONNECTING) {
+            // Matches the WebSocket specification.
+            throw new DOMException("Still in CONNECTING state.", "InvalidStateError");
+        }
+        if (this.readyState !== WailsSocket.OPEN) {
+            return;
+        }
+
+        this._chain = this._chain.then(async () => {
```
leaanthony, line 325: **[P2] Snapshot JavaScript buffers when `send` is called**

Conversion is deferred into the promise chain, and `toBytes` later returns a view over the original `ArrayBuffer`. Mutating or reusing a typed array immediately after `send()` can therefore change the eventual Streams frame; several queued sends using one reusable buffer can all observe its final contents. Snapshot the supported mutable inputs at the API boundary so queued frames have stable ownership.

v3/pkg/application/stream_transport.go
```diff
@@ -0,0 +1,388 @@
...
+		c.items[id] = set
+	}
+	if set.total != total {
+		delete(c.items, id)
+		return nil, false
+	}
+	if _, dup := set.parts[index]; dup {
+		return nil, false
+	}
+	if c.bytes+len(data) > streamMaxSendBytes {
```
leaanthony: **[P2] Distinguish rejected chunk sets from incomplete frames**

When aggregate data exceeds 64 MB, totals conflict, or a duplicate arrives, `add` returns the same `false` value used for a valid but incomplete frame. The HTTP layer consequently responds 204, so the JavaScript send chain completes while the frame is silently discarded. The inconsistent-total path also deletes the set without subtracting `set.size` from `c.bytes`, permanently consuming capacity for the session. Return a distinct error and a non-2xx response for rejected sets.

v3/internal/runtime/desktop/@wailsio/runtime/src/stream.ts
```diff
@@ -0,0 +1,466 @@
...
+ * ```js
+ * const s = JSONStream("telemetry");
+ * s.onmessage = (ev) => console.log(ev.data.temperature);
+ * s.onopen    = () => s.send({ subscribe: "sensors" });
+ * ```
+ *
+ * A frame that is not valid JSON raises an `error` event and is dropped, rather
+ * than throwing from the poll loop and taking the connection down with it.
+ */
+export function JSONStream(name: string): WailsSocket | WebSocket {
```
leaanthony: **[P2] Give `JSONStream` an object-stream TypeScript type**

`JSONStream` is declared as `WailsSocket | WebSocket`, whose `send` method does not accept objects. TypeScript consumers therefore cannot compile the documented `s.send({ ... })` usage even though runtime code replaces `send`. A dedicated return interface should expose `send(value: unknown)` and the decoded message type.

v3/internal/runtime/desktop/@wailsio/runtime/src/stream.ts
```diff
@@ -0,0 +1,466 @@
...
+    // handler only ever sees decoded objects.
+    let handler: ((ev: MessageEvent) => void) | null = null;
+    Object.defineProperty(socket, "onmessage", {
+        get: () => handler,
+        set(fn) { handler = typeof fn === "function" ? fn : null; },
+        configurable: true,
+        enumerable: true,
+    });
+
+    socket.addEventListener("message", (ev: Event) => {
```
leaanthony: **[P2] Make JSON decoding consistent for event listeners**

Only the `onmessage` property is intercepted. A consumer using the advertised `addEventListener("message", ...)` API receives the original `ArrayBuffer`, while `onmessage` receives the parsed object. `JSONStream` should expose decoded message data consistently through both listener styles.

v3/pkg/application/stream_transport.go
```diff
@@ -0,0 +1,388 @@
...
+	}
+	if windowID != 0 && s.windowID != windowID {
+		http.Error(rw, "session belongs to another window", http.StatusForbidden)
+		return nil
+	}
+	return s
+}
+
+func (a *App) serveStreamPoll(rw http.ResponseWriter, req *http.Request) {
+	if req.Method != http.MethodGet && req.Method != http.MethodHead {
```
leaanthony: **[P2] Do not let `HEAD` polls drain queued frames**

`HEAD` follows the normal `awaitFrames`/drain path. HTTP suppresses the response body, but the frames have already been removed from the session queue, causing data loss. Restrict this protocol endpoint to `GET`, or handle `HEAD` without creating a session or consuming state.

leaanthony (commented):
Stream review findings from a local implementation trace and targeted test pass. These comments evaluate the documented Wails Streams contracts; they do not assume that the desktop held-poll/POST transport is a WebSocket implementation. The one coder/websocket-specific comment is limited to `-tags server`.

v3/pkg/application/stream.go
```diff
@@ -0,0 +1,514 @@
...
+func (m *streamManager) session(id string, windowID uint, maySupersede bool) *streamSession {
+	m.mu.Lock()
+	if m.closed {
+		m.mu.Unlock()
+		return nil
+	}
+
+	s, ok := m.sessions[id]
+	var superseded []*streamSession
+	if !ok && maySupersede {
```
taliesin-ai: **[P1] Supersede old sessions when an open POST wins the race**

The constructor starts the open POST and the first poll concurrently, so the POST can reach Go first and create the new session with `maySupersede=false`. When the poll arrives, `ok` is already true and this condition is skipped, leaving the previous session and its handlers alive for up to the ten-minute live-connection grace period. Please supersede other sessions for the same window whenever a poll arrives, even if the current session id was already created, and cover the open-before-poll ordering in a test.

v3/pkg/application/stream_server.go
```diff
@@ -0,0 +1,229 @@
...
+		frame := w.out[0]
+		w.out[0] = outFrame{}
+		w.out = w.out[1:]
+		w.bytes -= len(frame.data)
+		w.cond.Broadcast()
+		w.mu.Unlock()
+
+		switch frame.kind {
+		case frameData:
+			if err := w.conn.Write(w.ctx, websocket.MessageBinary, frame.data); err != nil {
```
taliesin-ai: **[P1] Flush accepted WebSocket writes before cancelling the writer**

The WebSocket writer uses `w.ctx`, but `StreamConn.Close` queues the close frame and immediately calls `shutdown`, which cancels that same context. Any data accepted by a preceding `Send` can therefore fail here before it reaches the peer. A targeted handler using `defer c.Close(); c.Send(final); return` produced EOF without the final frame in 20/20 runs. Please give the writer an independent lifetime or wait for the queued data and close frame to flush before cancelling it.

v3/internal/runtime/desktop/@wailsio/runtime/src/stream.ts
```diff
@@ -0,0 +1,582 @@
...
+// provided whatever delay was appropriate.
+const BACKOFF_MIN = 250;
+const BACKOFF_MAX = 5000;
+
+function streamURL(endpoint: string): string {
+    return window.location.origin + "/wails/stream/" + endpoint;
+}
+
+let nextConnID = 1;
+const connections = new Map<number, WailsSocket>();
```
taliesin-ai: **[P1] Share the connection registry across runtime instances**

Only the session id is page-global; `nextConnID`, `connections`, and `pollRunning` remain local to each evaluated runtime module. When the injected runtime and the npm runtime coexist, both allocate connection id 1 and run competing polls against the same session. A duplicate open can stay in `CONNECTING`, and either poll can drain and discard frames whose id exists only in the other module map. Please place the connection-id allocator, registry, and polling state on the same page-global object as the session id, or otherwise enforce a single runtime instance.

v3/internal/runtime/desktop/@wailsio/runtime/src/stream.ts
```diff
@@ -0,0 +1,582 @@
...
+ * A frame that is not valid JSON raises an `error` event and is dropped, rather
+ * than throwing from the poll loop and taking the connection down with it.
+ */
+export function JSONStream(name: string): JSONSocket {
+    const socket = Stream(name) as WailsSocket;
+    const decoder = new TextDecoder();
+
+    // Decoding here rather than by wrapping onmessage means addEventListener
+    // listeners see the parsed object too.
+    socket._decode = (payload) => JSON.parse(decoder.decode(payload));
```
taliesin-ai: **[P1] Decode JSON messages from the native server WebSocket**

Under `-tags server`, `Stream(name)` returns a native `WebSocket`. Assigning `_decode` here only affects `WailsSocket._message`; native WebSocket dispatch never consults it, so both `onmessage` and `addEventListener("message", ...)` receive the raw `ArrayBuffer` instead of the parsed object. This breaks the documented transport-independent `JSONStream` contract. Please wrap native message dispatch or return a transport-neutral wrapper that performs the decode.

v3/internal/runtime/desktop/@wailsio/runtime/src/stream.ts
```diff
@@ -0,0 +1,582 @@
...
+        this._pending.push(snapshot);
+        if (this._flushing) return;
+        this._flushing = true;
+
+        this._chain = this._chain.then(async () => {
+            try {
+                while (this._pending.length > 0) {
+                    const batch = await Promise.all(this._pending.splice(0));
+                    const total = batch.reduce((n, b) => n + b.byteLength, 0);
+                    this._buffered += total;
```
taliesin-ai: **[P2] Count pending sends in bufferedAmount**

`_buffered` is incremented only after the pending promises are collected inside the asynchronous chain. Immediately after `send()`, and while additional frames wait behind an in-flight batch, `bufferedAmount` can therefore report zero even though bytes are queued. Applications using the documented WebSocket-style value for producer backpressure will over-send. Track pending bytes synchronously at the `send()` boundary and subtract them only after delivery or failure.

taliesin-ai (commented):
Additional findings against the current head after targeted lifecycle, server-transport, and runtime checks.

v3/pkg/application/stream.go
taliesin-ai: Fixed in `a8ee8774f`. You are right about the ordering: the constructor starts the open POST and the first poll concurrently, so the POST can create the session and the poll then finds `ok` already true.

Supersede now runs on **every** poll rather than only when the poll creates the session. One consequence worth noting — the sweep can now match the session being polled for, which it never could before, so it skips it explicitly.

Covered by `TestStreamPollSupersedesWhenOpenWonTheRace`, which drives exactly that ordering: session created by a send (asserting a send does *not* supersede), then polled, asserting the previous page's connection is retired and the polling session survives.

v3/internal/runtime/desktop/@wailsio/runtime/src/stream.ts
taliesin-ai: Fixed in `a8ee8774f`. Correct, and making only the session id page-global was half a fix — the half I left behind was the more damaging one.

`nextConnID`, `connections` and the polling flag now live on the same `window._wails` object as the session, so two instances share one allocator, one registry and one poll loop.

v3/internal/runtime/desktop/@wailsio/runtime/src/stream.ts
taliesin-ai: Fixed in `a8ee8774f`. Confirmed — `_decode` is only consulted by `WailsSocket._message`, so under `-tags server` the native socket delivered raw `ArrayBuffer`s and the transport-independent contract did not hold.

Both `addEventListener` and `onmessage` are now wrapped on the native instance, so `JSONStream` means the same thing on either transport. That was the whole point of sharing the API, so thank you for catching it.

v3/internal/runtime/desktop/@wailsio/runtime/src/stream.ts
taliesin-ai: Fixed in `a8ee8774f`. Agreed, and it got worse when batching landed, since frames now wait behind an in-flight batch rather than going out individually.

Everything except a `Blob` is now sized synchronously and counted at the `send()` boundary; a `Blob`, whose length is not known then, is added when it resolves. The synchronous path also removes a promise allocation per send, which suits the low-allocation goal for this transport.

v3/pkg/application/stream_server.go
taliesin-ai: **Confirmed and reproduced, but not fixed.** I am leaving this open rather than closing it with something that does not work.

I wrote your exact shape as a test — `defer c.Close(); c.Send(final)`, 20 iterations, asserting the frame arrives — and it fails, so the finding stands independently.

Three attempts, none of which made the frame arrive:

1. **Independent writer lifetime.** Gave the sink `context.Background()` instead of the connection context, added a `drained` channel, and made `serveStreamWS` wait for the pump before tearing the socket down.
2. **Separated the close flags.** `shut()` early-returned on `w.closed`, which `removeConn` had already set during `Close()`, so the graceful WebSocket close never ran at all. Split into a `shutOnce` guarding the socket close.
3. **Stopped forcing `CloseNow()`** after a graceful close, on the theory it was tearing down TCP before the peer read — leaving it only as the drain-timeout escape hatch.

Each left the client seeing `EOF` before the data frame. Something in the ordering between the pump, `removeConn`, and socket teardown is still wrong and I have not isolated it.

Reverted rather than shipped half-solved, so `main`-bound code is unchanged here and the four other findings are fixed cleanly in `a8ee8774f`. Anyone picking this up: the test above is the reproduction, and my instinct is that the read-pump goroutine's `defer c.shutdown()` is racing the write path — that is the one interaction I did not instrument.

leaanthony:
Found a lifecycle edge case on `868295f86`:

1. Create a Stream in a parent page and bridge it to an iframe over `MessageChannel`.
2. The iframe posts `{type: "close"}` and immediately calls `port.close()`.
3. The parent calls `socket.close()` when that message arrives.
4. Repeat create/transfer/close several times, then wait longer than the 60s session TTL.

If closing the port races message delivery, Go never receives `KIND_CLOSE`. The connection, handler, read/write pumps, and child process remain alive after the TTL; a goroutine dump still shows `streamSession.open` and `streamSession.awaitFrames`.

The adapter should wait for close acknowledgement, but because sessions with live connections are not reaped, one lost close can retain the session indefinitely. Is that intended?

leaanthony:
Re-review at `a8ee8774f` found two remaining cases:

1. **Supersede race:** create an old session, let a new-page `KIND_OPEN` request create its session, then process a late poll from the old page. Because every poll supersedes other sessions for the window, the old poll closes the fresh connection. The new-page poll then recreates an empty session, but the original connection is already dead. A targeted manager test reproduces this ordering deterministically.

2. **Blob `bufferedAmount`:** `send(blob)` still reports zero immediately after returning because Blob conversion is asynchronous and counting happens after `arrayBuffer()` resolves. `Blob.size` is available synchronously, so large Blob sends can bypass code using `bufferedAmount` for backpressure.

taliesin-ai:
Re-review at `a8ee8774f` — one fixed, one attempted and reverted, plus an answer on the lifecycle question.

**2. Blob `bufferedAmount` — fixed in `2be8b3ef7`.** Correct, and an obvious miss on my part: `Blob.size` is synchronous even though the conversion is not. Counted at the `send()` boundary now, so a large Blob no longer reports zero while queued.

**1. Supersede race — real, not fixed.** Your reproduction is right, and it is a regression from my previous fix: making *every* poll supersede cured the open-wins-the-race case and introduced its mirror.

The fix I tried is the obvious one — give each session a creation sequence and let a poll retire only sessions older than itself. It is written and it works for the new case, but it **deadlocks `TestStreamNewSessionSupersedesPreviousInSameWindow`**, which hangs for the full test timeout rather than failing. I could not isolate that before running out of room, so I reverted it. The branch keeps the known behaviour rather than gaining a hang.

For whoever picks it up: the ordering approach is sound, the deadlock is in the interaction between `session()` collecting superseded sessions and `close()` unwinding their handlers, and the reproduction is simply running the stream suite with `-race -timeout 90s`.

**Lifecycle question from the earlier comment: it is bounded, but you are right that it is too long.** Sessions with live connections *are* reaped — at `streamSessionGrace`, 10 minutes, not the 60s TTL. So a lost `KIND_CLOSE` retains the session for up to ten minutes, not indefinitely. The long grace is deliberate: reaping a session on idle time alone killed streams mid-run under load, measured at 230,077 frames in, because a busy page can go longer than the TTL between polls.

That said, ten minutes to clean up after one lost close is a poor trade, and your adapter should not need to ack a close for the transport to behave. The better signal is separating "has not polled at all" from "is polling but busy" — a session that has never been polled since its connections opened is gone regardless of `connCount`. I have not implemented that.

Both open items are now on the PR rather than in my head.

leaanthony:
## Review of `c851e36c0..2be8b3ef7`

Self-review of the five commits from the last round. Checks run: stream suite with `-race` (green, 1.86s), `internal/debounce` 50× (green), `tsc --noEmit` on the runtime (clean), and `go build -tags server ./...` + `go vet -tags server ./v3/pkg/application/` (clean). That last one matters here — `c851e36c0` added `wakeProducers()` to the `streamSink` interface and an implementation on `wsStreamSink`, and `stream_server.go` is behind `//go:build server`, so the normal test run never compiles the one file that gained a new method.

### 1. An open POST resurrects a retired session

New, and a composition of two commits in this range rather than a fault in either alone.

[`stream_transport.go#L290`](https://github.com/wailsapp/wails/blob/2be8b3ef70de9c6604bd5c6afa60d4c5a4dfae08/v3/pkg/application/stream_transport.go#L290) calls `sessionFor(rw, req, kind == frameOpen, false)`, so an open POST still passes `create = true`. In [`session()`](https://github.com/wailsapp/wails/blob/2be8b3ef70de9c6604bd5c6afa60d4c5a4dfae08/v3/pkg/application/stream.go#L372), `maySupersede` gates only the sweep — the `if !ok { … m.sessions[id] = s }` branch below it runs regardless. A stray open POST from a page that has already been retired therefore **recreates its session**, handler goroutine and all.

`c851e36c0` then gives any session with `connCount() > 0` the 10-minute `streamSessionGrace` instead of the 60s TTL. So a reload that races its predecessor's in-flight open POST leaves a duplicate handler and its queued payloads alive for ten minutes, with no poll ever arriving to notice. The long grace was defensible when the only way to reach a live-connection session was a real page; this path manufactures one.

Repro without adding a test: `session("page-2", 1, true)` to retire page-1, then `session("page-1", 1, false)` — page-1 is back in `m.sessions`, and `open()` on it starts a handler.

### 2. The supersede race, restated precisely

Still open (flagged above), and the mechanism is the same `!ok` branch: a late poll for `page-1` finds no entry, sweeps the window — deleting **and closing** page-2 — then recreates page-1. The old page kills its own replacement.

Both findings collapse into one rule: **a session id that has been superseded must never be creatable again.** Refusing resurrection (return nil, let the poll answer 410, which the client already handles by closing everything and stopping) is what makes this different from the `seq`-ordering attempt that deadlocked `TestStreamNewSessionSupersedesPreviousInSameWindow`: it closes nothing along a new path, so the `session()`/`close()` locking that deadlocked is left untouched.

### 3. Server-mode `JSONStream` wrapper

The native-WebSocket branch in `v3/internal/runtime/desktop/@wailsio/runtime/src/stream.ts`:

- **Parses once per listener.** Two `message` listeners means two `JSON.parse` calls of the same payload, and each listener gets a *distinct* object. The `WailsSocket` path parses once through `_decode` and shares one. That is both a cost and a semantic divergence, on the branch whose entire purpose is that the two transports mean the same thing.
- **Identical listeners no longer dedupe.** Each `addEventListener("message", fn)` builds a fresh wrapper, so adding the same function twice fires it twice — a real `WebSocket` would ignore the second add. The `wrapped` WeakMap also keys on the listener and holds only the most recent wrapper, so after a double-add `removeEventListener` detaches one and strands the other.

### 4. Smaller things

- `toBytesSync` and `toBytes` each construct `new TextEncoder()` per string send. A module-level instance is free, and this is the hot path for string senders.
- The harness comment at `v3/tests/stream-performance/main.go:556` reads backwards against itself: "blocking Send … is markedly faster than spinning on TrySend (measured 46k vs 432k frames/s at 4 KB)" puts the faster number on the losing side. If those are transposed it plants a wrong figure in a measurement record this branch has been treating as authoritative.
- The `sessionID` comment deleted in `a8ee8774f` carried the rationale for the crypto-backed nanoid (no window-id header to bind against in server mode). That change is still in the tree; its justification no longer is.

### 5. Debounce (`05f89fdc6`)

The fix is right for what it targets and 50 runs are clean. Two notes:

- The window between `d.add(fn1)` and `d.mu.Lock()` is still a flake — a >2ms preemption there lets fn1's timer fire while the generation still matches, fn1 runs, and the test fails legitimately. That window was in the original test too, so this is not a regression: the commit fixed the fn2 side and left the fn1 side. The test is less flaky, not deterministic.
- The header block still describes the old sequence ("Sleep 3ms", steps 2 and 3 in the other order) against code that now sleeps 10ms — a fair nit in a commit whose whole purpose was this test.

### What the green suite does and does not say

`ok … 1.859s` with `-race` covers the Blob accounting, the reap grace, `wakeProducers`, and the open-won-the-race supersede case. `TestStreamPollSupersedesWhenOpenWonTheRace` is a good test in particular — it asserts the negative, that a *send* must not supersede.

It says nothing about the open items. Neither the supersede race nor the server-mode writer flush has a test, and the resurrection path in §1 has none either.

leaanthony:
## Correction: the review above was against a stale head

I reviewed local `2be8b3ef7` without fetching first, so I missed `c05b3c997 fix(v3): harden stream lifecycle and delivery`. Most of that comment is already answered. Re-reviewed properly this time.

**Fixed in `c05b3c997`:** the open-POST resurrection path and the supersede race both close, via a client-supplied page generation plus a per-window `retiredThrough` watermark — and, importantly, the generation is anchored to `performance.timeOrigin`, so a page that clears `sessionStorage` cannot restart at 1 and brick the window. The `JSONStream` double-add is fixed by reusing the WeakMap entry. The server-mode writer flush — the P1 that defeated three earlier attempts — is fixed properly this time (write context rooted in the app context rather than the handler's, a `done` barrier, a bounded `abort()` escape hatch) and has a test. `decodeFrames` now validates a whole response before dispatching any of it, chunk sets survive a 429 rather than being deleted mid-retry, batches are bounded below the WebView2 body limit, and `_fail` no longer fires after a clean close.

Checks on `c05b3c997`: stream suite `-race` green, `-tags server` build and suite green, `tsc --noEmit` clean, `vitest run src/stream.test.js` 8/8.

### New: reaping an idle session permanently disables streams for a page that is still alive

`retiredThrough` is keyed on window and generation, and a page's generation is fixed for its whole lifetime. So retiring a session does not retire *that session* — it retires *that page*. Every deletion path now calls `retireLocked`, including `reap()` and `removeSession()`, which are cleanup of a session that may belong to a page still sitting in front of the user.

The sequence, all normal behaviour:

1. The page opens a stream and later closes it. `pollLoop` runs `while (G.connections.size > 0)`, so with no connections it exits and stops polling.
2. Nothing touches `lastSeen` after that. `connCount()` is 0, so `reap()` applies `streamSessionTTL`, not the 10-minute grace, and deletes the session 60–80s later — then calls `retireLocked(s)`, setting `retiredThrough[window] = s.generation`.
3. The same page, still loaded, calls `Stream("x")` again. Same session id, same generation. `session()` takes the `!ok` branch, finds `generation <= retiredThrough[windowID]`, and returns nil → 410 Gone. Its poll gets 410 too.

The page can never open another stream without a reload. An app that opens a stream on demand rather than at startup hits this on the second use.

Proved with a throwaway test that performs exactly what `reap()` does to a connection-less session — `delete(m.sessions, id)` then `m.retireLocked(s)` — and then reopens:

```
--- FAIL: TestReviewProbeReapRetiresLivePage (0.00s)
    live page cannot open a stream after its idle session was reaped
```

The same shape applies to `removeSession()` on the poll's `!alive` path: after any 410, that page is permanently out of streams.

Suggested direction: retire on *supersession*, not on cleanup. The supersede loop and `dropWindow` are the two places where the page really is gone, and the supersede loop already bumps the watermark through `s.generation - 1`, which is what blocks the late open POST from §1 of the previous comment. `reap()` and `removeSession()` deleting the map entry without touching the watermark would let a live page re-register its own generation while still refusing a genuinely superseded one. Worth confirming against `TestStreamRetiredSessionCannotBeRecreated`, which I think passes on the supersede-driven watermark alone.

### Still open from the previous comment

- Server-mode `JSONStream` parses once per *distinct* listener, so two `message` listeners means two `JSON.parse` calls of the same payload and two distinct objects. The `WailsSocket` path parses once through `_decode`. The double-add case is fixed; this one is not.
- `toBytesSync` and `toBytes` each construct `new TextEncoder()` per string send.
- `v3/tests/stream-performance/main.go:556` still reads "blocking Send … is markedly faster than spinning on TrySend (measured 46k vs 432k frames/s at 4 KB)", which puts the faster number on the losing side.

### Minor

`retiredThrough` entries are never removed, including in `dropWindow` where the window is gone for good. One `uint64` per window ever opened, so it is a note rather than a leak worth acting on — but if `dropWindow` did delete the entry, a late request from the destroyed window could recreate a session, which is presumably why it does not. Worth a comment saying so.

leaanthony:
## `performance.timeOrigin` availability, and what the real risk turned out to be

The generation anchor in `nextPageGeneration()` carries a comment about `performance.timeOrigin` being absent on older engines, so I measured it on the two non-macOS targets rather than reasoning from the comment.

**Linux** — lin-node1, Ubuntu 26.04, WebKitGTK 2.52.3, offscreen `WebKit2.WebView` under Xvfb:

```json
{"perf":"object","timeOrigin":1786421600569,"finite":true,"sessionStorage":"object"}
```

**Windows** — win-node1, WebView2 runtime 151.0.4129.72, headless Edge (same Chromium):

```json
{"perf":"object","timeOrigin":1786421636768.2,"finite":true,"ss":"object"}
```

`Number.isFinite(performance.timeOrigin)` is true on both, so the `Date.now()` branch is unreachable there. Not marginal either: Chromium has had it since Chrome 62, and the v3 Linux build requires `webkit2gtk-4.1` (the cgo `pkg-config` line in `v3/internal/operatingsystem/webkit_linux_gtk3.go`), an API line well past the point WebKit shipped it. The engine the comment actually describes is old WKWebView on macOS. The guard is correct to keep, and the fallback is unit-consistent — both sources are milliseconds since epoch, so `Math.floor(origin * 1000)` yields microseconds either way and a page using the fallback cannot produce a generation on a different scale from one that does not.

### The risk is not availability, it is that both sources are wall-clock

`timeOrigin` is wall-clock at page creation, not monotonic. A backward clock step — NTP correction, VM resume, dual-boot RTC skew, a user changing the clock — makes a later page load produce a *smaller* generation than an earlier one.

The `sessionStorage` counter is what actually guarantees monotonicity; `Math.max(current + 1, pageTime)` climbs straight over such a step. `timeOrigin` is only the anchor for the first load and for the storage-unavailable path. So the failure needs both conditions, and they share a branch: the `catch` that handles storage disabled by policy is precisely the branch with no counter to fall back on. A page reloading there after a backward clock step gets `generation <= retiredThrough` and 410 Gone with no recovery — the same terminal state as the reap finding above, reached a different way.

One precision note, already handled: WebKit clamps timer resolution to 1 ms — the Linux value above is a whole integer where Chromium's carries `.2` — so on WebKit the `* 1000` is fake precision and two reloads inside one millisecond produce an identical `pageTime`. The comment's claim that the stored counter disambiguates same-millisecond reloads is accurate, and on WebKit it is load-bearing rather than theoretical.

A protocol-level recovery would close both this and the reap case: return the window's watermark on the 410 (say `x-wails-stream-generation-min`) and let the client re-anchor above it and retry once. It adds no new exposure, since the generation is client-supplied already, and it converts two permanent-failure paths into a single retry.

---

## Outstanding, consolidated across this review

**Correctness**

1. **`reap()` and `removeSession()` retire the page, not the session.** A live page that closes its last stream stops polling, gets reaped at the 60s TTL, and can then never open another stream — 410 forever. Reproduced. Suggested fix: retire on supersession only (the supersede loop and `dropWindow`), since the supersede loop's `s.generation - 1` bump is what blocks the late open POST that the watermark exists for.
2. **Wall-clock generation with no storage fallback**, as above. Lower likelihood than 1, same terminal state. The 410-carries-watermark recovery covers both.

**Quality**

3. Server-mode `JSONStream` parses once per *distinct* listener — two `message` listeners means two `JSON.parse` calls of one payload and two distinct objects, where the `WailsSocket` path parses once via `_decode`. The same-listener double-add case is already fixed.
4. `toBytesSync` and `toBytes` each construct `new TextEncoder()` per string send; a module-level instance is free on the hot path.

**Record and docs**

5. `v3/tests/stream-performance/main.go:556` — "blocking Send … is markedly faster than spinning on TrySend (measured 46k vs 432k frames/s at 4 KB)" puts the faster number on the losing side. If transposed, it plants a wrong figure in a measurement record this branch treats as authoritative.
6. `retiredThrough` entries are never pruned, including in `dropWindow` where the window is gone for good. Presumably deliberate, so that a late request from a destroyed window cannot recreate a session — worth a comment saying so.
7. `TestGenerationCounter_StaleCallbacksDiscarded` still has a flake window between `d.add(fn1)` and `d.mu.Lock()` (pre-existing, not a regression), and its header block still describes the old sequence — "Sleep 3ms", steps 2 and 3 in the other order — against code that now sleeps 10 ms.

**Resolved in `c05b3c997`**, for the record: the open-POST resurrection path, the supersede race, the server-mode writer flush P1, `JSONStream` double-add, `_fail` after clean close, whole-response frame validation, chunk sets surviving a 429, and batch bodies bounded below the WebView2 limit.

Verification on `c05b3c997`: stream suite `-race` green, `-tags server` build and suite green, `tsc --noEmit` clean, `vitest run src/stream.test.js` 8/8.

leaanthony:
## Final review — `1e25e3c47`

The two correctness findings are fixed, and fixed in the right shape. One blocking issue is left, and it is the reason CI is red.

### Verified fixed

**Reap no longer retires the page generation.** `retireLocked` is now reachable only from the supersede loop (`stream.go:451`) and `dropWindow` (`stream.go:500`) — the two places where a page is provably gone. `reapStale` and `removeSession` are cleanup only. That is exactly the right split, and extracting `reapStale(now)` makes it testable without waiting on the ticker.

Re-checked independently rather than trusting the new tests, driving the real `reapStale` rather than simulating it: a live page survives an idle reap and can open and poll again, and a genuinely superseded page stays dead across a later reap of the current page's own session. Both pass. `TestStreamReapPreservesSupersededGenerationWatermark` is the regression guard that matters here — it is the one that would catch someone "simplifying" the split back out.

**Bonus, not something I flagged:** the `HTTPTransport.Start`/`Stop` lifecycle fix is a real bug caught in passing — a second `Start` leaked a janitor goroutine, a second `Stop` panicked closing a closed channel, and `Stop` now joins the goroutine instead of racing it.

### Blocking: the committed production runtime carries an inline sourcemap

`Bundled runtime matches source` fails, and it is not a stale-checkout artefact. Reproduced locally with `npx task runtime:build:assets`:

| | bytes |
|---|---|
| committed `runtime.js` | 433,186 |
| clean rebuild | 54,547 |

The extra 378 KB is a base64 `//# sourceMappingURL=data:application/json;base64,…` payload. Decoded, it carries `sourcesContent` for all 29 TypeScript files — 204,691 characters of original source — inside the **production** bundle that is embedded in every shipped Wails binary. `runtime.debug.js` is correct; only the minified bundle is affected.

Introduced in `c05b3c997`: `2be8b3ef7` was 53,277 bytes and every commit before it 50–53 KB. `1e25e3c47` did not touch the bundles, so it carries through.

The Taskfile is not at fault — running it produced a clean 54,547-byte bundle with no sourcemap, so this is a one-off artefact from a hand-run build using the debug flags. Rebuild and commit:

```
cd v3 && npx task runtime:build:assets
```

Worth fixing on its own merits even setting CI aside: 8× bloat on an asset that ships in every binary, and the full TypeScript source distributed with released apps.

`Cloudflare Pages: wails-v3-site` is also red, at 0s. `Check Docs` passes, so this looks like the deploy side rather than the `.mdx` change, but its log is not visible from the CLI — worth a glance.

### Still open

Unchanged by this commit, and all reasonable deferrals:

1. Generation is wall-clock on both branches; a backward clock step with `sessionStorage` unavailable still gives a permanent 410. The 410-carries-watermark recovery would close it.
2. Server-mode `JSONStream` parses once per distinct listener.
3. `new TextEncoder()` constructed per string send.
4. `v3/tests/stream-performance/main.go:556` — the `46k vs 432k` figures read backwards against the sentence's own claim.
5. `TestGenerationCounter_StaleCallbacksDiscarded` still has the fn1-side flake window, and its header block still describes the superseded sequence.

The `retiredThrough` pruning note is answered by the new paragraph in the internals doc; no code comment needed.

### Verification on `1e25e3c47`

Full `pkg/application` package with `-race` green (2.1s); `pkg/application` under `-tags server` with `-race` green; `tsc --noEmit` clean; full `vitest run` 1177/1177 across 6 files.

One correction to my earlier comments: I twice reported "`-tags server` build green" from a command whose exit status I had misread through a pipe. Accurately: `go build -tags server ./pkg/... ./internal/...` is clean apart from `internal/commands/build_assets/ios` failing with "function main is undeclared in the main package", which is a template package untouched by this branch and pre-existing on master — not a stream issue, but not something my earlier "green" claim had actually established either.

leaanthony:
Reviewed `60adebf3f..8820b3ee9` — the four stream commits plus the master merges.

**One real bug, proven below. Everything else is sound.**

## Verification

| check | result |
|---|---|
| `go test -race -count=1 ./pkg/application/` | ok, 2.359s |
| `tsc --noEmit` | exit 0 |
| `vitest run` | 6 files, 1207 tests passed |
| `task runtime:build:assets` | bundles byte-reproducible |

---

## Global lifecycle slots leak on a close/teardown race

`reserveOpen()` takes a slot from `m.lifecycles` (cap `streamMaxConnectionsGlobal = 4096`). Ownership then *transfers* to the queued close frame: `shutdown()` deliberately skips the release when `closeQueued` is set, on the basis that `releaseOutbound(frameClose)` will do it when the frame drains.

That handoff is not atomic, and `streamSession.close` calls `c.shutdown()` **directly** — outside the `closeOnce` gate that otherwise serialises `Close` against `closedByPeer`:

```go
// stream_session.go, close()
for _, c := range conns {
    c.shutdown()          // not gated by closeOnce
}
```

Interleaving:

1. **A** — `conn.Close()` wins `closeOnce`, sets `closeQueued = true`, enters `enqueue`, blocks on `s.mu`
2. **B** — `session.close()` holds `s.mu`, sets `closed = true`, releases queued frames, unlocks
3. **B** — `c.shutdown()` wins `shutdownOnce`, evaluates `c.lifecycle && !c.closeQueued.Load()` → `closeQueued` is still true → **skips `releaseLifecycle()`**
4. **A** — acquires `s.mu` inside `enqueue`, sees `s.closed`, returns `ErrStreamClosed`
5. **A** — `closeQueued.Store(false)`, but `shutdown` has already made its decision
6. **A** — `c.shutdown()` → `shutdownOnce` already consumed → no-op

No close frame was ever queued, so `releaseOutbound(frameClose)` never runs. The slot is gone for the life of the process.

### Reproduction

Racing `Close()` against `session.close()` and asserting the global allowance returns to zero:

```go
func TestProbeLifecycleLeak(t *testing.T) {
	const iterations = 3000
	leaks := 0
	for i := 0; i < iterations; i++ {
		mgr := newStreamManager(nil)
		s := newStreamSession("sess", 1, mgr)
		mgr.sessions["sess"] = s

		if !mgr.reserveOpen() {
			t.Fatal("reserveOpen failed")
		}
		ctx, cancel := context.WithCancel(context.Background())
		c := &StreamConn{id: 1, name: "t", sink: s, ctx: ctx, cancel: cancel, manager: mgr, lifecycle: true}
		c.inCond = sync.NewCond(&c.inMu)
		s.mu.Lock()
		s.conns[1] = c
		s.out = append(s.out, outFrame{connID: 1, kind: frameOpen})
		s.outControls++
		s.mu.Unlock()

		var wg sync.WaitGroup
		wg.Add(2)
		go func() { defer wg.Done(); _ = c.Close() }()
		go func() { defer wg.Done(); s.close() }()
		wg.Wait()

		if mgr.lifecycles.Load() != 0 {
			leaks++
		}
	}
	if leaks > 0 {
		t.Fatalf("lifecycle slot leaked in %d/%d iterations", leaks, iterations)
	}
}
```

```
--- FAIL: TestProbeLifecycleLeak (0.01s)
    lifecycle slot leaked in 3/3000 iterations
```

### Impact

The trigger is a page reload landing while a handler runs its `defer c.Close()` — the ordinary path, and frequent under `wails3 dev`. Leaks are permanent and cumulative. At 4096 the process refuses **every** new stream connection app-wide, with no recovery short of restart. Low rate, no self-correction, hard stop at the end.

### Suggested direction

Rather than making the handoff atomic, consider removing it: give the queued close its own reservation (`streamOutControlDepthGlobal` already exists for controls) and let `shutdown()` unconditionally own the lifecycle release. That deletes `closeQueued` and the ordering constraint with it, instead of adding a CAS that the next refactor has to preserve.

---

## What works well

- **The global accounting is the right call.** Per-session bounds genuinely did multiply by the session allowance into a multi-gigabyte ceiling; capping in the manager is the only place that can hold.
- **`decodeStreamBatch` is now overflow-safe** — `len(body)-off < 4` rather than `off+4 > len(body)`.
- **Input validation closes real gaps**: the explicit frame-kind allowlist, `streamMaxNameLen`, and rejecting a body on control frames before session lookup.
- **`ErrStreamTooLarge` on `Send`/`TrySend`** makes the 64 MB limit symmetric instead of only enforced on ingress.
- **`Receive` and `deliverWithBackpressure` no longer hold `inMu` across manager calls**, which removes a lock-ordering hazard between the connection and the manager budget.
- **Client-side `AbortController` plumbing** correctly tears down in-flight sends and the poll when the last connection goes.

## Smaller notes

**The check → reserve → recheck → rollback pattern now appears twice** (`enqueue` and `deliverWithBackpressure`). Both are correct, but the invariant they maintain — local and global capacity are guarded by independent locks, so the local check *must* be redone after the global reservation — is currently only reconstructible by reading the rollback calls. Worth stating once above `reserveOutbound`, since a future edit that drops the recheck would look harmless and would double-count silently.

**`if n < 0` in `decodeStreamBatch`** is only reachable on 32-bit builds, where `int(uint32)` can go negative. Harmless, but on every platform CI exercises it reads as dead code — a one-line comment would stop someone deleting it.

leaanthony:
Pushed `0a40242cb` fixing the lifecycle slot leak reported above.

**Approach: remove the handoff rather than make it atomic.** Closes now reserve from their own counter instead of inheriting the connection's lifecycle slot:

```go
case frameClose:
    reserved = reserveCounter(&m.outCloses, 1, streamOutCloseDepthGlobal)
```

`shutdown()` releases the lifecycle slot unconditionally — it runs exactly once and the slot belongs to the connection for exactly the connection's lifetime. `closeQueued` is deleted, and with it the window where a concurrent `shutdown` and a failing `enqueue` disagreed about who releases.

This also collapsed `rollbackOutbound` into `releaseOutbound`. The `frameClose` special case was its only reason to exist; every kind now releases exactly what it reserved, so the rollback and drain paths are the same call. Net fewer lines of logic despite the added counter.

The property the original design was protecting is kept: a queued close still cannot be starved by refused opens, because `streamOutCloseDepthGlobal` is separate from `streamOutControlDepthGlobal` and sized to the connection allowance. Each connection queues at most one close, and closes carry no payload.

**Tests.** `TestStreamCloseRollbackPreservesLifecycleReservation` asserted the transfer semantics that are now gone, so it is replaced by `TestStreamCloseAccountingIsIndependentOfLifecycle`, plus `TestStreamLifecycleSlotSurvivesCloseTeardownRace` which races `Close()` against `session.close()` and asserts both counters return to zero.

| check | result |
|---|---|
| the probe that failed 4/3000 on `0d3a4c3c9` | passes |
| `go test -race ./pkg/application/` | ok, 2.572s |
| same under `-tags server` | ok, 3.778s |
| `go vet` | clean |
| regression test at `-race -count=5` (15,000 iterations) | ok |

Server mode is unaffected: it reserves and releases its own slot via `defer` in `stream_server.go` and never sets `lifecycle`, so `shutdown` skips that block there.

Two things worth a second opinion. The regression test is timing-dependent — it caught this reliably here, but the deterministic coverage is the accounting test beside it. And this changes a global resource contract, so `streamOutCloseDepthGlobal` at 4096 is worth a sanity check against your intended sizing.

leaanthony:
Reviewed `0a40242cb`. The separate lifecycle and queued-close counters are the right fix: connection teardown now releases its lifecycle slot exactly once via `shutdownOnce`, while every close frame releases only the capacity it reserved, whether it is drained, discarded, or fails to queue. This removes the non-atomic ownership handoff that caused the teardown race.

Validated with:
- the new close/teardown regression tests, 20 repetitions under `-race`
- the full Stream tests under `-race`
- the server-tagged Stream/WebSocket tests under `-race`

All pass. I found no new correctness issue in this correction.

v3/pkg/application/stream_session.go
leaanthony: Addressed by taking the second option this comment offered — defining the ownership semantics explicitly rather than copying.

`Send` transfers ownership of the slice to the transport and does not copy it; the caller must not mutate or reuse it after a successful call. This is documented on the Go API in the Streams guide and in the internals guide, and the direction is deliberately asymmetric: JavaScript `send()` *does* snapshot mutable binary input synchronously, matching native `WebSocket`, because a socket-shaped object can be handed to code that never read our docs.

A Go-side copy was measured rather than assumed: on the JS→Go path the equivalent copy costs ~2.5% against a pooled run-to-run deviation of 7.3%, i.e. inside the noise. The cost was not the deciding factor though — the deciding factor is that a Go caller almost always hands over freshly marshalled bytes, and one that needs to reuse storage can hand over a fresh slice.

v3/pkg/application/stream_server.go
leaanthony: Fixed. The writer no longer uses the handler context.

`newWSStreamSink(conn, a.ctx, a.streams)` roots the socket I/O context in the **application** context, so `StreamConn.Close` cancelling the handler context can no longer kill writes that a preceding `Send` already accepted. The reason is recorded at the constructor so it does not get undone by a later refactor.

`serveStreamWS` then waits for the queue to drain before tearing the socket down, with a bounded escape hatch for a peer that has stopped reading:

```go
select {
case <-sink.done:
case <-time.After(5 * time.Second):
    sink.abort()
    <-sink.done
}
```

The reported case — `defer c.Close(); c.Send(final); return` — now delivers the final frame and the close.

v3/pkg/application/stream_server.go
```diff
@@ -0,0 +1,266 @@
...
+	})
+	if err != nil {
+		a.error("stream websocket accept failed: %w", err)
+		return
+	}
+
+	// coder/websocket defaults to a 32 KiB read limit, which would close the
+	// connection on any frontend frame past that — well below the 64 MB this
+	// transport documents.
+	conn.SetReadLimit(streamMaxSendBytes)
```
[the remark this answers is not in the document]
leaanthony: Implemented as recommended, including the second half about the events broadcaster.

Both endpoints now share one accept policy:

```go
func (a *App) websocketAcceptOptions() *websocket.AcceptOptions {
	server := a.options.Server
	return &websocket.AcceptOptions{
		OriginPatterns:     append([]string(nil), server.WebSocketOriginPatterns...),
		InsecureSkipVerify: server.WebSocketAllowAllOrigins,
	}
}
```

`/wails/stream/ws` and the events broadcaster both call it, so the policy cannot drift between them. Defaults are same-origin: the request host is always permitted and `WebSocketOriginPatterns` is empty unless an application adds patterns. Any-origin access is now the explicit `WebSocketAllowAllOrigins` opt-in, documented as unsafe for servers reachable by untrusted browser content.

PR #5952, opened 2026-08-12
fix(v3/windows): bound the WebView2 embed wait and re-enable the cookie test

leaanthony:
Follow-up to #5951, which skipped `TestCookieManager` in CI as a stopgap. This fixes the underlying cause and re-enables the test.

Reproduced and verified on a real Windows host, not inferred.

## The hang is a production bug, not a test bug

`Chromium.Embed` pumped messages with `GetMessageW` keyed only on `e.inited`. `CreateCoreWebView2Controller` is asynchronous and is **not guaranteed to call back at all** — in a session with no interactive window station it returns success, logs no error, and simply never completes. `GetMessageW` returns 0 only on `WM_QUIT`, so the loop had no way out.

The consequence outside CI: a user whose WebView2 runtime fails to bring up a controller gets a permanently hung window with nothing logged. In CI, Go's 10 minute test timeout was the only thing that ever broke the loop, which is why one flaky test killed the whole Windows job.

Verified on `win-node1` over SSH (session 0, no interactive window station) — this reproduces **deterministically**, unlike the 1-in-3 flake on GitHub runners:

```
=== RUN   TestCookieManager
[WebView2] Environment created successfully
panic: test timed out after 1m30s
	running tests:
		TestCookieManager (1m30s)
...
edge.(*Chromium).Embed(...) chromium.go:218
```

Note `Environment created successfully` — the environment succeeds and the *controller* never completes, with no error on any path.

## Changes

**Bound the wait.** `Embed` now waits with a deadline via `MsgWaitForMultipleObjects` + `PeekMessageW` instead of parking in `GetMessageW`. On timeout it reports through `errorCallback`, which is fatal by design — a webview that never initialises leaves nothing to host the application, so there is no recovery to return to. Adds those two procs and the four constants they need to `internal/w32`. The 30s constant is a backstop, not a budget — controller creation is normally well under a second.

**Pin the test to its apartment.** WebView2 has thread affinity: every call on the control must happen on the thread that created it. The test goroutine could migrate between OS threads, so it is now pinned before `CoInitializeEx` (kept from #5951, thanks @copilot).

**Flatten the subtests.** `t.Run` invokes each subtest on a *new* goroutine, and `runtime.LockOSThread` does not extend to it — so the cookie-manager calls and `Release` inside the two subtests were running outside the apartment that created the control. They are now plain blocks in the parent test. Thanks @coderabbitai, this was the substantive catch.

**Re-enable the test**, dropping the `WAILS_TEST_WEBVIEW2` / `CI` skip from #5951. This also moots CodeRabbit's note that the guard accepted any non-empty value rather than exactly `1`.

## Verification

**1. The timeout path, against a real WebView2** — same host and session that used to hang for 10 minutes:

```
=== RUN   TestCookieManager
[WebView2] Environment created successfully
[WebView2 Error] timed out after 30s waiting for the WebView2 controller; the WebView2 runtime did not complete initialisation
FAIL	github.com/wailsapp/wails/v3/internal/webview2/pkg/edge	30.047s
ELAPSED: 31s
```

10 minute hang that killed the job → 30 second stop naming the cause.

**2. The success path.** Replacing `GetMessageW` changes the path every Windows app takes, so it needed proving rather than assuming. A standalone harness on the same host, using the identical loop, posts a message and confirms it is dispatched to the `WndProc`, then confirms the deadline fires when nothing is posted:

```
dispatched=true elapsed=0s
timeout_path_returned=false elapsed=2.008s
RESULT: PASS
```

**3. Build.** `GOOS=windows go build ./...`, `go vet ./internal/webview2/...` and `go test -c` all clean. (`internal/commands/build_assets/ios` fails to build identically on clean master — pre-existing and unrelated.)

## What is not verified here

I could not run the full green path of `TestCookieManager` end to end: it needs an interactive window station, and the only interactive session on that host belongs to another user. CI is the check for that — the test passed there in ~14s before, and this change does not alter what happens once the controller completes.

## Summary by CodeRabbit

* **Bug Fixes**
  * WebView2 initialization now has a 30-second timeout instead of waiting indefinitely.
  * Initialization failures are reported clearly, improving application responsiveness.
  * WebView2 cookie operations now reliably run on the required Windows thread.
  * Message processing during startup is more responsive and stops cleanly when the application exits.

* **Tests**
  * Improved WebView2 embedding and cookie tests to better reflect Windows thread-affinity requirements.

### DIFF
--- a/v3/internal/webview2/internal/w32/w32.go
+++ b/v3/internal/webview2/internal/w32/w32.go
@@ -41,6 +41,13 @@ var (
 	User32SetWindowLongPtrW  = user32.NewProc("SetWindowLongPtrW")
 	User32AdjustWindowRect   = user32.NewProc("AdjustWindowRect")
 	User32SetWindowPos       = user32.NewProc("SetWindowPos")
+
+	// GetMessageW blocks until a message arrives, which makes any loop built on
+	// it unbounded. These two allow a message loop to wait with a deadline
+	// instead: MsgWaitForMultipleObjects returns on either input or timeout,
+	// and PeekMessageW then drains whatever arrived.
+	User32MsgWaitForMultipleObjects = user32.NewProc("MsgWaitForMultipleObjects")
+	User32PeekMessageW              = user32.NewProc("PeekMessageW")
 )
 
 const (
@@ -57,6 +64,17 @@ const (
 
 const (
 	SWShow = 5
+)
+
+const (
+	WM_QUIT = 0x0012
+
+	PM_REMOVE = 0x0001
+
+	// QS_ALLINPUT wakes MsgWaitForMultipleObjects for any queued message.
+	QS_ALLINPUT = 0x04FF
+
+	WAIT_TIMEOUT = 0x00000102
 )
 
 const (
--- a/v3/internal/webview2/pkg/edge/chromium.go
+++ b/v3/internal/webview2/pkg/edge/chromium.go
@@ -210,25 +210,74 @@ func (e *Chromium) Embed(hwnd uintptr) bool {
 		e.errorCallback(fmt.Errorf("error getting Webview2 runtime version: %s", err.Error()))
 	}
 
-	var msg w32.Msg
-	for {
-		if atomic.LoadUintptr(&e.inited) != 0 {
-			break
-		}
-		r, _, _ := w32.User32GetMessageW.Call(
-			uintptr(unsafe.Pointer(&msg)),
-			0,
-			0,
-			0,
-		)
-		if r == 0 {
-			break
-		}
-		w32.User32TranslateMessage.Call(uintptr(unsafe.Pointer(&msg)))
-		w32.User32DispatchMessageW.Call(uintptr(unsafe.Pointer(&msg)))
+	// Pump until the controller reports ready, but never without a deadline.
+	//
+	// CreateCoreWebView2Controller is asynchronous and is not guaranteed to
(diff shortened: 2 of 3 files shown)


leaanthony:
Correction to the description above, found after opening this.

The `return false` on the timeout path is **unreachable**. `Chromium.errorCallback` ends with `os.Exit(1)`, so it never returns to its caller:

```go
func (e *Chromium) errorCallback(err error) {
	e.globalErrorCallback(err)
	os.Exit(1)
}
```

The measured improvement stands — the 10 minute hang becomes a 30 second stop with the cause logged, verified on hardware — but the mechanism is process termination, not an error return. `Embed`'s `bool` was already decorative before this PR: all four of its internal `errorCallback` sites are fatal too.

Filed as #5953, since making `Embed` recoverable changes behaviour for every Windows app and should not ride along in a test-flakiness fix. Happy to either drop the dead `return false` here, or switch the timeout path to `globalErrorCallback` so the return is real — say which you prefer.

leaanthony:
Resolved: a failed WebView2 is fatal by design, so `errorCallback` exiting is correct and there is nothing to return to.

Pushed `b761d716e` dropping the unreachable `return false` and rewriting the comment to state the fatality explicitly rather than implying a recoverable path. The description above is corrected to match. Behaviour is unchanged from what was measured — bounded wait, then a prompt exit with the cause logged, instead of a ten minute hang.

Closing #5953; its premise was that `Embed` should be recoverable, which is not the design.

PR #5973, opened 2026-08-13
fix(v3/linux): claim the single instance name under the app's own id

overlordtm:
# Description

`SingleInstance` cannot be used in a sandboxed application without handing it the
entire session bus.

`linuxLock.acquire` derived its name from `UniqueID` by mangling it into a shape
unrelated to the application id:

    com.example.app  ->  org.wails_app_com_example_app.SingleInstance

A flatpak may only own names prefixed with its app id, so the request is refused
unless the manifest grants `--socket=session-bus`, which is unfiltered access to
the whole bus. Granting that to obtain one well-known name is a poor tradeoff, and
it is what every sandboxed Wails app using this feature has to do today.

The mangling was not arbitrary. One string served as bus name, interface name and
object path, and D-Bus spells those differently: bus names may contain hyphens,
interface names may not, and object paths separate elements with `/` and allow
neither. Collapsing everything to underscores under an `org.` prefix produced one
string valid in all three positions.

This derives the three separately, so the bus name can keep `UniqueID` verbatim —
which is what puts it under the app id, and what the documented convention for
that field ("unique per application, e.g. com.myapp.myapplication") already
suggests:

    UniqueID                 bus name                                interface                               object path
    com.myapp.myapplication  com.myapp.myapplication.SingleInstance  com.myapp.myapplication.SingleInstance  /com/myapp/myapplication/SingleInstance
    net.my-company.my-app    net.my-company.my-app.SingleInstance    net.my_company.my_app.SingleInstance    /net/my_company/my_app/SingleInstance

An id that cannot yield a valid name, like empty element, leading digit, character
D-Bus disallows - is now rejected with a message naming the offending part,
instead of silently producing a name that can never be claimed.

*Breaking change*: the well-known name changes, so an old instance and a new
instance will not see each other. This only matters across an upgrade, while both
are running.

## Type of change
  
# How Has This Been Tested?
  
- [ ] Windows
- [ ] macOS
- [x] Linux (Ubuntu 26.04)
      
## Test Configuration

```
Wails v3.0.0-beta.8 › Wails Doctor
                                                                                                                                                                                                                                      
# System
        
┌────────────────────────────────────────────────────────────────────────────────────────────────┐
| Name                | Ubuntu                                                                   |
| Version             | 26.04                                                                    |
| ID                  | ubuntu                                                                   |
| Branding            | 26.04 LTS (Resolute Raccoon)                                             |
| Platform            | linux                                                                    |
| Architecture        | amd64                                                                    |
| Desktop Environment | ubuntu:GNOME                                                             |
| NVIDIA Driver       | 595.84 (BD898001360CB9BA4655D6C)                                         |
| XDG_SESSION_TYPE    | wayland                                                                  |
| CPU                 | AMD Ryzen 9 9900X 12-Core Processor                                      |
| GPU                 | Granite Ridge [Radeon Graphics] (Advanced Micro Devices, Inc. [AMD/ATI]) |
| Memory              | 61GB                                                                     |
└────────────────────────────────────────────────────────────────────────────────────────────────┘
                   
# Build Environment
                   
┌──────────────────────────────┐
| Wails CLI    | v3.0.0-beta.8 |
| Go Version   | go1.25.12     |
| -buildmode   | exe           |
| -compiler    | gc            |
| CGO_CFLAGS   |               |
| CGO_CPPFLAGS |               |
| CGO_CXXFLAGS |               |
| CGO_ENABLED  | 1             |
| CGO_LDFLAGS  |               |
| GOAMD64      | v1            |
| GOARCH       | amd64         |
| GOOS         | linux         |
└──────────────────────────────┘
```

# Checklist:

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved single-instance handling on Linux to prevent conflicts between separate application instances.
  - Improved detection of duplicate launches and unexpected startup failures.
  - Added clearer error context when the application cannot claim its single-instance lock.
  - Improved notification delivery between existing and newly launched instances.
- **Tests**
  - Added coverage for valid and invalid application identifiers and single-instance naming scenarios.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->


v3/pkg/application/single_instance_linux.go
```diff
@@ -22,11 +23,48 @@ func (f dbusHandler) SendMessage(message string) *dbus.Error {
...
+// UniqueID. They cannot all be the same string, because D-Bus spells them
+// differently: bus names may contain hyphens, interface names may not, and
+// object paths separate elements with "/" and allow neither hyphens nor dots.
+//
+// The bus name keeps UniqueID verbatim. That matters for sandboxed builds: a
+// flatpak may only own names prefixed with its own app id, so an app that
+// follows the documented convention for UniqueID ("unique per application, e.g.
+// com.myapp.myapplication") claims a name it is allowed to own and needs no
+// extra portal permission.
+func singleInstanceNames(uniqueID string) (busName, interfaceName, objectPath string, err error) {
```
atterpac, line 44: UniqueID could differ from the actual FLATPAK scope and puts the onus on the developer to ensure their ApplicationID == SingleInstance.UniqueID

For example if we set an app-id in flatpak of 
`app-id: com.example.MyApp`

but set up a config of 
```
      Name: "My App",

      SingleInstance: &application.SingleInstanceOptions{
          UniqueID: "com.example.my-app-lock",
      },
  })
```

we would create dbus id of `com.example.my-app-lock.SingleInstance`
but flatpak would only permit ids under `com.example.MyApp.*`

Not sure the best solution potentially using the embedded `FLATPAK_ID` environment so we know we are using the configured `app-id` and append a hash of the unique id config option? 

https://docs.flatpak.org/en/latest/flatpak-command-reference.html
> Flatpak sets the environment variable FLATPAK_ID to the application ID of the running app.

atterpac (changes requested):

v3/pkg/application/single_instance_linux.go
atterpac: Granted the env path would likely fail outside of flatpak environment

overlordtm: I can add additional property `ForceFlatpakUniqueId` to `SingleInstanceOptions`, defaulting to `false`. In that case, `FLATPAK_ID` would override developer set `UniqueID`, ensuring dbus id has correct format. If `FLATPAK_ID` env variable is not present, behavior stays same as it is in current master.

In case developer has good reason to force own `UniqueID`, he can set `ForceFlatpakUniqueId=true`.

I will also revert current changes (well, will have make special codepath for flatpak, so we do not prepend `org.wails` when generating dbus id), so behaviour change is minimal, making this more a bugfix rather than a breaking change. Agree?

atterpac (approved):

PR #5998, opened 2026-08-18
fix(v3): stop sqlite Prepare spinning forever under concurrent calls

archy-rock3t-cloud:
# Description

The id allocation loop in `PrepareContext` never reloads `nextId`, so when two concurrent calls race, the loser spins forever at 100% CPU.

## Type of change

# How Has This Been Tested?

New unit test that fails before the fix and passes after (incl. `-race`); package suite green.

- [ ] Windows
- [x] macOS
- [ ] Linux

## Test Configuration

```
macOS 26.5.2 (darwin/arm64), go1.26.4, Wails CLI v3.0.0-dev (source checkout)
```

# Checklist:

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability when preparing SQLite statements concurrently.
  - Prevented stale statement identifiers from causing preparation failures or hangs under contention.
  - Added safeguards for exhausted statement identifiers.

- **Tests**
  - Added concurrent preparation coverage to verify completion and prevent duplicate statement identifiers.


leaanthony (approved):

archy-rock3t-cloud:
@leaanthony I tightened the test: the old one only reproduced the spin about one run in five, the new one fails 8 out of 8 against the unfixed code (32 workers released from a barrier, 100 prepares each). Rebased onto master.

PR #6024, opened 2026-08-24
fix(v2): detect versioned webkit2gtk package on Arch

### LINKED ISSUE #5775: wails doctor reports libwebkit missing on Arch (webkit2gtk split into versioned packages)
### Description

`wails doctor` reports `libwebkit` as **Not Found** on Arch Linux (and derivatives) even when WebKit is correctly installed and Wails builds fine with `-tags webkit2_41`.

Root cause: recent Arch split `webkit2gtk` into versioned packages (`webkit2gtk-4.1`, `webkit2gtk-4.0`) and **removed the unversioned `webkit2gtk`**. The pacman package manager in `v2/internal/system/packagemanager/pacman.go` only knows the old `webkit2gtk` name:

```go
"libwebkit": []*Package{
    {Name: "webkit2gtk", SystemPackage: true, Library: true},
},
```

`Dependencies()` calls `PackageAvailable` → `pacman -Si webkit2gtk`, which now fails, so the dep is marked missing.

Same class of bug as #4457 (Fedora), but distro-specific to the pacman package name.

### To Reproduce

1. Arch Linux with `webkit2gtk-4.1` installed (and no unversioned `webkit2gtk`):
   ```
   $ pacman -Q webkit2gtk-4.1
   webkit2gtk-4.1 2.52.3-1
   $ pacman -Si webkit2gtk
   error: package 'webkit2gtk' was not found
   ```
2. Run `wails doctor`.

### Expected behaviour

`libwebkit` detected as installed; diagnosis "Your system is ready for Wails development!".

### Actual behaviour

```
| libwebkit | Unknown | Not Found | |
...
Fatal: Required dependencies missing: libwebkit
```

### Suggested fix

The `[]*Package` slice is already treated as ordered alternatives (`Dependencies()` stops at the first available candidate). List the versioned names first, keep the old name as fallback:

```go
"libwebkit": []*Package{
    {Name: "webkit2gtk-4.1", SystemPackage: true, Library: true},
    {Name: "webkit2gtk-4.0", SystemPackage: true, Library: true},
    {Name: "webkit2gtk", SystemPackage: true, Library: true},
},
```

Verified locally: patched `wails doctor` then shows `libwebkit | webkit2gtk-4.1 | Installed | 2.52.3-1`. Happy to send a PR.

### System Details

```
Wails    | v2.13.0
OS       | Arch Linux (rolling), kernel 7.0.10-arch1-1
Go       | go1.26.5
libgtk-3 | gtk3           | Installed | 1:3.24.52-1
libwebkit| webkit2gtk-4.1 | Installed | 2.52.3-1  (via pkg-config; doctor misreports)
```

4RH1T3CT0R7:
`wails doctor` reports `Fatal: Required dependencies missing: libwebkit` on Arch machines where WebKit is installed and builds work fine. Arch dropped the unversioned `webkit2gtk` package and only ships `webkit2gtk-4.1`, but the v2 pacman map still lists only the old name. v3's map already lists 4.1, so this just adds the second candidate; `Dependencies()` already treats the slice as ordered alternatives, and the unversioned name stays first for machines still on 4.0. Also fixed a doc note claiming Arch ships both.

Caveat: on a 4.1-only box doctor will now say ready, but a default build there still needs `-tags webkit2_41`. Happy to print that hint next to the resolved package if you'd rather it said so outright.

I don't have an Arch box, so this is untested, would be good if someone could confirm `libwebkit` resolves to `webkit2gtk-4.1`.

Fixes #5775

## Summary by CodeRabbit

- **New Features**
  - `wails doctor` now identifies when an installed WebKit package requires the `webkit2_41` build tag and explains how to apply it when building or developing.

- **Bug Fixes**
  - Improved Linux WebKit package detection by preferring installed packages and correctly handling available alternatives.
  - Added build-tag support for newer WebKit packages.

- **Documentation**
  - Updated Arch Linux support information and the unreleased changelog with the improved WebKit guidance.


4RH1T3CT0R7:
Good catch, dropped the Arch override from the ABI 4.0 example too.

leaanthony:
I think it'd be worth showing that tag requirement... I'm wondering how we could keep it consistent though. `wails3 setup` could auto set that in the task files, but that seems like a bit more of a stretch...

4RH1T3CT0R7:
Added it. Package entries can now carry the build tags they need (a BuildTags field), and doctor prints a note for whatever WebKit package it resolved to when that's set, so it works the same on every distro that lists a 4.1 package, currently Arch and Void. I left the task-file / wails3 setup side alone for now; happy to look at that separately if you want it.

4RH1T3CT0R7:
Good catch. Dependencies now keeps scanning for an installed alternative and only falls back to the first available one when nothing is installed, so a Manjaro box with just webkit2gtk-4.1 is reported as installed with the webkit2_41 note. Added a small table test with a fake package manager covering the both-available case.

4RH1T3CT0R7:
@leaanthony both points are in now, the doctor note about -tags webkit2_41 and the installed-alternative fallback CodeRabbit asked for. Ready for another look whenever you have a moment.

PR #6025, opened 2026-08-24
chore(v3): remove precompiled example binaries

### LINKED ISSUE #5927: [v3] Remove precompiled menu_demo binary from menu example
### Have you read the Documentation Contribution Guidelines?

- [x] I have read the [Documentation Contribution Guidelines](https://v3.wails.io/contributing/).

### Description

The v3/examples/menu directory contains a ~17 MB precompiled macOS menu_demo binary.

Since the example already includes main.go and can be built from source, the binary appears unnecessary and adds significant size to the Go module.

It would be good to remove the binary from the repository.

### Self-service

- [ ] I'd be willing to address this documentation request myself.

4RH1T3CT0R7:
`v3/examples/menu/menu_demo` (17 MB) and `v3/examples/server/server` (14.8 MB) are committed macOS arm64 binaries. Nothing references either path and both examples build from source. #5927 only names `menu_demo`, but the server binary is the same thing in the same tree, so I removed both. Say the word and I'll keep this to the ticket.

`v3/.gitignore` already ignores the per-example binary for six other examples; these two slipped through on naming. A bare `go build .` in `examples/server` produces `server`, which had no entry, and the menu binary was committed as `menu_demo` rather than the `menu` that's already ignored. Both names are listed now.

`go build ./examples/menu` works with the binary deleted. `examples/server` doesn't build, but it fails the same way on unmodified master (`go mod tidy` needed), so that's pre-existing and unrelated.

Fixes #5927

## Summary by CodeRabbit

* **Chores**
  * Updated ignored example build outputs to keep generated files out of version control.

* **Documentation**
  * Added an unreleased changelog entry for removing precompiled macOS binaries from the examples.
  * Removed an outdated changelog entry about the obsolete implementation tracker.


leaanthony (approved):
Nice one! Can happen 😅

PR #6035, opened 2026-08-26
fix(v3): offer the Fn/Globe key to the input method on macOS

### LINKED ISSUE #6033: macOS: Fn/Globe key (keyCode 63) swallowed by WKWebView before reaching IME — IME voice shortcuts fail in Wails apps
## Summary

On macOS, the **Fn / Globe key (`keyCode 63`) is swallowed by WKWebView before it ever reaches the input method**, so any IME feature triggered by holding Fn (e.g. voice input / push-to-talk shortcuts in Chinese input methods like 豆包输入法/DoubaoIme) silently fails **inside every Wails app window**, while the same IME works fine in native apps and Chromium apps.

## Environment

- macOS 26.2, Apple Silicon (arm64)
- Wails app: Reasonix Desktop (Wails/WKWebView variant, v1.31.4, `com.wails.reasonix-desktop`) — but this affects **any** Wails app on macOS
- IME: DoubaoIme (豆包输入法), voice shortcut bound to "hold Fn/Globe"

## Observed behavior

| Target | Pressing Fn (physical or remapped) while IME voice input is active |
|---|---|
| Native app (e.g. TextEdit, SwiftUI app) | ✅ IME receives the key, voice input starts |
| Chromium app (Chrome, Codex desktop) | ✅ works |
| **Wails app (WKWebView)** | ❌ IME never receives it — no reaction at all |

Typing in the Wails app works fine (`keyDown` path → `NSTextInputContext` → IME). The problem is specific to the **bare Fn modifier event**: it arrives as `flagsChanged` and is dropped.

## Root cause (upstream WebKit)

WebKit's `WebViewImpl::flagsChanged()` explicitly filters `keyCode 63` (Fn/Globe) and returns **before** calling `interpretKeyEvent()` (the call that hands the event to the current `NSTextInputContext` / IME):

- https://github.com/WebKit/WebKit/blob/main/Source/WebKit/UIProcess/mac/WebViewImpl.mm#L5916-L5920
- https://github.com/WebKit/WebKit/blob/main/Source/WebKit/UIProcess/mac/WebViewImpl.mm#L6826-L6840

Chinese IMEs commonly use "hold Fn/Globe to speak" as a voice shortcut, and hardware remappers (e.g. turning a Bluetooth remote's mic button into Fn) produce exactly this event, so both physical and remapped Fn fail.

## Request

Please consider a macOS-side fix in Wails so Fn/Globe events are forwarded to the input context instead of being silently dropped, e.g. install a `NSEventMaskFlagsChanged` local monitor on the Wails window that, for `keyCode 63`, forwards the event to the current first responder's `[inputContext handleEvent:event]` while still passing the original event through to AppKit (per Apple's `NSTextInputContext.handleEvent` docs: https://developer.apple.com/documentation/appkit/nstextinputcontext/handleevent(_:)).

A WebKit-upstream fix would be "deliver Fn to the IME while still not generating a DOM keyboard event".

## Minimal repro

1. Install any Chinese IME that supports a "hold Fn to speak" voice shortcut (e.g. DoubaoIme) and bind its voice key to Fn/Globe.
2. Open any Wails app, focus a text field, and make sure the IME is active.
3. Hold Fn and speak → nothing happens.
4. Repeat in TextEdit or Chrome → works.

## Related

- Wails v3 Chinese-input issues (closed, different symptom): #4550, #2933

taliesin-ai:
Fixes #6033 — **pending confirmation from the reporter**, see Testing.

## The problem

The Fn/Globe key arrives as a *modifier change*, not a key press. WebKit drops `keyCode 63` in `WebViewImpl::flagsChanged()` before it reaches `interpretKeyEvent()`, so the current input method never sees it. An input method that binds a shortcut to holding Fn — push-to-talk voice input is the common one, and hardware remappers produce the same event — does nothing at all inside a Wails window, while working in native and Chromium apps.

Nothing in Wails handled `flagsChanged` at all, so there was no existing code path to correct; this adds one.

## The change

A local event monitor, alongside the two `application_darwin.go` already installs for mouse events, offers a keyCode 63 modifier change to the first responder's input context:

```objc
[NSEvent addLocalMonitorForEventsMatchingMask:NSEventMaskFlagsChanged handler:^NSEvent *(NSEvent *event) {
    if ([event keyCode] != kWailsFunctionKeyCode) { return event; }
    ...
    [inputContext handleEvent:event];
    return event;   // always
}];
```

The event is **returned unchanged either way**, so AppKit's own handling is untouched and no DOM keyboard event is synthesised — the page still doesn't see Fn, which is both the existing behaviour and the right one. The monitor ignores every other key code, and events belonging to windows that aren't ours.

This is the approach @1939869736luosi proposed in the issue, citing [`NSTextInputContext.handleEvent`](https://developer.apple.com/documentation/appkit/nstextinputcontext/handleevent(_:)).

`inputContext` is declared on `NSView`, not `NSResponder`, and the first responder isn't necessarily a view — hence the `isKindOfClass:` check, which also keeps the cgo build free of warnings.

## Testing

**Verified:** build and `go vet` clean, `pkg/application` tests pass, and sending real Fn presses to a running application changes nothing else — key bindings fire before and after, and nothing crashes. So the regression risk (swallowing events, or breaking normal keyboard handling) is covered.

**Not verified:** the symptom itself. Reproducing it needs an input method with a hold-Fn voice shortcut bound, which I don't have. Synthesising the event proves *delivery*, not that the input method acts on it.

@1939869736luosi — would you be able to build this branch and confirm your IME's Fn voice shortcut now works inside a Wails window? I'd rather not have this merged on the strength of reasoning alone.

## Summary by CodeRabbit

- **New Features**
  - Improved macOS keyboard handling for the Fn/Globe key in desktop windows.
  - Preserves text input behavior while preventing unintended web-based keyboard events.

### DIFF
--- a/v3/pkg/application/application_darwin.go
+++ b/v3/pkg/application/application_darwin.go
@@ -19,10 +19,49 @@ extern void registerListener(unsigned int event);
 
 static AppDelegate *appDelegate = nil;
 
+// kVK_Function from Carbon's Events.h, spelled out so that header is not
+// needed for one number.
+static const unsigned short kWailsFunctionKeyCode = 63;
+
+
 static void init(void) {
     [NSApplication sharedApplication];
     appDelegate = [[AppDelegate alloc] init];
     [NSApp setDelegate:appDelegate];
+
+	// The Fn/Globe key arrives as a modifier change rather than a key press,
+	// and WebKit drops keyCode 63 in WebViewImpl::flagsChanged() before it
+	// reaches interpretKeyEvent(), so the current input method never sees it.
+	// Input methods that bind a shortcut to holding Fn - push-to-talk voice
+	// input is the common one - therefore do nothing at all inside a Wails
+	// window, while working in native and Chromium applications.
+	//
+	// Offer the event to the first responder's input context here instead. The
+	// event is returned unchanged either way, so AppKit's own handling is
+	// untouched and no DOM keyboard event is synthesised: the page still does
+	// not see Fn, which is the existing behaviour and the correct one.
+	[NSEvent addLocalMonitorForEventsMatchingMask:NSEventMaskFlagsChanged handler:^NSEvent * _Nullable(NSEvent * _Nonnull event) {
+		if ([event keyCode] != kWailsFunctionKeyCode) {
+			return event;
+		}
+		NSWindow* eventWindow = [event window];
+		if (eventWindow == nil) {
+			return event;
+		}
+		if (![[eventWindow delegate] isKindOfClass:[WebviewWindowDelegate class]]) {
+			return event;
+		}
+		// inputContext is declared on NSView, not on NSResponder, and the
+		// first responder is not necessarily a view.
+		NSResponder* responder = [eventWindow firstResponder];
+		if ([responder isKindOfClass:[NSView class]]) {
+			NSTextInputContext* inputContext = [(NSView*)responder inputContext];
+			if (inputContext != nil) {
+				[inputContext handleEvent:event];
+			}
+		}
+		return event;
+	}];
 
 	[NSEvent addLocalMonitorForEventsMatchingMask:NSEventMaskLeftMouseDown handler:^NSEvent * _Nullable(NSEvent * _Nonnull event) {
 		NSWindow* eventWindow = [event window];


1939869736luosi:
@taliesin-ai Sorry for the late reply — I hadn’t checked my GitHub notifications for a few days. I’ve now tested the Fn/Globe behavior with the proposed fix on my actual macOS setup, and the IME voice shortcut works normally inside the Wails window; thank you for the fix.

Issue #5552, opened 2026-06-08
[v3][Linux][GTK3] getUserMedia is denied by default because WebKitGTK permission-request is not handled


jalfdev:
### Description

On Linux with Wails v3 GTK3, frontend calls to `navigator.mediaDevices.getUserMedia()` for camera or microphone are denied by WebKitGTK by default.

The frontend receives:

```text
NotAllowedError: The request is not allowed by the user agent or the platform in the current context, possibly because the user denied permission.
```

This happens even when the Linux user has OS-level access to `/dev/video*` and audio devices.

From debugging, WebKitGTK emits a `permission-request` signal for `WebKitUserMediaPermissionRequest`, but the current Wails v3 GTK3 wrapper does not appear to handle or expose this signal. Because nobody handles the request, WebKitGTK rejects it by default.

### Environment

- Wails: v3.0.0-alpha.97
- Platform: Linux / UOS Desktop 25 Professional
- Backend: Go 1.25.5
- GTK3: 3.24.41
- WebKitGTK 4.1: 2.50.4
- libsoup 3.0: 3.6.5

### Reproduction

1. Build a Wails v3 GTK3 app on Linux.
2. In the frontend, call:

```js
await navigator.mediaDevices.getUserMedia({ video: true })
```

or:

```js
await navigator.mediaDevices.getUserMedia({ audio: true })
```

3. The call fails with `NotAllowedError`.

### Current workaround

We patched `pkg/application/linux_cgo_gtk3.go` locally to handle WebKitGTK `permission-request` and allow only `WebKitUserMediaPermissionRequest`:

```c
static gboolean allow_media_permission_request(WebKitWebView *web_view, WebKitPermissionRequest *request, gpointer user_data) {
    (void)web_view;
    (void)user_data;
    if (WEBKIT_IS_USER_MEDIA_PERMISSION_REQUEST(request)) {
        webkit_permission_request_allow(request);
        return TRUE;
    }
    return FALSE;
}

static void connect_media_permission_request(GtkWidget *webview) {
    g_signal_connect(webview, "permission-request", G_CALLBACK(allow_media_permission_request), NULL);
}
```

Then after creating the WebKit web view:

```go
webView := C.webkit_web_view_new_with_user_content_manager(manager)
C.connect_media_permission_request(webView)
```

After this patch, camera and microphone access work.

### Expected behavior

Wails v3 should provide a supported way to handle WebKitGTK permission requests on Linux, especially for camera and microphone access.

[...]

Issue #4270, opened 2025-05-09
[v3]在开发模式下无法获得录音等权限


exment:
### Description

运行wails3 dev命令的时候不能弹出获取麦克风权限

### To Reproduce

1. 简单开发一个页面，获取浏览器录音权限，在浏览器上能正常运行
2. wails3 dev 运行的时候 控制台显示 The operation was aborted
3. 当直接运行 bin 目录下的二进制文件时 又能弹出申请录音权限

### Expected behaviour

当运行 wails3 dev的时候 能正常弹出 请求录音等权限

### System Details

```shell
# System 

┌────────────────────────────────────────────────────────────────────────────┐
| Name          | MacOS                                                      |
| Version       | 13.6.4                                                     |
| ID            | 22G513                                                     |
| Branding      | Ventura                                                    |
| Platform      | darwin                                                     |
| Architecture  | amd64                                                      |
| Apple Silicon | unknown                                                    |
| CPU           | Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz                  |
| CPU           | Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz                  |
| GPU           |  cores, Metal Support: Metal 3                             |
|       Metal Support: Metal 3                                               |
| Memory        | 64 GB                                                      |
└────────────────────────────────────────────────────────────────────────────┘

# Build Environment 

┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
| Wails CLI      | v3.0.0-alpha.9                                                                                                                                             |
| Go Version     | go1.23.8                                                                                                                                                   |
| -buildmode     | exe                                                                                                                                                        |
| -compiler      | gc                                                                                                                                                         |
| CGO_CFLAGS     |                                                                                                                                                            |
| CGO_CPPFLAGS   |                                                                                                                                                            |
| CGO_CXXFLAGS   |                                                                                                                                                            |
| CGO_ENABLED    | 1                                                                                                                                                          |
| CGO_LDFLAGS    |                                                                                                                                                            |
| DefaultGODEBUG | asynctimerchan=1,gotypesalias=0,httpservecontentkeepheaders=1,tls3des=1,tlskyber=0,winreadlinkvolume=0,winsymlink=0,x509keypairleaf=0,x509negativeserial=1 |
| GOAMD64        | v1                                                                                                                                                         |
| GOARCH         | amd64                                                                                                                                                      |
[...]

Issue #3735, opened 2024-09-08
[Documentation] Camera and Mic access - Darwin


leaanthony:
My Wails apps need microphone access and days ago it work perfectly normal without any config, today it can't get access to `navigator.mediaDevices` saying it's undefined.

I'm on MacOS, seems like adding this to `Info.plist` and `Info.dev.plist` between `<dict> </dict>` solves it

```
<key>NSMicrophoneUsageDescription</key>
<string>This app requires access to the microphone</string>
```

If you need camera access in your wails app and it keep saying `navigator.mediaDevices` is undefined, try using this:
```
<key>NSCameraUsageDescription</key>
<string>This app requires access to the camera</string>
```

_Originally posted by @madeindra in https://github.com/wailsapp/wails/discussions/883#discussioncomment-10575706_

qdxiao:
Thank you very much, just encountered this problem.

henmalib:
It will also just say "NotAllowedError: The request is not allowed by the user agent or the platform in the current context, possibly because the user denied permission." even with those permission added. Worked fine in production build, but failed in the dev one

turns out, you also have to give permission to the mic to the "runner", in my case it was ghostty

leaanthony:
Thanks for documenting this — the solution is correct. macOS requires usage description keys in `Info.plist` before granting access to the microphone or camera.

Add these keys to your `build/darwin/Info.plist` (they'll be preserved on regeneration):

```xml
<key>NSMicrophoneUsageDescription</key>
<string>This app requires microphone access.</string>
<key>NSCameraUsageDescription</key>
<string>This app requires camera access.</string>
```

Also add them to `build/darwin/Info.dev.plist` if you need access during `wails3 dev`. We'll add a note about this to the documentation. Closing — thanks for the clear writeup!

Issue #5883, opened 2026-08-03
[v3] docs(tutorial-04): `wails init` flags (`--name`, `--template`) do not match current CLI


ZiplEix:
### Have you read the Documentation Contribution Guidelines?

- [x] I have read the [Documentation Contribution Guidelines](https://v3.wails.io/contributing/).

### Description

While following [Tutorial 4 (Self-update a Wails app)](https://v3.wails.io/tutorials/04-self-update-a-wails-app/), the suggested `wails init` command fails to execute because the `--name` and `--template` flags are not recognized by the current Wails v3 CLI.

### Steps to Reproduce

1. Run the command specified in Tutorial 4:
   ```bash
   wails init --name updater --template vanilla
   ```

2. Observe the CLI error regarding unrecognized flags (`--name`, `--template`).

### Expected Behavior

Either:
  - The CLI should support the --name and --template flags as documented.
  - OR the documentation in Tutorial 4 should be updated to reflect the current CLI argument syntax.

### Additional Context

I am fairly new to the Wails v3 project, so I'm not entirely sure which direction is preferred by the maintainers. However, I'd be happy to submit a PR to fix the documentation once confirmed!

### Self-service

- [x] I'd be willing to address this documentation request myself.

PR #6050, opened 2026-08-28
fix(darwin): honour Permissions for media capture requests

### LINKED ISSUE #6067: [v3][macOS] The cross-platform Permissions option is ignored on macOS
### Wails version family

v3

### Exact Wails version

v3.0.0-beta.13 (also on `master` @ 9f2fbd3)

### Operating System

macOS

### Description

`WebviewWindowOptions.Permissions` is honoured on Linux (#5552) and on Windows through WebView2 (#5567), but has no effect on macOS. `WebviewWindowDelegate` is the webview's UI delegate and declares `WKUIDelegate` conformance, but never implements `webView:requestMediaCapturePermissionForOrigin:initiatedByFrame:type:decisionHandler:` — so nothing on darwin reads the map.

Capture itself still works. `WKUIDelegate.h` says an unimplemented method behaves as `WKPermissionDecisionPrompt`, so macOS acts as though every window were permanently `PermissionDefault`. What's lost is the other two policies: you can't drop WebKit's prompt for a window that only loads your own content, and you can't deny capture to a window that shouldn't have it.

The docs already describe this as a known limitation — permissions.mdx says the map is "currently ignored on macOS".

### To Reproduce

1. Build a macOS app with `NSCameraUsageDescription` and `NSMicrophoneUsageDescription` in its `Info.plist`.
2. Create a window that should refuse capture:
   ```go
   app.Window.NewWithOptions(application.WebviewWindowOptions{
       Permissions: map[application.PermissionType]application.Permission{
           application.PermissionMicrophone: application.PermissionDeny,
           application.PermissionCamera:     application.PermissionDeny,
       },
   })
   ```
3. Call `navigator.mediaDevices.getUserMedia({audio: true, video: true})` from it.
4. WebKit prompts, and capture proceeds if the user accepts — the `PermissionDeny` never applies. Configured with `PermissionAllow`, the prompt still appears.

Found by reading the source rather than running it: `git grep requestMediaCapturePermission v3/` comes back empty, and `webview_window_darwin.go:209` is the only `setUIDelegate:`.

### Expected behaviour

The three-way policy the other platforms implement, mapped onto `WKPermissionDecision`:

| `Permission` | macOS decision |
| --- | --- |
| `PermissionDefault` | `WKPermissionDecisionPrompt` — unchanged from today |
| `PermissionAllow` | `WKPermissionDecisionGrant` |
| `PermissionDeny` | `WKPermissionDecisionDeny` |

TCC still gates the device on top, so `PermissionAllow` removes WebKit's prompt but not the system one. Camera and microphone only, matching the current Linux scope — the other three types have no `WKUIDelegate` equivalent.

### Additional context

#6050 implements this.

### System Details

```
Wails CLI      v3.0.0-beta.13
Go             go1.26.2
OS             macOS 26.6.2 (25G83), darwin/arm64, Apple M4 Pro
Xcode          26.6 (17F113), cli tools 2416
CGO_ENABLED    1
```

fan711:
# Description

The cross-platform `Permissions` option has no effect on macOS. It is honoured on Linux (#5552) and on Windows through WebView2, but `WebviewWindowDelegate` — which is set as the webview's UI delegate — does not implement `webView:requestMediaCapturePermissionForOrigin:initiatedByFrame:type:decisionHandler:`, so nothing on that platform ever consults it.

Left unimplemented, WebKit takes the request's default action, which on Cocoa is `promptForGetUserMedia` ([`UserMediaPermissionRequestProxy::doDefaultAction`](https://github.com/WebKit/WebKit/blob/main/Source/WebKit/UIProcess/UserMediaPermissionRequestProxy.cpp)). Capture therefore works today — but macOS behaves as though the option were permanently `PermissionDefault`, and `PermissionAllow` and `PermissionDeny` are silently ignored.

Implementing the delegate against the same option gives them their meaning there:

| `Permission` | macOS decision |
| --- | --- |
| `PermissionDefault` | `WKPermissionDecisionPrompt` — WebKit's own prompt, which is what already happens today |
| `PermissionAllow` | `WKPermissionDecisionGrant` |
| `PermissionDeny` | `WKPermissionDecisionDeny` |

So a window that does not configure `Permissions` behaves exactly as before. `PermissionAllow` is the case that matters in practice: for a window that only ever loads the app's own content, WebKit's prompt asks a question the user has already answered by clicking the button that called `getUserMedia`, and granting leaves just the one system prompt that actually gates the device.

A request for the camera and the microphone together (`WKMediaCaptureTypeCameraAndMicrophone`) gets a single answer, and it is no more permissive than either half on its own: deny beats prompt beats grant.

Two notes on the implementation:

- `Permission` and `WKPermissionDecision` happen to agree case for case (0/1/2). The mapping is still written out rather than cast — an agreement between two unrelated ABIs is not something to build on.
- The delegate method is macOS 12+, so it carries `API_AVAILABLE(macos(12.0))` and is never called below that. macOS 11 is unaffected.

This does not replace `NSCameraUsageDescription` / `NSMicrophoneUsageDescription`, or the sandbox's `com.apple.security.device.audio-input` / `.camera` where an app is sandboxed. Those still gate the device; this only decides whether the request reaches them.

Fixes #6067 — filed by @Grantmartin2002 in review. This was found while adding voice recording to a Wails v3 app and had no issue behind it until then.

Related: **#3735** is useful background on the `Info.plist` / entitlement side. **#4270** is *not* fixed by this and I am not claiming it is — that report is about `wails3 dev`, and the triage there (an unbundled dev binary that TCC cannot attribute) is a separate problem with its own fix.

## Type of change

No WEP: this adds no public API and proposes no new behaviour. `Permissions`, `PermissionType` and `Permission` are all existing public API, and the default case is unchanged. Happy to convert it if you read it differently.

## Documentation

`features/windows/permissions.mdx` said the map was ignored on macOS, in four places — the platform section, the media-capture pattern, the support matrix and the troubleshooting entry. All four now describe the two layers: the policy answers WKWebView's request, and TCC still gates the device underneath it, so `PermissionAllow` drops WebKit's prompt but not the system one. The matrix reads like Linux — camera and microphone ✅, the other three ❌ not yet, since they have no `WKUIDelegate` equivalent. The `id/` translation is updated in the same commit, which is what the recent history of that tree does.

The page also warns about a trap this option creates on macOS, which @Grantmartin2002 raised: a capability that reaches AVFoundation with no usage description string does not fail — the OS terminates the app. `PermissionDefault` is the zero value, so `{PermissionMicrophone: PermissionAllow}` on its own leaves the camera on WebKit's prompt, and an app declaring only `NSMicrophoneUsageDescription` dies if the user accepts it. The advice is to set `PermissionDeny` explicitly for anything you have no usage string for.

# How Has This Been Tested?

**Not on macOS — I do not have Mac hardware, and I would rather say so than tick a box.** Please treat the macOS column as unverified.

@Grantmartin2002 has since reviewed it there and reports `go vet`, `go build` and `go test ./pkg/application/` clean, the `WKPermissionDecision` cases matching the SDK header, and no warnings from recompiling the `.m` at the package's 10.13 deployment target with `-Wunguarded-availability-new`.

What I did verify, on Linux:

- `permissions_darwin_test.go` (new) pins `captureDecisionPrompt`/`Grant`/`Deny` to 0/1/2 — that is the thing which makes the hand-written mapping safe rather than lucky. It also covers the strictest-wins rule from both argument orders, the `Permission` → decision mapping, and the lookup: an unset entry, a window with no map at all, and a window the manager no longer knows about. Same `darwin && !ios && !server` tags as the code it tests.
- To run it here, the test file and the non-cgo half of `permissions_darwin.go` were temporarily copied into `pkg/application` without the build tags and run on Linux against the real `App` and `WebviewWindow`. All cases pass. The copies were then deleted — none of that is in the branch.
- The Go half type-checks and builds. `permissions_darwin.go` was temporarily retagged `linux && cgo && !android && !server` and built as part of `pkg/application`, which exercises the cgo signature, the `*WebviewWindow` cast and the `Permissions` lookup. Clean.
- `gofmt` clean, and the Linux and Windows paths are untouched.
- Both permissions pages compile as MDX (`@mdx-js/mdx`), including the new `<Aside>` and the code block inside it.

The Objective-C in `webview_window_darwin.m` has been reviewed but not compiled by me.

- [ ] Windows
- [ ] macOS
- [x] Linux

Linux: Debian 13, GTK 4.18.6, WebKitGTK 2.52.3 — only to confirm nothing regressed there, since this change does not touch Linux.

## Test Configuration

`wails doctor` is not meaningful here — this is verified by compilation and unit tests rather than at runtime, and not on the platform it targets. Built against `master` (f2260d15b).

# Checklist:

On the unticked boxes:

- **v3 changelog.** The checklist says v3 entries are added automatically, but `CONTRIBUTING.md` asks for `v3/UNRELEASED_CHANGELOG.md` to be updated by hand. I followed CONTRIBUTING and it is a separate commit — say the word and I will drop it.
- **Warnings.** I cannot compile the Objective-C, so I cannot claim this myself. @Grantmartin2002 has confirmed it on macOS.

---

*Edited: the first version of this description claimed WebKit **denies** capture when the delegate is unimplemented, and that `getUserMedia` therefore never worked on macOS. That was wrong — @coderabbitai caught it. `doDefaultAction` prompts on Cocoa, and has done so at least as far back as the safari-613 branch. The fix is real but narrower than I first described, and I have corrected the code comments and the changelog entry to match.*

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed macOS `getUserMedia` support for camera and microphone access in web views on macOS 12 and later.
  - Media permissions now honor configured settings consistently with Linux and Windows.
  - Requests for multiple media devices apply the strictest applicable permission decision.

- **Documentation**
  - Updated macOS permissions guidance, support details, troubleshooting, and usage-description requirements.
  - Added a changelog entry documenting the macOS media-capture permission improvement.

### DIFF
--- a/docs/src/content/docs/features/windows/permissions.mdx
+++ b/docs/src/content/docs/features/windows/permissions.mdx
@@ -77,11 +77,21 @@ WebView2 has a native permission prompt and a per-kind permission API. All five 
 
 This means if you configure `Permissions` at all on Windows, any capability you don't explicitly list will show a prompt rather than being silently allowed. Set the capabilities you need explicitly.
 
-### macOS (TCC)
+### macOS (WKWebView + TCC)
 
-macOS manages camera, microphone, geolocation, and notification access through its system privacy framework. The OS prompt appears automatically the first time web content requests a capability, and the user's choice is remembered per-app in System Settings → Privacy & Security.
+Two layers have to agree on macOS. WKWebView asks the application before it starts a capture session, and that request is what the `Permissions` map answers. Beneath it, the system privacy framework (TCC) gates the device itself: the OS prompt appears the first time the app actually reaches for the camera or the microphone, and the user's choice is remembered per-app in System Settings → Privacy & Security.
 
-This works correctly without any `Permissions` configuration. The map is **currently ignored on macOS** — all requests go through TCC regardless of what you set. The practical gap is that `PermissionDeny` has no effect on macOS: you cannot block a webview from using a capability that TCC has already granted at the system level.
+Wails handles **camera and microphone** requests on macOS 12 and later. Geolocation, notifications, and clipboard read have no `WKUIDelegate` equivalent, so they are not wired up and are left to TCC alone — as on Linux, the policy you set for them has no effect.
+
+| Policy | Camera / Microphone | Geolocation, Notifications, Clipboard |
+|---|---|---|
+| `PermissionDefault` | WebKit shows its own permission prompt | TCC only |
+| `PermissionAllow` | WebKit's prompt is skipped — **TCC still applies** | TCC only |
+| `PermissionDeny` | Denied before the device is touched | TCC only |
+
+`PermissionAllow` grants the webview's request, not the device. The first capture still raises the TCC prompt, and an app the user has denied in System Settings stays denied — no application can grant itself device access. What `PermissionAllow` removes is WebKit's own prompt in front of that.
+
+Below macOS 12 the delegate method does not exist, so the map is ignored there and every request falls back to WebKit's prompt.
 
 Ensure your `Info.plist` includes the appropriate usage description keys:
 
@@ -91,6 +101,21 @@ Ensure your `Info.plist` includes the appropriate usage description keys:
 <key>NSCameraUsageDescription</key>
 <string>Used for video calls</string>
 ```
+
+<Aside type="caution" title="Deny what your Info.plist does not declare">
+
+A capability that reaches AVFoundation without its usage description key does not fail — macOS terminates the app.
+
+`PermissionDefault` is the zero value, so `{PermissionMicrophone: PermissionAllow}` on its own leaves the camera on WebKit's prompt. If the user accepts that prompt and the app declares only `NSMicrophoneUsageDescription`, the app dies. Set `PermissionDeny` explicitly for every capability you have no usage description for:
+
+```go
+Permissions: map[application.PermissionType]application.Permission{
+    application.PermissionMicrophone: application.PermissionAllow,
+    application.PermissionCamera:     application.PermissionDeny,
+},
+```
+
+</Aside>
 
 ## Common Patterns
 
@@ -107,7 +132,7 @@ Permissions: map[application.PermissionType]application.Permission{
 
 On **Linux** this explicitly allows both devices; other capabilities remain denied.  
 On **Windows** this allows both; any other capability you don't list will show a native prompt.  
-On **macOS** this has no effect; TCC handles everything.
+On **macOS** this allows both at the WebKit layer, so no browser prompt appears; TCC still asks for the devices themselves on first use, and both usage description keys must be present in `Info.plist`.
 
 ### Deny media capture on Linux
 
@@ -185,11 +210,13 @@ The evaluation order on Windows is:
 
 | Capability | Linux | Windows | macOS |
 |---|---|---|---|
-| Microphone | ✅ | ✅ | TCC only |
-| Camera | ✅ | ✅ | TCC only |
-| Geolocation | ❌ not yet | ✅ | TCC only |
-| Notifications | ❌ not yet | ✅ | TCC only |
-| Clipboard Read | ❌ not yet | ✅ | TCC only |
+| Microphone | ✅ | ✅ | ✅ (macOS 12+) |
+| Camera | ✅ | ✅ | ✅ (macOS 12+) |
+| Geolocation | ❌ not yet | ✅ | ❌ not yet |
+| Notifications | ❌ not yet | ✅ | ❌ not yet |
+| Clipboard Read | ❌ not yet | ✅ | ❌ not yet |
+
+Where macOS is ✅, the policy answers WebKit's request; TCC still gates the device on top of it. Where it is ❌, the capability is left to TCC alone.
 
 ## Troubleshooting
 
@@ -203,7 +230,11 @@ Once any entry is present in `Permissions`, Wails no longer sets the blanket `Al
 
 **macOS permissions aren't working**
 
-The `Permissions` map has no effect on macOS. Ensure your `Info.plist` includes the correct usage description keys (`NSMicrophoneUsageDescription`, `NSCameraUsageDescription`, etc.) and that the user has granted access in System Settings → Privacy & Security.
+`Permissions` covers camera and microphone on macOS 12 and later; geolocation, notifications, and clipboard read are not wired up yet and ignore the policy. Where it is honoured, TCC still gates the device on top: `PermissionAllow` removes WebKit's prompt but not the system one. Ensure your `Info.plist` includes the correct usage description keys (`NSMicrophoneUsageDescription`, `NSCameraUsageDescription`) and that the user has granted access in System Settings → Privacy & Security.
+
+**My macOS app quits when web content asks for the camera or microphone**
+
+macOS terminates an app that reaches AVFoundation without the matching usage description key. Add the key, or set `PermissionDeny` for that capability so the request never gets that far — `PermissionDefault` leaves it on WebKit's prompt, which the user can accept.
 
 **Geolocation/notifications/clipboard have no effect on Linux**
 
--- a/docs/src/content/docs/id/features/windows/permissions.mdx
+++ b/docs/src/content/docs/id/features/windows/permissions.mdx
@@ -77,11 +77,21 @@ WebView2 memiliki prompt permission native dan API permission per-jenis. Kelima 
 
 Artinya jika Anda mengonfigurasi `Permissions` di Windows, kemampuan yang tidak Anda cantumkan secara eksplisit akan menampilkan prompt alih-alih diizinkan diam-diam. Setel kemampuan yang Anda butuhkan secara eksplisit.
 
-### macOS (TCC)
+### macOS (WKWebView + TCC)
 
-macOS mengelola akses kamera, mikrofon, geolokasi, dan notifikasi melalui framework privasi sistem. Prompt OS muncul otomatis saat konten web pertama kali meminta kemampuan, dan pilihan pengguna diingat per-aplikasi di System Settings → Privacy & Security.
+Dua lapisan harus sepakat di macOS. WKWebView menanyakan aplikasi sebelum memulai sesi capture, dan permintaan itulah yang dijawab oleh map `Permissions`. Di bawahnya, framework privasi sistem (TCC) menjaga perangkatnya sendiri: prompt OS muncul saat aplikasi benar-benar mengakses kamera atau mikrofon untuk pertama kali, dan pilihan pengguna diingat per-aplikasi di System Settings → Privacy & Security.
 
-Ini berfungsi dengan benar tanpa konfigurasi `Permissions`. Map **saat ini diabaikan di macOS** — semua permintaan melalui TCC terlepas dari yang Anda setel. Celah praktisnya: `PermissionDeny` tidak berpengaruh di macOS — Anda tidak dapat memblokir webview menggunakan kemampuan yang TCC sudah berikan di level sistem.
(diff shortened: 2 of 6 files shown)


fan711:
Both findings were right. Thanks — the first one was load-bearing and I had it backwards.

**On the WebKit fallback.** I claimed WebKit *denies* a capture request when the delegate method is unimplemented. It does not. `UIDelegate::UIClient::decidePolicyForUserMediaPermissionRequest` calls `request->doDefaultAction()`, and on Cocoa that is:

```cpp
void UserMediaPermissionRequestProxy::doDefaultAction()
{
#if ENABLE(MEDIA_STREAM) && PLATFORM(COCOA)
    if (requiresDisplayCapture())
        promptForGetDisplayMedia(UserMediaDisplayCapturePromptType::UserChoose);
    else
        promptForGetUserMedia();
#else
    deny();
#endif
}
```

`deny()` is the non-Cocoa branch. I checked `main`, `safari-613-branch` and `safari-7614-branch` — it has prompted on Cocoa throughout, so my claim was not merely out of date, it was wrong.

That changes what this PR is. `getUserMedia` **does** work on macOS today via WebKit's own prompt; what is broken is narrower and still real: the window's `Permissions` are never consulted, so macOS behaves as though the option were permanently `PermissionDefault` and `PermissionAllow` / `PermissionDeny` are silently ignored. Since `PermissionDefault` maps to `WKPermissionDecisionPrompt`, a window that does not configure the option behaves exactly as before.

The code did not need to change — the mapping was already right — but the reasoning around it did. I have rewritten the comments in both files, the commit message, the changelog entry and the PR description, and force-pushed.

**On the changelog scope.** Applied. It now reads:

> Fix the `Permissions` option being ignored on macOS 12 and later: the `WKUIDelegate` media-capture method is now implemented, so `PermissionAllow` and `PermissionDeny` apply to camera and microphone requests as they do on Linux and Windows. `NSCameraUsageDescription` / `NSMicrophoneUsageDescription` and, where sandboxed, the matching device entitlements are still required

**One thing I withdrew on my own account.** The first description speculated that the triage on #4270 might be incomplete, on the strength of the same wrong premise. It was not incomplete — @leaanthony's explanation holds, and I have removed that passage. Apologies for the noise.

Standing caveat unchanged: I have no Mac and the Objective-C has never been compiled, only reviewed. Verification on macOS 12+ would still be very welcome.

Grantmartin2002:
The code looks right to me — built it on macOS 26.6/arm64, and the hand-written mapping matches `WKPermissionDecision` in the SDK header. One thing I think blocks it though.

The docs page for this option still says the feature doesn't exist. `docs/src/content/docs/features/windows/permissions.mdx` has "The map is **currently ignored on macOS**", a support matrix listing macOS as "TCC only" for camera and microphone, and a troubleshooting entry telling people it has no effect. Worth folding in the point from your description while you're there — TCC still gates the device, so `PermissionAllow` drops WebKit's prompt but not the system one.

A test pinning `captureDecisionPrompt/Grant/Deny` to 0/1/2 would be worth having too. That's the invariant you avoided casting for.

No linked issue, so I opened #6067 if you want to `Fixes` it.

Grantmartin2002:
One more for the docs change, if you take it — on macOS, `Prompt` for a capability whose `Info.plist` usage string is missing just gets the user a dialog the platform can't honour. `NSCameraUsageDescription` / `NSMicrophoneUsageDescription` have to be declared, and a missing key is a TCC violation rather than a graceful denial. #3735 was closed by documenting exactly those two keys, and #4270 is the neighbouring dev-mode case where TCC attributes to the terminal running the binary instead of the app.

So `{PermissionMicrophone: PermissionAllow}` on its own is a bit of a trap: camera stays unset → `Prompt`, and an app that only declared the mic key gets asked for something it can't deliver. Harmless on Windows and Linux, macOS-only, and it only becomes reachable once there's a map to configure.

Something like "set `PermissionDeny` explicitly for capabilities you haven't declared a usage string for" would cover it. We got bitten by this in an app that grants mic and denies everything else, which is why I noticed.

fan711:
Thanks — that is a thorough pass, and the docs gap was a fair block. All of it is in, in two commits on top of the two you reviewed.

**Docs.** `permissions.mdx` is rewritten in the four places you listed, and the `id/` translation is updated in the same commit — that is what the recent history of that tree does (#5883 is the pattern).

The macOS section is two layers now rather than one: the `Permissions` map answers WKWebView's request, and TCC gates the device underneath it. `PermissionAllow` drops WebKit's prompt but not the system one, and no app can grant itself device access — that is stated in the section text, in the policy table, and again in troubleshooting, because you are right that it is the thing people will get wrong. The matrix reads ✅ / ✅ / ❌ ❌ ❌ for macOS, mirroring Linux, with a line under it saying what ✅ means there. It also now says that below macOS 12 the delegate method does not exist, so the map is ignored on those versions.

**The `Info.plist` trap.** Added, as a caution in the macOS section and as a troubleshooting entry ("My macOS app quits when web content asks for the camera or microphone"). Thank you for that one — I would have shipped the option that makes it reachable without ever mentioning it. The page says to set `PermissionDeny` explicitly for any capability you have no usage string for, and uses `{Microphone: Allow, Camera: Deny}` as the example, since `PermissionDefault` being the zero value is exactly what turns a half-configured map into a crash rather than an oversight.

**Tests.** `v3/pkg/application/permissions_darwin_test.go`, same build tags as the code it tests. `TestCaptureDecision_Constants` pins Prompt/Grant/Deny to 0/1/2 against `WKPermissionDecision` — agreed that this is the thing which makes writing the mapping out safe rather than merely tidier. Beyond that: the strictest-wins rule from both argument orders, the `Permission` → decision mapping, and the lookup (unset entry, window with no map, window the manager no longer knows about).

To get the camera-and-microphone combination reachable from a plain Go test, `resolveMediaCapturePermission` is now only the cgo boundary — one line down to `mediaCaptureDecision(uint, bool, bool)`, which is what the tests call. No behaviour change.

I still cannot run them on macOS. They pass here, but by copying the test file and the non-cgo half of `permissions_darwin.go` into `pkg/application` without the build tags and running that on Linux against the real `App` and `WebviewWindow`; the copies were deleted and none of that is in the branch. So `go test ./pkg/application/` on a Mac is worth one more run from you, if you are still willing.

**Template and issue.** The body was on the template, but the documentation box genuinely was not ticked — that one and the tests box are now, and the description says `Fixes #6067`. Thanks for filing it; it states the problem better than my description did.

No need to push the docs yourself. If the TCC-layering wording reads wrong to you on the rendered page, say so and I will take it.

leaanthony:
Ported the authored macOS permissions documentation to M-Press in 995d9618eb085050032159e8cd86f10a8bca7e32, including the existing Indonesian changes and updates to all other eight translations. The ten pages now describe the WebKit/TCC layers, macOS 12 requirement, usage-description warning, permission matrix, and troubleshooting consistently. The new microphone-only Go example matches across every language.

Dependency: merged #6137 at 7b0f6dc094bfe7bf693df0320338a9bfc5f86ec8; the PR now targets `codex/mpress-markdown`. Original implementation edits are preserved, with current-base changes merged safely. Removed the obsolete Indonesian Starlight source left by the modify/delete conflict; its authored content is ported into Markdown. No translation sidecar state or audit settings were changed. Pushed fast-forward to the original fork branch as taliesin-ai.

Validation: M-Press 1.0.17 `build --strict --no-purge-css`, link/asset check, and rendered-site check passed (2,010 content pages; 2,011 HTML pages; zero site errors). The documented permission-map example compiled and passed its local Go check on Linux. Manually reviewed the docs, locale/code parity, and the preserved implementation diff. Native WKWebView/TCC runtime tests require macOS and were not run on this Linux host.

Known gate: the global translation audit remains failing (7,676 findings versus 7,678 on the pinned migration base); findings outside the permissions page are unchanged. The migration-wide audit blocker remains separate work, with the gate enabled. `coderabbit --plain` was attempted but unsupported; the one narrow fallback review hit the shared rate limit. No usage credits were spent.

leaanthony:
Stefan, I am sorry. PR #6050 contained the actual Darwin implementation and tests, but it was closed when its temporary migration base branch was deleted after the docs merge, which stranded your work. I also mischaracterized it earlier as documentation-only — that was wrong.\n\nI recovered the code, tests, changelog, M-Press docs, and translation-audit updates in replacement PR #6144: https://github.com/wailsapp/wails/pull/6144\n\nThe new PR preserves you as co-author.

PR #6078, opened 2026-09-05
fix(v3): add SingleInstance.OnSecondInstanceExit for pre-exit cleanup

### LINKED ISSUE #6052: [v3] SingleInstance: os.Exit inside application.New skips the caller's defers (silent cleanup loss)
### Wails version family

v3 (alpha/beta)

### Exact Wails version

v3.0.0-beta.15 (sites verified identical at master `9d49e8e` and at beta.9/beta.12)

### Operating System

All (the `os.Exit` call is in shared code; darwin/linux have one extra exit site each)

### Description

When `SingleInstance` detects an already-running instance, `application.New` calls `os.Exit` directly:

- [`v3/pkg/application/application.go:206-218`](https://github.com/wailsapp/wails/blob/9d49e8e0ef0dfc40328d435bff8ea0f3d72cfe80/v3/pkg/application/application.go#L206-L218) — on `alreadyRunningError`: `notifyFirstInstance()`, then `os.Exit(appOptions.SingleInstance.ExitCode)`
- plus `single_instance_darwin.go:82` and `single_instance_linux.go:99`

`os.Exit` terminates without running deferred functions (Go spec), so any cleanup the caller scheduled **before** calling `New()` — flushing a write-ahead file, releasing a lock file, closing a log — is silently skipped in the second process. `New()` looks like an ordinary constructor, so having opened resources before it is natural; the constraint is invisible in the API.

In our production app ([gadak](https://github.com/midagedev/gadak/blob/main/desktop/main.go)) the second-instance exit was skipping a persistence flush scheduled with `defer`. We had to reorder initialization so nothing requiring cleanup exists before `New()`, and pin that ordering with a test — that works, but every `SingleInstance` adopter has to discover it the same way.

What this issue is **not** asking: `SingleInstance.ExitCode` already covers the exit *value* (#3723), and the second instance correctly must not continue into `Run()` and show a window — no change to that. The ask is a seam for cleanup, in either shape:

1. `New()` returns a distinguished error (e.g. `errors.Is(err, application.ErrAlreadyRunning)`) after `notifyFirstInstance()`, and the caller exits — possibly opt-in to avoid breaking apps that don't check the error; or
2. a hook, e.g. `SingleInstance.OnSecondInstanceExit func()`, invoked after `notifyFirstInstance()` and before `os.Exit`.

Related: #3301 ("in idiomatic Go we return an error"); #4066 already converted service-startup fatals into returned errors; #5170 recently touched this path. Happy to send a PR for whichever shape you prefer.

### To Reproduce

1. `SingleInstance` enabled; in `main()`, open any resource needing cleanup and schedule it with `defer` (e.g. `defer f.Sync()`), **then** call `application.New(...)`.
2. Start the app once, then start a second instance.
3. The second process exits with `ExitCode` — the deferred cleanup never runs (observable with a defer that writes a sentinel file: the file is not written).

### Expected behaviour

A way for the second instance to run its cleanup before terminating — via a returned error or a pre-exit hook — while wails keeps owning the "second instance does not open a window" behavior.

### Attempted Fixes

Reordered our initialization so that nothing requiring cleanup exists before `New()`, pinned with a test. Works, but is a per-app rediscovery of an API-invisible constraint.

4RH1T3CT0R7:
When SingleInstance finds the lock already held, the second process ends with os.Exit inside application.New, so anything the caller deferred before New (flushing a file, releasing a lock, closing a log) never runs. The darwin and linux notify implementations also had their own os.Exit, giving three exit sites for the same path.

This adds SingleInstanceOptions.OnSecondInstanceExit, called in the second instance after the first has been notified and right before os.Exit(ExitCode), and collapses the three exit sites into one manager method. Apps that don't set the hook behave exactly as before. A subprocess test checks the hook runs after notification and the process exits with ExitCode; it skips on macOS, where notifyFirstInstance runs an AppKit loop to capture the launch URL. The Linux notify overlaps with the open #5973, so whichever lands second needs a small rebase.

Fixes #6052

## Summary by CodeRabbit

- **New Features**
  - Added an `OnSecondInstanceExit` callback for cleanup or resource handling before a second application instance exits.
  - Configurable exit codes are now applied after the second instance notifies the primary instance.

- **Documentation**
  - Updated the Single Instance guide with callback behavior, exit-code handling, and guidance for flushing or releasing resources before termination.

### DIFF
--- a/docs/src/content/docs/guides/single-instance.mdx
+++ b/docs/src/content/docs/guides/single-instance.mdx
@@ -42,6 +42,8 @@ The `SingleInstanceOptions` struct has the following fields:
     - `WorkingDir`: The working directory of the second instance
     - `AdditionalData`: Any additional data passed from the second instance (if provided)
 - `AdditionalData`: Optional map of string key-value pairs that will be passed to the first instance when subsequent instances are launched
+- `ExitCode`: The exit code the second instance terminates with after notifying the first instance
+- `OnSecondInstanceExit`: A callback run in the second instance after it has notified the first instance and just before it exits. The exit uses `os.Exit`, so deferred functions in `main` never run; release or flush anything you opened before `application.New` here instead
 
 :::danger[Warning]
 The Single Instance feature implements an optional encryption protocol using AES-256-GCM. Without encryption enabled,
--- a/v3/pkg/application/application.go
+++ b/v3/pkg/application/application.go
@@ -208,11 +208,7 @@ func New(appOptions Options) *App {
 		manager, err := newSingleInstanceManager(result, appOptions.SingleInstance)
 		if err != nil {
 			if errors.Is(err, alreadyRunningError) && manager != nil {
-				err = manager.notifyFirstInstance()
-				if err != nil {
-					globalApplication.error("failed to notify first instance: %w", err)
-				}
-				os.Exit(appOptions.SingleInstance.ExitCode)
+				manager.exitSecondInstance()
 			}
 			result.fatal("failed to initialize single instance manager: %w", err)
 		} else {
--- a/v3/pkg/application/single_instance.go
+++ b/v3/pkg/application/single_instance.go
@@ -40,6 +40,13 @@ type SingleInstanceOptions struct {
 	// ExitCode is the exit code to use when the second instance exits
 	ExitCode int
 
+	// OnSecondInstanceExit is called in the second instance after it has
+	// notified the first instance and just before it exits with ExitCode.
+	// The exit uses os.Exit, so deferred functions in the caller never run:
+	// anything opened before New() that needs flushing or releasing should
+	// be handled here.
+	OnSecondInstanceExit func()
+
 	// EncryptionKey is a 32-byte key used for encrypting instance communication
 	// If not provided (zero array), data will be sent unencrypted
 	EncryptionKey [32]byte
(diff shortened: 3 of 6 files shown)


leaanthony:
Ported this PR's ExitCode/OnSecondInstanceExit documentation to M-Press in 397faec2df7ba23f7fda41848ba798c3316f06ad and updated the corresponding existing page in all nine languages (de, fr, id, ja, ko, pt, ru, zh-cn, zh-tw).

Dependency: merged #6137 at 7b0f6dc094bfe7bf693df0320338a9bfc5f86ec8 and changed this PR's base to `codex/mpress-markdown`. The only code conflict was composed mechanically: Linux notify now uses the current base's `dbusInterface` while retaining this PR's return-to-centralized-cleanup behavior instead of calling os.Exit there. Comparison of implementation edits confirmed preservation of the PR changes with that interface-name adjustment. No Starlight sources remain; no translation sidecar state or audit settings were changed. Pushed fast-forward to the original fork branch as taliesin-ai.

Validation: M-Press 1.0.17 strict build, link/asset check, and rendered-site check passed (2,010 content pages; 2,011 HTML pages; zero site errors). Linux `go test ./pkg/application -run 'Test.*(SingleInstance|SecondInstance)' -count=1` passed, including cleanup/exit and D-Bus name tests. A local Go example check verified the documented option fields/callback. Manual scoped diff review and whitespace checks passed.

Known gate: the global translation audit remains failing with 7,678 findings, matching the pinned migration baseline count; findings outside this page are unchanged. The inherited migration audit blocker is separate work and the gate remains enabled. `coderabbit --plain` was attempted but unsupported; the single narrow fallback review hit the shared rate limit. No usage credits were spent.

4RH1T3CT0R7:
Thanks for carrying this over to M-Press and sorting out the notify conflict.

PR #6084, opened 2026-09-07
fix(v3): add AutostartManager.DisableWithOptions for identifier-targeted removal

### LINKED ISSUE #6057: v3 Autostart Disable cannot target a configured identifier
## Problem

`AutostartManager.EnableWithOptions` accepts a stable `AutostartOptions.Identifier`, but `Disable` discovers registrations by the current executable instead of deleting that identifier. On Windows, a Run value left by an older installation path cannot be removed after the executable moves. The same discovery-based mismatch exists for Linux desktop files and macOS LaunchAgents.

A consumer can work around this by calling `EnableWithOptions` and then `Disable`, but that is not failure-atomic: if enable succeeds and disable fails, a previously disabled app becomes registered at login. Retrying the same sequence cannot reliably restore the disabled state.

## Reproduction

1. Write a Windows Run value under a stable identifier, with the command pointing to an old executable path.
2. Start the application from a new path.
3. Call `AutostartManager.Disable()`.
4. The stable value remains because `windowsAutostart.find()` matches only the current executable.

## Proposed API

Add `DisableWithOptions(AutostartOptions)` (or an equivalent identifier-targeted operation) that validates and removes the exact registration identity used by `EnableWithOptions`. Keep existing `Disable()` behavior for compatibility.

## Acceptance

- Windows removes the exact registry value even when its command points to a moved executable.
- Linux removes the exact `.desktop` filename.
- macOS removes the exact LaunchAgent label while preserving SMAppService cleanup.
- Empty options retain the existing default behavior.
- Invalid identifiers are rejected consistently.
- Native/platform tests cover idempotency and moved executable state.

hellopoisonx:
# Description

`AutostartManager.EnableWithOptions` accepts a stable `Identifier`, but `Disable()` discovers registrations by matching the **current executable path** (Linux `findDesktopFile` matches `Exec=`, Windows `find()` matches the registry command's first token, macOS `findLaunchAgent` matches `ProgramArguments[0]`). Once the executable moves or is renamed, a registration created under a stable identifier can never be removed, and the `Enable`→`Disable` workaround is not failure-atomic.

This adds `AutostartManager.DisableWithOptions(AutostartOptions)`, which removes the exact registration identity — the registry value name on Windows, the `.desktop` filename on Linux, the LaunchAgent label on macOS — regardless of where the registered command points:

- `Disable()` keeps its current discovery-based behaviour (empty options are equivalent).
- Identifier-targeted removal is idempotent (removing an absent registration returns nil).
- Invalid identifiers are rejected consistently with `EnableWithOptions` via the existing `validateAutostartIdentifier`.
- On macOS, identifier-targeted removal does not touch the bundle-keyed `SMAppService` registration (it is not identified by `Identifier`); the legacy dual-path cleanup is preserved for empty options.

On process: CONTRIBUTING puts "a new public API" under WEP-required and "a confirmed bug fix" under no-WEP. This is a bug fix whose remedy adds one method mirroring `EnableWithOptions`, following the precedent of #6078 (new public `SingleInstanceOptions.OnSecondInstanceExit` for #6052, standard PR, no WEP). Happy to convert to a WEP first if maintainers prefer.

Fixes #6057

## Type of change

- [ ] Breaking change (fix or feature which would cause existing functionality to not work as expected)

# How Has This Been Tested?

New per-platform tests (`TestLinuxAutostartDisableWithOptions`, `TestWindowsAutostartDisableWithOptions`, `TestDarwinAutostartDisableWithOptions`) create a stale registration whose command points at a nonexistent path, assert that discovery-based `disable` cannot see it, that `DisableWithOptions{Identifier}` removes exactly that entry while an unrelated entry survives, that a second call returns nil, and that invalid identifiers are rejected.

- [ ] Windows
- [ ] macOS
- [x] Linux

Linux: Arch Linux (kernel 7.2.3-arch1, x86_64). Verified with `go test ./pkg/application/ -run Autostart -count=1 -v` (all pass, including `TestLinuxAutostartRoundTrip` / `CustomIdentifier` regressions), `go build` / `go vet` of the package, and `GOOS=windows go build ./pkg/application/` cross-compilation (the Windows test file also type-checks under `GOOS=windows go vet`). Windows/macOS test binaries are compile-excluded on Linux and run in this repo's CI matrix.

## Test Configuration

Arch Linux x86_64, kernel 7.2.3-arch1-2, Go toolchain (go1.x), GTK4 dev headers present; `wails doctor` unavailable in this environment — the package-level `go build`/`go vet`/`go test` commands above were used instead. CodeRabbit CLI review: no findings.

# Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added the ability to disable autostart registrations using an exact identifier, even when the registered command has moved or been renamed.
  * Repeated removal of an absent registration is safely treated as a no-op.
  * Existing discovery-based disable behavior remains unchanged when no identifier is provided.

* **Documentation**
  * Added usage guidance for identifier-based autostart removal.

### DIFF
--- a/docs/src/content/docs/features/autostart/basics.mdx
+++ b/docs/src/content/docs/features/autostart/basics.mdx
@@ -67,6 +67,14 @@ Removes the autostart registration. Returns `nil` if the application wasn't regi
 func (m *AutostartManager) Disable() error
 ```
 
+### `DisableWithOptions`
+
+Removes the registration created under a specific `AutostartOptions.Identifier`, regardless of where the registered command points — use it to clean up a registration left behind by a moved or renamed executable. An empty `Identifier` keeps the discovery-based behaviour of `Disable`. Idempotent: removing an absent registration returns nil.
+
+```go
+func (m *AutostartManager) DisableWithOptions(opts AutostartOptions) error
+```
+
 ### `IsEnabled`
 
 Reports whether a registration exists. Fast — doesn't validate the registered path.
--- a/v3/UNRELEASED_CHANGELOG.md
+++ b/v3/UNRELEASED_CHANGELOG.md
@@ -17,6 +17,7 @@ After processing, the content will be moved to the main changelog and this file 
 
 ## Added
 <!-- New features, capabilities, or enhancements -->
+- Add `AutostartManager.DisableWithOptions` to remove an autostart registration by its exact identifier, so entries left by moved or renamed executables can be cleaned up (#6057)
 
 ## Changed
 <!-- Changes in existing functionality -->
--- a/v3/pkg/application/autostart.go
+++ b/v3/pkg/application/autostart.go
@@ -162,6 +162,6 @@ type AutostartStatus struct {
 
 type autostartImpl interface {
 	enable(opts AutostartOptions) error
-	disable() error
+	disable(opts AutostartOptions) error
 	status() (AutostartStatus, error)
 }
--- a/v3/pkg/application/autostart_android.go
+++ b/v3/pkg/application/autostart_android.go
@@ -7,7 +7,7 @@ type unsupportedAutostart struct{}
 func newAutostartImpl(_ *App) autostartImpl { return unsupportedAutostart{} }
 
 func (unsupportedAutostart) enable(AutostartOptions) error  { return ErrAutostartNotSupported }
-func (unsupportedAutostart) disable() error                 { return ErrAutostartNotSupported }
+func (unsupportedAutostart) disable(AutostartOptions) error { return ErrAutostartNotSupported }
 func (unsupportedAutostart) status() (AutostartStatus, error) {
 	return AutostartStatus{}, ErrAutostartNotSupported
 }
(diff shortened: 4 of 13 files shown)


leaanthony:
Ported this PR's authored documentation to M-Press in 1560d21324473f52580968dca44bac368ce02fbc and updated the corresponding existing pages in all nine languages (de, fr, id, ja, ko, pt, ru, zh-cn, zh-tw).

Dependency: this branch now merges #6137 at 7b0f6dc094bfe7bf693df0320338a9bfc5f86ec8 and targets `codex/mpress-markdown`. The original implementation edits are preserved. No Starlight sources were restored and no translation sidecar state or audit settings were changed. Pushed fast-forward to the original fork branch as taliesin-ai.

Validation: pinned M-Press 1.0.17 strict build (2,010 pages), link/asset check, and rendered-site check (2,011 HTML pages, zero errors) passed. Linux `go test ./pkg/application -run 'Test.*Autostart' -count=1` passed, as did a local Go check of the documented DisableWithOptions signature. All ten language pages have matching code blocks. Scoped diff/manual review and whitespace checks passed.

Known gate: the global translation audit remains failing (7,678 findings on both the pinned migration base and this branch); findings outside this page are unchanged. The migration audit blocker is separate work, and the gate remains enabled. `coderabbit --plain` was attempted but is unsupported; the single narrow fallback review was rate-limited. No usage credits were spent.

PR #6099, opened 2026-09-09
fix: start default signal handler in App.Run

### LINKED ISSUE #6096: [v3] Default signal handler is constructed but never started — SIGINT/SIGTERM bypass App.Quit and all shutdown hooks
## Description

In v3.0.0-beta.19, the default application signal handler is constructed but never started, so SIGINT/SIGTERM do not go through `App.Quit()` — the OS default disposition kills the process immediately and none of the app's cleanup runs (service `ServiceShutdown`, window/tray teardown, `PostShutdown`, etc.).

`pkg/application/signal_handler_desktop.go` builds the handler:

```go
func (a *App) setupSignalHandler(options Options) {
	if !options.DisableDefaultSignalHandler {
		a.signalHandler = signal.NewSignalHandler(a.Quit)
		a.signalHandler.Logger = a.Logger
		a.signalHandler.ExitMessage = func(sig os.Signal) string {
			return "Quitting application..."
		}
	}
}
```

…but nothing ever calls `a.signalHandler.Start()`. `internal/signal.SignalHandler` only registers its `signal.Notify` channel inside `Start()`:

```go
func (s *SignalHandler) Start() { ... }
```

The only `Start()` caller in the whole module is the dev watcher (`internal/commands/watcher.go:67`), not `pkg/application`. So `DisableDefaultSignalHandler` currently has no observable effect: handler or no handler, signals bypass the app.

## Impact

Any v3 app that manages external resources leaks them on Ctrl-C / SIGTERM. In our case (an app that supervises a Docker container and spawned media-player processes), SIGTERM left the container running and the child processes orphaned; the same happens when `wails3 dev` sends its interrupt.

## To Reproduce

1. `wails3 init -t vanilla`, add a `ServiceShutdown` (or `PostShutdown`) that logs/touches a file.
2. `go run .` (or run the built binary), then send SIGTERM (or Ctrl-C).
3. The process dies immediately; the shutdown hook never runs. Expected: "Quitting application..." message and a graceful `App.Quit()`.

## Workaround

```go
	DisableDefaultSignalHandler: true,
	// ...
})
go func() {
	ch := make(chan os.Signal, 1)
	signal.Notify(ch, os.Interrupt, syscall.SIGTERM)
	<-ch
	go app.Quit()
	<-ch // second signal: force exit
	os.Exit(1)
}()
```

## Suggested fix

Call `a.signalHandler.Start()` during `App.Run()` (guarded by `DisableDefaultSignalHandler`). Happy to send a PR if that's the preferred shape.

## System

- Wails v3.0.0-beta.19
- Verified on macOS 26 (arm64), Go 1.27; the construct-but-never-start path is platform-independent (`//go:build !ios`).

jabrailkhalil:
## Summary

Fixes #6096. In v3.0.0-beta.19 the default application signal handler was constructed in `setupSignalHandler` (`pkg/application/signal_handler_desktop.go`) but never started, so SIGINT/SIGTERM bypassed `App.Quit()`: the OS default disposition killed the process and none of the app's cleanup ran (service `ServiceShutdown`, window/tray teardown, `PostShutdown`, etc.).

`App.Run()` now starts the handler right after the run lock is taken, so the first Ctrl+C / SIGTERM goes through `App.Quit()` and the shutdown hooks. The start is guarded by `a.signalHandler` being non-nil, so:

- `DisableDefaultSignalHandler: true` stays a no-op (the handler is never constructed),
- platforms without a desktop signal handler (iOS, where `platformSignalHandler` is empty) stay a no-op.

## Tests

- `go build ./internal/signal/... ./pkg/application/...` — passes
- `go test ./pkg/application/...` — passes (including the service lifecycle suites)
- `go test ./...` — passes except four pre-existing environment-dependent failures (`internal/generator`, `internal/wake`, `internal/wake/exec`, `internal/webview2/pkg/edge`), which fail identically on a clean `master` checkout in this environment
- `go vet ./pkg/application/...` — no findings on the changed lines (the `unsafe.Pointer` findings are pre-existing Windows-only files, identical on clean `master`)
- `golangci-lint run --new-from-rev=origin/master --new ./pkg/application/...` — 0 issues (the repository ships no lint configuration and its CI does not run golangci-lint, so this used the default profile)

## Verification

The change is a two-step platform-independent sequence: forward the first signal into the already-existing `internal/signal.SignalHandler` goroutine, which logs the configured exit message and calls `a.Quit()` (the same path `wails3 dev` and the watcher already use). A full GUI signal round-trip needs an interactive desktop session and could not be executed in this environment; the handler and its wiring are covered by the package tests above.

Fixes #6096

## Summary by CodeRabbit

- **Bug Fixes**
  - Applications now handle Ctrl+C and SIGTERM through the application shutdown flow, allowing registered shutdown hooks to run before exiting.
  - Signal handling remains unchanged when disabled or unavailable on the current platform.

### DIFF
--- a/v3/pkg/application/application.go
+++ b/v3/pkg/application/application.go
@@ -647,6 +647,15 @@ func (a *App) Run() error {
 	a.starting = true
 	a.runLock.Unlock()
 
+	// Start the default signal handler, when one was set up at creation, so
+	// that Ctrl+C and SIGTERM go through App.Quit and the application's
+	// shutdown hooks instead of the OS default disposition. It is a no-op
+	// when DisableDefaultSignalHandler was requested or on platforms that
+	// carry no signal handler.
+	if a.signalHandler != nil {
+		a.signalHandler.Start()
+	}
+
 	// Ensure application context is cancelled in case of failures.
 	defer a.cancel()
 


leaanthony:
Thank you for investigating this and putting together a fix! We were working on the same issue at the same time and opened #6098 just a few minutes apart from this PR, so the overlap was simply parallel work.

We'll consolidate the fix in #6098, which now includes signal regression tests and the mobile-platform handling, and close this PR to keep one path forward for #6096. We appreciate your contribution and the time you spent validating it—thanks again!

